PR #5, opened 2016-09-16
Adding the migration that was missing for the 'analyze' app.

rzelayafavila:
Quick pull request - just adding the migration created by the makemigrations mgmt command. @dongbohu @mhuyck @cgreene

### DIFF
--- a/adage/analyze/migrations/0005_auto_20160915_1523.py
+++ b/adage/analyze/migrations/0005_auto_20160915_1523.py
@@ -1,0 +1,90 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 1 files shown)


mhuyck:
This is just the generated migration that comes after 0004, right? LGTM.

Anyone else notice that PR#5 is migration #0005?

cgreene:
Probably didn't need to name all 3 of us, but I'll go ahead and approve for clarity :+1:

mhuyck:
The UI here is definitely more cluttered than Bitbucket.

Go into the diff view and find the green "Review Changes" button. Clicking that brings up a menu that includes "approve."

rzelayafavila:
Yep, totally, the interface is more cluttered than in bitbucket. Thanks everyone for approving! Also nice catch @mhuyck with the PR Number, I hadn't noticed that!

cgreene:
Ok - thanks for the tip @mhuyck! I got it now. I agree that the number of features they added are a bit overwhelming.

PR #6, opened 2016-09-17
Enable CircleCI & run Django tests on it

cgreene:
Offers docker support for open source. Don't approve/merge this for a bit. Just testing what we need to configure.


cgreene:
At this stage, CircleCI is setup and tests are running. Going to use this as a baseline for container generation.

cgreene:
useful resource: https://karthikv.net/articles/circleci-docker-flow/

cgreene:
Also worth looking at - the custom settings bit of this but it seems a bit hacky and may be difficult to do on something like the google container engine:
http://michal.karzynski.pl/blog/2015/04/19/packaging-django-applications-as-docker-container-images/

cgreene:
If you want to run the docker container, something like this will do:
`docker run -p 8000:8000 greenescientist/adageserver:latest`

Then hit port 8000 to see the template missing error :)

adage/adage/settings.py
```diff
@@ -157,7 +165,7 @@
 STATIC_URL = '/static/'
 # STATIC_ROOT = os.path.join(BASE_DIR, '/static/')
 STATICFILES_DIRS = (
-    os.path.join(BASE_DIR, '/static/'),
+    os.path.join(BASE_DIR, 'static'),
```
cgreene: This is actually @rzelayafavila  change.

README.md
```diff
@@ -1,3 +1,142 @@
 # adage-server
```
cgreene: These reflect @rzelayafavila's changes that got pulled it when I attempted to rebase on master.

cgreene:
Ok - this enables CircleCI for this repository. The current CircleCI tasks include running the django tasks and building a docker image that containerizes the django components. This docker image is then automatically uploaded to dockerhub. It does not yet enable continuous deployment, but I'd like to get the testing & build notifications live for future PRs so I propose that we review this now. Continuous deployment will be a separate PR.

mhuyck:
Awesome! Will review tonight.

mhuyck:
In looking this over, I am realizing that I think @cgreene and @rzelayafavila are both trying to cajole Django's `runserver` into serving up the interface files via steps like `python manage.py collect static` and having link across via the `static` symlink. While that might be convenient for development, we have never set it up to work that way, so that is probably why @rzelayafavila ran into trouble.

While convenient, that configuration is a step removed from a realistic production setup and my preference has been to develop and test on a configuration as closely matched to production as possible. What we have done instead is to use `nginx` to reverse proxy requests for Django and directly serve the interface files. This is why both @dongbohu and I have `nginx` running on our development configurations.

adage/adage/config.py.template
```diff
@@ -39,25 +43,26 @@ OS_CONFIG = {
...
+                                     'PseudomonasAnnotation.tsv'),
 })
 
 # test server deployment credentials and configuration (maybe on AWS)
 TEST_CONFIG = {
     'data':          DATA_CONFIG,
     'host_conn':     AWS_CONFIG['host_conn'].copy(),
     'virt_env':      OS_CONFIG['home_dir'] + '.virtualenvs/adage',
-    'django_dir':    OS_CONFIG['home_dir'] + 'adage-server/adage',
+    'django_dir':    os.path.join(REPO_ROOT, 'adage'),
```
mhuyck: Okay, these are the kinds of changes you wanted to test, right? I'll see if they break my development configuration.

adage/adage/settings.py
```diff
@@ -23,10 +24,17 @@
     CONFIG = CI_CONFIG
     pg_user = os.environ.get('PG_USER')
     pg_pass = os.environ.get('PG_PASSWORD')
-    CONFIG['dbmaster']['USER'] = pg_user
-    CONFIG['dbmaster']['PASSWORD'] = pg_pass
```
mhuyck: Thank you for cleaning out my cruft, and sorry I left it in there! I got Codeship running more easily than I expected and I never went back and removed this.

adage/adage/settings.py
```diff
@@ -23,10 +24,17 @@
...
     CONFIG['databases']['default']['USER'] = pg_user
     CONFIG['databases']['default']['PASSWORD'] = pg_pass
+elif os.environ.get('CIRCLECI') == 'true':
+    with open(os.path.join(BASE_DIR, 'adage', 'config.py.template')) as f:
+        exec f
```
mhuyck: a bit of a hack, but it's a simple one and kinda cool that this works, isn't it?

adage/analyze/management/commands/import_data.py
```diff
@@ -12,6 +11,10 @@
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
+# Django imports
+from django.core.management.base import BaseCommand, CommandError
+from django.conf import settings
```
mhuyck: I get why we want the django imports after the python libraries, but moving it below the code that sets up logging is going to make `flake8` cranky.

Oh, wait. `flake8` is already cranky about the import hacking below. Never mind. It makes sense to me to leave the logging setup above. Just know that once we correct the hack below we will still have a code quality bot complaining.

adage/analyze/management/commands/import_data.py
```diff
@@ -22,7 +25,8 @@
 #        repository into `adage-server` or (b) make the code we need into a
 #        pip-installable utility package that gets imported by both the `ADAGE`
 #        project and the `adage-server` project.
-sys.path.append(os.path.abspath('../../'))
+repo_parent = os.path.dirname(os.path.dirname(settings.BASE_DIR))
+sys.path.append(os.path.abspath(repo_parent))
```
mhuyck: - [ ] test on dev setup (@mhuyck)

build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+IFS=$'\n\t'
```
mhuyck: A quick comment about why these two terse `bash`-isms are here would be helpful to me.

push.sh
```diff
@@ -0,0 +1,41 @@
...
+#docker push $REMOTE/$NAME:$HASH
+#docker tag -f $NAME $REMOTE/$NAME:latest
+#docker push $REMOTE/$NAME:latest
+
+#docker logout https://$REMOTE
+
+docker login --username=$DOCKER_USER --password=$DOCKER_PASSWD \
+  --email=$DOCKER_EMAIL
+# Flag --email has been deprecated, will be removed in 1.13. CircleCI Still
+# requires this.
```
mhuyck: thank you for documenting these little notes

mhuyck:
Okay, I got through the code and just had some minor questions and suggestions. In the morning I will test the sections I marked for myself with checklists.

adage/adage/settings.py
cgreene: indeed!

adage/adage/settings.py
cgreene: no problem

adage/adage/config.py.template
cgreene: Yup - wanted to know if this stuff still works :)

adage/analyze/management/commands/import_data.py
cgreene: Since my changes weren't an improvement anyway, I moved them back up.

build.sh
cgreene: Added comment. Short version: we want bash to report failure if anything goes wrong.

cgreene:
@mhuyck : fixed the bad things that you noticed 👍  - looking forward to hearing about whether or not the relative path stuff botched your deployment setup. Thanks!

adage/adage/config.py.template
```diff
@@ -7,6 +7,10 @@
 # and replace with values tailored to your deployment.
 import random
 
+import os
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
mhuyck: This does not produce the right result when using `fab` to deploy. The reason is that the execution context for `fab` is not the same as the deployment target, and the location of `__file__` on the deploying server (my laptop, in this case) is not the same as where `__file__` belongs when deployed.

The specific line that's failing is `fabfile/adage_server.py`:141, but it's because we're attempting to `cd` into `CONFIG['django_dir']` before running those commands, and that construct is everywhere.

There might be another way to fix this... I'm looking into it.

adage/analyze/management/commands/import_data.py
mhuyck: This change *should* be okay because `import_data` is run from an already-deployed server. Testing just this part...

adage/adage/settings.py
```diff
@@ -23,10 +24,17 @@
...
+    CONFIG = CI_CONFIG
+    CONFIG['databases']['default']['NAME'] = 'circleci_test'
+    CONFIG['databases']['default']['USER'] = 'ubuntu'
+    CONFIG['databases']['default']['PASSWORD'] = ''
```
mhuyck: So... does user `ubuntu` automatically have access to the `circleci_test` database with no password? I don't see where this is set up.

adage/adage/config.py.template
mhuyck: Given the above, I don't think there is any way around having to *specify* the destination directory for the repo. That said, I don't see a reason we need to use an absolute path in that specification.

What we need here is a way to determine at deploy time where the `REPO_ROOT` will land, right? Is there a way we can reliably determine where that will be inside the Docker image? Does something fairly standard like `"$HOME"` or `~/` get us the right location? If so, we should switch back to using `home_dir` or define a `repo_root` in the `OS_CONFIG` parameters and I can tweak the `fab` commands to do the right thing.

adage/analyze/management/commands/import_data.py
mhuyck: Test deployment still running, but it's past the `import_data` step so I think we're good here... aha! Just finished.

Yep. Working. Data loaded fine as well.

mhuyck: Oh, drat. Of course the deployment pulled from the official repo, not this branch, so this was not a successful test. Trying again.

cgreene:
That is how circleci configures posters in their testing environment.

adage/analyze/management/commands/import_data.py
mhuyck: Okay, `import_data` still succeeds when pulling from the branch that actually has the changes we're testing. Success on this front.

adage/adage/config.py.template
cgreene: I see why this was specified now. I don't know a way around specifying where the repo will land in your deployment pipeline. With the CI providers we can depend on them to clone, and then use paths relative to that. But for the first deployment, I think a location for the destination needs to be specified. Do you want to switch back to a home_dir? If we leave the other locations relative it will aid our deployment in other contexts.

adage/adage/settings.py
cgreene: Replying to the e-mail didn't produce this comment in precisely the right place. So I'll put it here and delete the email response (sorry for the duplicate e-mails):
> That is how circleci configures posters in their testing environment.

cgreene:
@mhuyck - in github's interface I can't reply directly to a comment so I'll quote the first part here:

> In looking this over, I am realizing that I think @cgreene and @rzelayafavila are both trying to cajole Django's runserver into serving up the interface files via steps like python manage.py collect static and having link across via the static symlink. While that might be convenient for development, we have never set it up to work that way, so that is probably why @rzelayafavila ran into trouble.

In general, I think it is convenient to users if we don't make choices that preclude this option. As far as I can tell, the changes currently enable this without affecting nginx configuration in any way. That is, if you had nginx you still do, but now if you want to use the lightweight dev server you can do that too.

adage/analyze/management/commands/import_data.py
cgreene: woo! 👍 thrilled!

mhuyck:
"it is convenient to users if we don't make choices that preclude this option"

I fully agree! Did @rzelayafavila get it working? I hadn't heard that. My comment was just an attempt at adding insight to the problem in hopes of finding resolution.

adage/adage/settings.py
mhuyck: Gotcha. This is similar to how Codeship did things.

adage/adage/config.py.template
mhuyck: I just realized another approach would be to put CI-specific configuration into a separate section of the CONFIG. I did that for Codeship (see CI_CONFIG). So we can revert the changes to TEST_CONFIG and set up whatever we want in CIRCLECI_CONFIG. That would keep the various deployment scheme settings isolated.

cgreene:
@mhuyck : pretty sure the changes got it working for @rzelayafavila. @rzelayafavila, is it true that you're now having success with the dev server?

rzelayafavila:
@cgreene and @mhuyck: Yep, the change to doing the os.path.join (removing the slashes from 'static') definitely fixed it for me - in fact I was wondering why it was working for you and @dongbohu in the first place until I learned that you both use nginx to serve the interface files. I agree that we should make it so that users/developers can start a development server whichever way, that we don't force one path (especially if django offers this possibility). I think the 'collectstatic' command is not even required for running the django test server, just having this STATICFILES_DIRS point to where the static files are, and this should not have extra slashes since os.path.join already adds one.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ENV ADAGE_SRVSRC=$ADAGE_SRV/$ADAGE_SRC
+
+RUN apt-get update && apt-get install -y \
+  python \
+  python-pip \
+  python-psycopg2
+
+# Create required directories
+WORKDIR $ADAGE_SRV
+RUN mkdir media static logs
```
cgreene: I have no idea. If nothing I can remove this. @mhuyck : are we using this feature?

adage/adage/settings.py
```diff
@@ -23,10 +24,17 @@
...
     pg_pass = os.environ.get('PG_PASSWORD')
-    CONFIG['dbmaster']['USER'] = pg_user
-    CONFIG['dbmaster']['PASSWORD'] = pg_pass
     CONFIG['databases']['default']['USER'] = pg_user
     CONFIG['databases']['default']['PASSWORD'] = pg_pass
+elif os.environ.get('CIRCLECI') == 'true':
+    with open(os.path.join(BASE_DIR, 'adage', 'config.py.template')) as f:
+        exec f
+    CONFIG = CI_CONFIG
+    CONFIG['databases']['default']['NAME'] = 'circleci_test'
```
cgreene: Copied the approach from @mhuyck with the goal of making minimal changes necessary to enable CI - thoughts @mhuyck?

adage/adage/settings.py
```diff
@@ -23,10 +24,17 @@
...
+    CONFIG['databases']['default']['USER'] = 'ubuntu'
+    CONFIG['databases']['default']['PASSWORD'] = ''
+    CONFIG['databases']['default']['PORT'] = '5432'
+    print(CONFIG['data']['data_dir'])
```
cgreene: Good catch!

circle.yml
```diff
@@ -0,0 +1,39 @@
+
+machine:
+  pre:
+    # Add get_pseudomonas to the machine for two required files
+    - hg clone https://bitbucket.org/greenelab/get_pseudomonas
```
cgreene: Yup

Dockerfile
cgreene: Looks like it is unused via my inspection of settings. Removed! 👍

mhuyck: Right. No `media`. What usually goes there? Things like images (we only have a couple) are currently in `interface/src/assets/` and get copied to `interface/bin/assets/` upon a successful `grunt` build of the interface. The `adage/static/` directory is a symlink to `interface/bin/`. I believe this is also the way Tribe is set up (?) and what makes Django's `collectstatic` command superfluous.

adage/adage/settings.py
mhuyck: Good point. If they are static, then the config does not need to be set here and could be left in `config.py.template`. In the case of Codeship, the database user and password are passed as environment variables (`PG_USER` and `PG_PASSWORD`) so we need to look them up at runtime.

mhuyck: I called it `CI_CONFIG` but I might have been too optimistic about generalizing. It is really `CODESHIP_CONFIG`, I guess, since CircleCI seems to do things differently. Maybe we should rename it that way to avoid confusion and create a `CIRCLECI_CONFIG` that is separate. Will we be using both Codeship and CircleCI going forward?

adage/adage/config.py.template
mhuyck: I think if we build a `CIRCLECI_CONFIG` by starting with `CIRCLECI_CONFIG = TEST_CONFIG.copy()` and then putting the @cgreene's changes to `TEST_CONFIG` into that via `CIRCLECI_CONFIG.update({...` (following what I did for `CI_CONFIG`) then all of the `BASE_DIR` and `REPO_ROOT` references, which confuse `fab` deployment, will be isolated where they are needed.

With this change, I think we could merge and not have any conflicts with the existing AWS and development deployment `fab` scripts. Does that make sense?

mhuyck: True, but see my comment above about incompatibility with `fab` scripts. As long as we keep `CONFIG` settings specific to a deployment scenario this should work fine.

cgreene: @mhuyck : Can you make a proposal via edits to this branch? I don't clearly see which changes you are suggesting. I think it may be faster and more efficient for you to directly propose changes. My only preference here is to see the repository know where it needs to be placed in the minimum number of places. Thanks!

rzelayafavila:
Ok, LGTM for all the changes I suggested!

mhuyck:
I've been delayed in tweaking and testing due to kid still home from school and a number of errands this morning. Update & final testing coming tonight. Sorry for the delay!

adage/adage/config.py.template
```diff
@@ -136,9 +141,30 @@ CI_CONFIG.update({
                         'PORT': '5434',  # v. 9.4
                      },
     'data':          {
-                        'data_dir': DATA_CONFIG['data_dir'],
-                        'annotation_file': DATA_CONFIG['annotation_file'],
+                        'data_dir': '/home/rof/clone/data/',
```
cgreene: Any chance we can use the DATA_CONFIG here? This would allow it to still use this configuration for CircleCI.

adage/adage/config.py.template
```diff
@@ -136,9 +141,30 @@ CI_CONFIG.update({
...
+                        # database name and user cannot have upper case letters
+                        'NAME': 'circleci_test',
+                        'USER': 'ubuntu',
+                        'PASSWORD': '',
+                        'HOST': '127.0.0.1',
+                        'PORT': '5434'
+                        }
+                     },
+    'data':          {
+                        'data_dir': os.path.join(REPO_ROOT, 'data'),
```
cgreene: Oh - I see - this got added here. What do you think about a (minor) reduction in code duplication by using a single CI section and then just updating the 'databases' key?

mhuyck: Note that I also reverted the `data_dir` in the `OS_CONFIG` above (to keep AWS deployment happy), so maybe not much reduction by doing that.

mhuyck: Aha... you did see that. Hadn't seen your other comment first.

Actually, this should be possible. Now I see what you were asking. Yes, a general CI_CONFIG with base CI settings and separate CIRCLECI_CONFIG should work fine. As long as they don't try to use the `OS_CONFIG`, which I set back to AWS-compatible paths.

PR #7, opened 2016-09-19
Adding instructions to README.md on how to get a development instance of the server running.

rzelayafavila:
@cgreene @mhuyck @dongbohu : Let me know how this looks and if we should make any changes to the content/structure of the instructions! Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,11 +121,16 @@ Download the `get_pseudo_sdrf.py` and `gen_spreadsheets.py` files from
 and put them in whichever parent directory the adage-server repository has been
 cloned into.
 
-
 ### Build adage-server web interface
 
 ```shell
 cd interface/
+
+# Download newest version of Node.js. Also, the command
+# "curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -"
+# is needed before the "sudo apt-get install -y nodejs" if you are
+# running a Ubuntu version older than 16.04.
+sudo apt-get install -y nodejs
 
 # Install needed interface packages
 sudo npm -g install grunt-cli karma-cli bower
--- a/README.md
+++ b/README.md
@@ -1,3 +1,112 @@
 # adage-server
 
 [![Code Climate](https://codeclimate.com/github/greenelab/adage-server/badges/gpa.svg)](https://codeclimate.com/github/greenelab/adage-server)
+
+## Get a working instance of the adage-server running:
+
+### Clone the adage-server repository
+```shell
+git clone git@github.com:greenelab/adage-server.git
+```
+
+### Edit settings in config.py file
+Set the full path of parent directory where adage-server repository was
+cloned into in the `'home_dir'` key of the `OS_CONFIG` setting.
+```python
+OS_CONFIG = {
+    'home_dir':  '/<path_to_directory>/',
+}
+
+```
+
+Set up Database Name, User, Password and Host in `databases` settings:
+```python
+CONFIG.update({
+    'databases': {'default': {
+                     'ENGINE': 'django.db.backends.postgresql_psycopg2',
+
+                      # database name and user cannot have upper case letters
+                      'NAME': '<your_adage_database_name>',
+                      'USER': '<your_adage_db_username>',
+                      'PASSWORD': '<your_db_username_password>',
+
+                      # Wherever PostgreSQL is being hosted,
+                      # usually localhost for development
+                      'HOST': 'localhost',
+
+                      # Port where it is being hosted from,
+                      # usually 5432
+                      'PORT': '5432',
+                     }
+                 },
+
+})
+```
+
+### Set up Database Name, User, Password and Host on PostgreSQL
+```shell
+# Switch to postgres user, enter the superuser password when you
+# are prompted
+sudo su - postgres
+
+# Create adage-server database specified in previous step
+createdb <your_adage_database_name>
+
+# Create adage-server database user specified in previous step
+createuser -P <your_adage_db_username>
+# This prompts you to enter the password for new role/user (also
+# the one specified in config.py file in previous step)
+
+# Enter psql interface:
+psql
+
+# Give all privileges of this newly created database to the
+# newly created user
+GRANT ALL PRIVILEGES ON DATABASE <your_adage_database_name> TO <your_adage_db_username>;
+
+# Also give this user permissions to create databases - this is needed to be
+# able to run the Django test suite (since a test database is created).
+ALTER USER <your_adage_db_username> CREATEDB;
+
+```
+
+### Install Python dependencies
+```shell
+# Start a Python virtual environment in the adage-server root
+# or wherever you keep your Python virtual environments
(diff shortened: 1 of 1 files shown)


rzelayafavila:
Hmmm, I see codeclimate is complaining a lot about a lack of blank lines in the new lines in the README - I will put these in so it is happy.

dongbohu:
Should we also include web server configuration (nginx, gunicorn, elasticsearch)?

mhuyck:
Wow... codeclimate sure wants a lot of white space! I guess if that's the convention we should abide.

@dongbohu makes a good point but I think we should make a statement near the top that says something like "the steps here assume you have already installed `postgresql` (>=9.4), `nginx` (>=1.10), `supervisord` (>=3.2), `python` (2.7)  and `elasticsearch` (1.7) on `Ubuntu` (16.04).

`gunicorn` will be installed (along with `Django` itself) by the `pip install -r requirements.txt` step (line 81).

Oh, and then there are the build dependencies. This is silly. We have scripts to set these up and I've gotten them all updated and working. Shouldn't we refer to that instead of re-documenting it all? I'll submit the PR with my updates this afternoon.

README.md
```diff
@@ -1,3 +1,112 @@
 # adage-server
 
 [![Code Climate](https://codeclimate.com/github/greenelab/adage-server/badges/gpa.svg)](https://codeclimate.com/github/greenelab/adage-server)
+
+## Get a working instance of the adage-server running:
+
+### Clone the adage-server repository
+```shell
+git clone git@github.com:greenelab/adage-server.git
```
mhuyck: If developers will be submitting pull requests, isn't it easier to make a personal fork first and then clone that? The README should walk people through it the way we want them to do it.

README.md
```diff
@@ -1,3 +1,112 @@
...
+whichever parent directory the adage-server repository has been cloned into.
+
+
+### Build adage-server web interface
+```shell
+cd interface/
+
+# Download newest version of Node.js
+curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -
+sudo apt-get install -y nodejs
```
mhuyck: This step is not necessary if running on Ubuntu 16.04.

README.md
```diff
@@ -1,3 +1,112 @@
...
+
+# Install needed interface packages
+sudo npm -g install grunt-cli karma-cli bower
+npm install
+bower install
+
+# Run Grunt commands to build the interface
+grunt build
+grunt compile
+grunt test
```
mhuyck: All three commands are run by default if you just type `grunt`.

mhuyck:
That big red x is kinda harsh. This is good stuff, @rzelayafavila... I just had some minor suggestions to add!

README.md
```diff
@@ -1,3 +1,112 @@
...
 [![Code Climate](https://codeclimate.com/github/greenelab/adage-server/badges/gpa.svg)](https://codeclimate.com/github/greenelab/adage-server)
+
+## Get a working instance of the adage-server running:
+
+### Clone the adage-server repository
+```shell
+git clone git@github.com:greenelab/adage-server.git
+```
+
+### Edit settings in config.py file
```
cgreene: Where does this file live? Confused me when I first started with this project.

README.md
```diff
@@ -1,3 +1,112 @@
...
+                      # Port where it is being hosted from,
+                      # usually 5432
+                      'PORT': '5432',
+                     }
+                 },
+
+})
+```
+
+### Set up Database Name, User, Password and Host on PostgreSQL
```
cgreene: Maybe preface this with: example setup using postgresql. Don't think we want to consider postgresql required unless we know other databases won't work.

README.md
```diff
@@ -1,3 +1,112 @@
...
+
+### Run Django migration of database tables
+```shell
+python manage.py migrate
+```
+
+### Download other necessary files
+Download the `get_pseudo_sdrf.py` and `gen_spreadsheets.py` files from
+https://bitbucket.org/greenelab/get_pseudomonas/src , and put them in
+whichever parent directory the adage-server repository has been cloned into.
```
cgreene: Technically don't these go into <repo-dir>/adage/? That's where I had to put them.

mhuyck: On my setup these are in `/home/adage/`, which is the home directory of the user account that hosts the source repository. That is the same thing as saying "the parent directory into which the adage-server repository has been cloned." Both answers are right, but I think @rzelayafavila's way of putting it is more general.

README.md
mhuyck: The `config.py` file is an extension to `settings.py` that pulls out any secrets and deployment-specific parameters, so it goes alongside `settings.py`. The idea is that you should edit `adage-server/adage/adage/config.py.template` from the repository with your deployment's information and save it as `config.py`. This puts all of the secrets and deployment-specific information into a single, easily-controlled file.

It is currently a bit noisy in `config.py.template`, but most of those parameters are used by `fab` for building a deployment. The minimal runtime sections are currently `django_key`, `databases`, and `haystack`, though as I think about it we should make `elasticsearch` connection information configurable as well. The default-supplied `haystack` works fine if `elasticsearch` is configured as specified, and (LOL) it appears you can get by with a `django_key` of "&lt;config-secret&gt;" for development purposes, so the only *essential* change is the `databases`, as @rzelayafavila has described below.

cgreene: I'd recommend that:

> The idea is that you should edit adage-server/adage/adage/config.py.template from the repository with your deployment's information and save it as config.py. This puts all of the secrets and deployment-specific information into a single, easily-controlled file.

goes somewhere in the readme.

README.md
cgreene: I did not realize that these could be above the parent of the adage-server repository. Do I want to know how that works?

README.md
rzelayafavila: Both of these commands (including the `sudo apt-get install -y nodejs`) or just the `curl -sL https://deb.nodesource.com/setup_4.x | sudo -E bash -` is not needed for Ubuntu 16.04?

mhuyck: Oh, good question. What I meant was that the node that `apt-get install` chooses from the Ubuntu 16.04 distribution is sufficient. (I believe it's v4.)

mhuyck: This is definitely not the case on Ubuntu 14.04, so if you're on that distro you definitely need to take extra steps for node 4.

README.md
rzelayafavila: So I think both work because the import_data.py file adds that parent directory as well ('../../') to the Python path. There is already an issue in this repo to have that import standardized.

mhuyck: That would be logical to assume. To avoid conflicts during development, I've tried to not mingle source trees. That means that I did some hackish python `import` gymnastics to get them to import. This is precisely why we have Bitbucket Issue 2.

rzelayafavila:
Ok, I have made some changes - let me know how these look!

rzelayafavila:
Ok, thanks everyone!

PR #9, opened 2016-09-26
Adding changes necessary for tribe-client integration. Also adding code so that our Angular app can access the user that is logged in to Tribe.

rzelayafavila:


adage/templates/base.html
```diff
@@ -0,0 +1 @@
+index.html
```
cgreene: Do we have to add this symlink? Can the location of base.html be a setting for tribe_client to make it more configurable?

interface/src/app/app.js
```diff
@@ -19,15 +20,37 @@ angular.module( 'adage', [
...
+.run( function run ( $state, $location ) {
+
+    // Will only redirect if there is no state AND the url is not a
+    // route for tribe_client
+    String.prototype.startsWith = function(prefix) {
+        return this.slice(0, prefix.length) == prefix;
+    };
+
+    if ($state['current']['name'] === '') {
+      if (window.location.pathname.startsWith('/tribe_client')) {
```
cgreene: What does this bit do?

interface/src/app/tribe_auth/resource.js
```diff
@@ -0,0 +1,14 @@
...
+    .factory('User', ['$resource', function($resource) {
+        return $resource('/tribe_client/return_user', {}, {
+            query: {
+                method: 'GET',
+                isArray: false
+            }
+        });
+    }])
+
+;
```
cgreene: github points out that there's no newline at the end of the file.

interface/src/app/tribe_auth/tribe-profile-button.tpl.html
```diff
@@ -0,0 +1,14 @@
...
+  </a>
+  <ul class="dropdown-menu" uib-dropdown-menu aria-labelledby="simple-dropdown">
+    <li><a href="/tribe_client">
+      <i class="fa fa-cubes"></i> My Gene Sets</a></li>
+    <li class="divider"></li>
+    <li><a href="/tribe_client/logout">
+      <i class="fa fa-sign-out" style="color:SteelBlue"></i>
+      &nbsp;&nbsp;Sign Out</a></li>
+  </ul>
+</li>
```
cgreene: A few of these no newlines (one more below)

interface/src/index.html
```diff
@@ -70,12 +70,22 @@
...
+            <profile-button></profile-button>
           </ul>
         </div>
       </div>
     </div>
 
-    <div class="container" ui-view="main"></div>
+    <div class="container" ui-view="main">
+    <br>
+      {% block content %}
```
cgreene: Will this prevent us from using nginx or a CDN to serve this as a simple static file?

cgreene:
The codeclimate new issues also seem reasonable from a DRY point of view - is there a way to avoid the code duplication around the promise?
https://codeclimate.com/github/greenelab/adage-server/pull/9

adage/templates/base.html
rzelayafavila: Mmm, great question - it should probably be a setting to make it more adaptable, like you say

interface/src/app/app.js
rzelayafavila: Good question - I actually wanted to ask this to all the reviewers ( @dongbohu @mhuyck @cgreene ): Right now, this bit of code (and commenting out the "$urlRouterProvider.otherwise('/home');" function on line 14 of this file) prevents the angular app from redirecting to '/#/home' whenever the url starts with '/tribe_client' (since those urls are not part of the Angular app). However, this bit of code still redirects the user to '/#/home' if no state is defined in the current url AND the url doesn't start with 'tribe_client'.

This approach seems pretty hacky to me though, but I couldn't find an 'official' way of avoiding redirection with ui-router if the url starts with 'tribe_client'. Does anybody have any more ideas on this? Thanks!

interface/src/app/tribe_auth/resource.js
rzelayafavila: Ok, will add those.

interface/src/app/app.js
dongbohu: This solution does seem like a hack. Maybe we can enable HTML5 mode on the web server so that "/#/" won't be needed any more?

rzelayafavila: Hmmm that seems like a possibility - another possibility that I was thinking of was just porting all of the interface functionality of tribe-client to our Angular app (so that we don't have to deal with those annoying urls). This way, the only way we are using the '/tribe_client/' urls is as API endpoints (to get the access token, get the user that is logged in, and the genesets), but we don't actually send the users to these urls in the browser.

interface/src/app/tribe_auth/tribe_user.js
```diff
@@ -0,0 +1,61 @@
+angular.module('adage.tribe_auth.user', [
+    'adage.tribe_auth.resource',
+])
+
+.factory( 'UserFactory', function( User ) {
```
dongbohu: Use inline array annotation instead.

rzelayafavila: Aha - good catch, I will change this.

interface/src/app/tribe_auth/tribe_user.js
```diff
@@ -0,0 +1,61 @@
...
+                delete user[member]; // delete properties from user
+            }
+            for (member in newUser) {
+                user[member] = newUser[member];// add properties back to user
+            }
+        }
+    };
+})
+
+.directive('profileButton', function(UserFactory) {
```
dongbohu: Inline array annotation.

interface/src/app/tribe_auth/resource.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * The resources for fetching the OAuth2 token from the tribe_client url
+ */
+angular.module("adage.tribe_auth.resource", ['ngResource'])
+    .factory('User', ['$resource', function($resource) {
+        return $resource('/tribe_client/return_user', {}, {
```
dongbohu: If you want, you can simplify the code into:
```
return $resource('/tribe_client/return_user', {}, {}
)};
```
Then in line #15 of **tribe_user.js**, instead of calling `User.query(...)`, call `User.get(...)`.

interface/src/app/app.js
cgreene: @rzelayafavila : Can there be a reusable angular app associated with tribe_client that provides the interface functionality? Thinking about how it can be modular across servers but also provide the functionality that you describe.

rzelayafavila: @cgreene - I agree, I think that would probably be the most pragmatic way of doing it so that we could then just use it in different webservers. The catch then will be that the UI of that webserver will have to use Angular to be able to use that app, but it is definitely much better than having not having a re-installable app for the front end.

cgreene: Well - someone could always write their own for another javascript framework. Seems like an angular one makes sense here though, since it is in angular and, probably, your PILGRM2 server.

rzelayafavila: Ok - so probably what I am going to do for the purpose of this Pull request is:
a) Move all the interface functionality in tribe-client to the Angular app, but keep it in a separate module/folder from the other code.
b) File an issue to extract that code later on and make it into a reusable Angular module

mhuyck: I like keeping all of our front end in Angular. @rzelayafavila: the approach you describe sounds like a good pragmatic way to get there.

adage/adage/settings.py
```diff
@@ -195,3 +196,7 @@
         'handlers': ['console'],
     },
 }
+
+TRIBE_ID = CONFIG['tribe_id']
+TRIBE_SECRET = CONFIG['tribe_secret']
+TRIBE_REDIRECT_URI = CONFIG['tribe_redirect_uri']
```
mhuyck: Thank you for following my `CONFIG` scheme for handling secrets! Can you also update the `config.py.template` so people know what to fill in for these parameters?

interface/src/app/app.js
```diff
@@ -19,15 +20,37 @@ angular.module( 'adage', [
   $resourceProvider.defaults.stripTrailingSlashes = false;
 }])
 
-.run( function run () {
+.run( function run ( $state, $location ) {
+
+    // Will only redirect if there is no state AND the url is not a
+    // route for tribe_client
+    String.prototype.startsWith = function(prefix) {
```
mhuyck: Nice catch! Changing object prototypes is hazardous for exactly this reason. I'm okay with doing it where necessary, but that should be very rare and it should be done in a way that does not alter the behavior of existing code (including libraries). At minimum we should have tests designed to fail when this type of modification creates a conflict.

interface/src/app/app.js
```diff
@@ -19,15 +20,37 @@ angular.module( 'adage', [
...
+.controller( 'AppCtrl', function AppCtrl ( $scope, $location, UserFactory ) {
+
   $scope.$on('$stateChangeSuccess', function(event, toState, toParams, fromState, fromParams){
     if ( angular.isDefined( toState.data.pageTitle ) ) {
       $scope.pageTitle = toState.data.pageTitle + ' | adage' ;
     }
   });
+
+  UserFactory.getPromise().$promise.then( function() {
+        $scope.user = UserFactory.getUser();
```
mhuyck: I remember digging into this a little on a Tribe PR, but it was a little while back. This line doesn't force a login, right? As I recall, it just ensures that a `User` object is bound to the session.

interface/src/app/tribe_auth/resource.js
mhuyck: Great suggestion, @dongbohu. This idea would simplify the code here. I think I have some instances that would benefit from that change as well.

interface/src/app/app.js
rzelayafavila: Right - it actually just checks to see **if** there is a user bound to the session. If not (if the user returns as null), then just show the session as not being logged in.

adage/adage/settings.py
rzelayafavila: Yes for sure - thanks for reminding me of that!

interface/src/app/tribe_auth/resource.js
mhuyck: The way I think about this is that it's not actually a "line" of the file unless the line is terminated! It works fine, but only because the parser is not being overly strict.

interface/src/app/app.js
rzelayafavila: Right, I agree - good thing is I have done away with all of this now that the Angular app handles the tribe-client

interface/src/app/tribe_auth/tribe_user.js
mhuyck: Good habit. Thank you to @dongbohu for tracking this down while reviewing one of my PRs and giving us a name for it.

interface/src/index.html
rzelayafavila: Yep, this will now change with the new tribe_client UI.

interface/src/index.html
```diff
@@ -70,12 +70,22 @@
                 Download
               </a>
             </li>
+            <li ui-sref-active="active" ng-if="!user">
+              <a href="/tribe_client">
+                Login with Tribe
```
mhuyck: Is it okay if we just call this "Login" and then mention that people can use their Tribe ID if they already have one? It could be confusing otherwise if people think they must be Tribe users to use ADAGE.

interface/src/app/tribe_client/directives.js
```diff
@@ -0,0 +1,28 @@
+angular.module('adage.tribe_client.directives', [
+    'adage.tribe_client.resource',
+])
+
+.directive('profileButton', [ 'UserFactory', function(UserFactory) {
```
mhuyck: 👍

interface/src/app/tribe_client/tribe-profile-button.tpl.html
```diff
@@ -0,0 +1,14 @@
...
+    (Tribe) {{ userObj.username }} <b class="caret"></b>
+  </a>
+  <ul class="dropdown-menu" uib-dropdown-menu aria-labelledby="simple-dropdown">
+    <li><a href="/tribe_client">
+      <i class="fa fa-cubes"></i> My Gene Sets</a></li>
+    <li class="divider"></li>
+    <li><a href="/tribe_client/logout">
+      <i class="fa fa-sign-out" style="color:SteelBlue"></i>
+      &nbsp;&nbsp;Sign Out</a></li>
+  </ul>
```
mhuyck: This looks cool. Can't wait to see it in action to get a feel for how it works.

interface/src/app/tribe_client/tribe_client.js
```diff
@@ -0,0 +1,39 @@
+angular.module('adage.tribe_client', [
+    'adage.tribe_client.resource',
+    'adage.tribe_client.directives',
+])
+
+.factory( 'UserFactory', ['User', function( User ) {
```
mhuyck: Aha. Here is that code. Somehow I missed it before.

interface/src/app/tribe_client/tribe_client.js
```diff
@@ -0,0 +1,39 @@
...
+                promise = User.get({}, function( data ) {
+                    if (data.meta.total_count !== 0) {
+                        user = data.objects[0];
+                    }
+                });
+            }
+            return promise;
+        },
+        resetPromise: function() { // reset the promise in case we need to check the user again
+            promise = false;
```
mhuyck: Maybe a silly question: is it better to use `false` here (and above at line 7) than `null`? Why?

interface/src/app/tribe_client/tribe_client.js
```diff
@@ -0,0 +1,39 @@
...
+        },
+        setUser: function( newUser ) { //set the user object to have properties from the passed user
+            var member = null;
```
mhuyck: I'm not sure this line is adding anything. Is it? I believe the `for` loops below each create a local `member`, which will mask this one.

mhuyck:
Just looked at the circleci tests. Why is it loading the `fabfile.py`? It's failing because `fabric` is not installed, but we don't do that because it's not needed in the build target.

rzelayafavila:
@mhuyck: This branch hasn't incorporated the changes that y'all did with the circle ci settings - is that what is going on?

interface/src/index.html
rzelayafavila: Sure, I'll change it to just "Login"

mhuyck:
@rzelayafavila I didn't realize that. Yes, that would certainly do it.

cgreene:
@mhuyck @rzelayafavila : This pre-dates merging of CircleCI's setup. I think there is a way to rebase against master if you want those tests to run. I attempted it with only moderate success before (i.e. I rebased but it also duplicated all of @rzelayafavila's commits). I think @dhimmel may be more competent in this regard if we want to see the tests run here.

dhimmel:
> I rebased but it also duplicated all of @rzelayafavila's commits

Isn't that what we want? All of @rzelayafavila's commits on top of the current master? Are you saying each of @rzelayafavila's commits appeared twice?

cgreene:
@dhimmel : that happened when I attempted to rebase in a previous PR. That was not what I was aiming for - just what I happened to achieve.

dhimmel:
Okay well I can work with @rzelayafavila when he returns on how to rebase this pull request.

interface/src/app/tribe_client/tribe_client.js
rzelayafavila: Well, my impression was that null does not evaluate to false in JavaScript, right? This is basically just to see if the promise has been returned, which should be a true/false value. On the other hand, the 'user' variable actually holds the user object, and that's why that one is set to 'null' if no User has been returned. Do you think that approach makes sense?

interface/src/app/tribe_client/tribe_client.js
rzelayafavila: Ah, yes, you're right - good catch! I will remove this line.

interface/src/app/tribe_client/tribe_client.js
mhuyck: A `null` is interpreted as a "falsy" value, yes. That's why it made more sense to me to either have the promise object itself or `null`. You could define the `var promise = null;` in line 7 and reset it in line 25 and everything else should work the same.

interface/src/app/tribe_auth/resource.js
rzelayafavila: I agree, great call to simplify this - this is now added.

interface/src/app/tribe_client/tribe_client.js
rzelayafavila: Ok, sounds good, I will change the promise to null as well.

rzelayafavila:
Ok, @mhuyck, @cgreene, @dongbohu: I think I have addressed all the things you mentioned, but let me know if I missed anything. 

**Also**, to log in with Tribe right now, the user clicks on the "Sign in" button, which opens a modal, and clicks a link which opens a new browser tab/window with Tribe's login page. Once there, the user enters their login credentials, agrees for the ADAGE server to have read and write access to their Tribe genesets, and gets sent back to the ADAGE home page in that new browser tab/window. 
**However**, the modal will remain open in the original window, and has a 'Refresh' button that users can click so there is a page refresh but the page & state they are currently on gets preserved (in case they were working on something when they decided to log in with Tribe). I know this is not ideal - I think the ideal setup would be to have users be automatically redirected from Tribe to the page & state where they were in ADAGE in the original window, but this is hard because the OAuth2 protocol requires them to go to the Tribe website before returning to ADAGE. So, in the end, this is the workaround I built.

Do any reviewers have any thoughts on this approach and/or possible alternative ways of doing this login process?

interface/src/index.html
```diff
@@ -70,6 +70,10 @@
                 Download
               </a>
             </li>
+
+            <login-button></login-button>
+            <profile-button></profile-button>
```
cgreene: Should the ng-if logic on profile-button go here? It initially seems odd to have the condition on whether or not the button should be displayed inside the button itself.

cgreene:
Side note - link functions for both buttons seem identical (codeclimate picked this up too). You may consider whether or not you want to define a single function in the interests of DRY:
https://codeclimate.com/github/greenelab/adage-server/interface/src/app/tribe_client/directives.js?from_sha=c325e34c&to_sha=dbcf72a2

interface/src/index.html
rzelayafavila: I agree with that - it does seem much more intuitive to have the main AppCtrl make a request for the User, and then have the ng-if logic here where these directives are inserted.

mhuyck:
@rzelayafavila: I see the conundrum with doing the "OAuth2 dance" and then getting back to where you started. Is there any way to capture the UI state before starting the Tribe login and then have Tribe's redirect take you back to that state?

This is probably a solved problem, but off the top of my head I can think of a couple of ways: perhaps OAuth2 has a parameter we can use as a "callback URL" when complete. If no such facility is available on OAuth2, it should be possible to set a session cookie (maybe using Django's existing mechanisms) with the right information and have Tribe redirect to an adage-server URL that knows to check for session state and restore that when loading.

interface/src/index.html
mhuyck: Agree as well. I think the way I would put it is that the button should "know" how to draw itself and what to do when clicked, but where (and if) the button appears should be up the the "parent" view.

mhuyck:
A question on the duplicate link function code that @cgreene mentions above: does this create two handlers for the same event such that we execute that block of code twice when the `user.update` event fires? If these are separate scopes then this might not pose a direct conflict, but even then it does seem like duplicate work for us to be making two separate requests for user information (we're making an `XMLHttpRequest` somewhere in `getPromise()`, right?) just to get that response into the right scope.

Okay... just dug into this deeper and it looks as if we have *either* the `profileButton` *or* the `loginButton` showing, but never both. I *think* that takes care of the concern about duplicate event handlers, but I'm not so familiar with the inner workings of `ng-if` that I can confidently say we don't accidentally trigger two event responders as described above. A brief review of the docs is reassuring, though: "when an element is removed using ngIf its scope is destroyed and a new scope is created when the element is restored."

interface/src/app/app.js
```diff
@@ -23,14 +23,18 @@ angular.module( 'adage', [
 .run( function run () {
 })
 
-.controller( 'AppCtrl', function AppCtrl ( $scope ) {
+.controller( 'AppCtrl', function AppCtrl ( $scope, UserFactory ) {
```
mhuyck: This needs inline array annotation.

interface/src/index.html
```diff
@@ -71,8 +71,8 @@
               </a>
             </li>
 
-            <login-button></login-button>
-            <profile-button></profile-button>
+            <login-button ng-if="!userObj"></login-button>
+            <profile-button ng-if="userObj"></profile-button>
```
mhuyck: Much clearer this way!

mhuyck:
I hit approve to indicate that you covered that first round of review changes I requested, but see above for a few smaller questions to look into before merging.

rzelayafavila:
Ok @cgreene , @mhuyck I think I got everything - let me know if these new changes look ok!

PR #14, opened 2016-10-04
Add "heavy_genes" filter to NodeResource

dongbohu:
The changes in this pull request is related to gene-gene network web UI. They include:

(1) Add a new "heavy_genes" filter in **api.py** and a test case in **tests.py**.

(2) Instead of including only gene ID, **EdgeResource** now includes all information of the two genes connected by the edge;

(3) A minor typo correction in **tests.py**.

### DIFF
--- a/adage/analyze/api.py
+++ b/adage/analyze/api.py
@@ -2,6 +2,7 @@ from django.conf.urls import url
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Q
 from django.http import HttpResponse
+from genes.api import GeneResource
 from tastypie import fields, http
 from tastypie.resources import Resource, ModelResource
 from tastypie.utils import trailing_slash
@@ -212,6 +213,36 @@ class NodeResource(ModelResource):
         queryset = Node.objects.all()
         resource_name = 'node'
         allowed_methods = ['get']
+        filtering = {
+            'name': ('exact', 'in', ),
+            'heavy_genes': ('exact', ),  # New filter, see apply_filters().
+        }
+
+    def apply_filters(self, request, applicable_filters):
+        object_list = super(NodeResource, self).apply_filters(
+            request, applicable_filters)
+        heavy_genes = request.GET.get('heavy_genes', None)
+        if heavy_genes:
+            # Instead of relying on tastypie/resources.py to catch the
+            # ValueError exception that may be raised in this function,
+            # we catch the one that may be raised by int() below and
+            # raise a customized BadRequest exception with more details.
+            try:
+                # Convert genes to a set of integers so that the
+                # duplicate(s) will be removed implicitly.
+                query_genes = {int(id) for id in heavy_genes.split(',')}
+            except ValueError:
+                raise BadRequest("Invalid gene IDs: %s" % heavy_genes)
+
+            for (i, q) in enumerate(query_genes):
+                q_nodes = {p.node.id for p in
+                           Participation.objects.filter(gene=q)}
+                if i == 0:
+                    related_nodes = q_nodes
+                else:
+                    related_nodes = related_nodes & q_nodes
+            object_list = object_list.filter(id__in=related_nodes)
+        return object_list
 
 
 class ActivityResource(ModelResource):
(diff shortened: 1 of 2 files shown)


adage/analyze/api.py
```diff
@@ -212,6 +213,36 @@ class Meta:
...
+            # we catch the one that may be raised by int() below and
+            # raise a customized BadRequest exception with more details.
+            try:
+                # Convert genes to a set of integers so that the
+                # duplicate(s) will be removed implicitly.
+                query_genes = {int(id) for id in heavy_genes.split(',')}
+            except ValueError:
+                raise BadRequest("Invalid gene IDs: %s" % heavy_genes)
+
+            for (i, q) in enumerate(query_genes):
```
cgreene: Instead of this loop, can this be `Participation.objects.filter(gene_id__in=query_genes).values_list('node_id', flat=True)`? You could make it a set if you wanted to.

dongbohu: But this will give the **union** (instead of **intersection**) of the nodes that are related to both genes, right?

cgreene: Indeed it will! Bad suggestion! Don't trust @cgreene!

adage/analyze/api.py
```diff
@@ -212,6 +213,36 @@ class Meta:
...
+                q_nodes = {p.node.id for p in
+                           Participation.objects.filter(gene=q)}
+                if i == 0:
+                    related_nodes = q_nodes
```
rzelayafavila: Interesting - I usually initialize variables like 'related_nodes' outside of a for-loop in Python, but I guess it doesn't really make a difference and actually saves a line of code to initialize the variable inside the loop, huh?

rzelayafavila:
Ok, I looked over this PR and I think it looks great overall @dongbohu! I am wondering about the failed CircleCI test - did you do a rebase in your repo with the changes @cgreene made for CircleCI?

adage/analyze/api.py
dongbohu: Since the **for** loop does not have its own scope (one of the features in C++/Java that I miss most when writing python), it doesn't really matter whether **related_nodes** is initialized inside or outside the loop.

dongbohu:
Thanks Rene. I will take a look at the failed CI test.

cgreene:
@dongbohu : failed CI test is because it's not configured in your master branch. The commit itself, when it happens, should go cleanly into the master at which point CI would run.

adage/analyze/api.py
```diff
@@ -212,6 +213,36 @@ class Meta:
...
+        heavy_genes = request.GET.get('heavy_genes', None)
+        if heavy_genes:
+            # Instead of relying on tastypie/resources.py to catch the
+            # ValueError exception that may be raised in this function,
+            # we catch the one that may be raised by int() below and
+            # raise a customized BadRequest exception with more details.
+            try:
+                # Convert genes to a set of integers so that the
+                # duplicate(s) will be removed implicitly.
+                query_genes = {int(id) for id in heavy_genes.split(',')}
```
mhuyck: Thank you for reminding me about Python sets! I could probably use these more often.

adage/analyze/api.py
```diff
@@ -255,8 +286,10 @@ def apply_filters(self, request, applicable_filters):
 
 
 class EdgeResource(ModelResource):
-    gene1 = fields.IntegerField(attribute='gene1_id', null=False)
-    gene2 = fields.IntegerField(attribute='gene2_id', null=False)
+    # gene1 = fields.IntegerField(attribute='gene1_id', null=False)
+    # gene2 = fields.IntegerField(attribute='gene2_id', null=False)
+    gene1 = fields.ForeignKey(GeneResource, "gene1", full=True)
+    gene2 = fields.ForeignKey(GeneResource, "gene2", full=True)
```
mhuyck: I like this approach. Having written some front-end views, I think we should make use of `full=True` more often where it makes sense.

adage/analyze/tests.py
```diff
@@ -226,7 +226,7 @@ def create_participations(num_nodes, num_genes):
         for i in range(num_genes):
             Gene.objects.create(entrezid=(i + 1),
                                 systematic_name="sys_name #" + str(i + 1),
-                                standard_name="sys_name #" + str(i + 1),
+                                standard_name="std_name #" + str(i + 1),
```
mhuyck: Nice catch!

dongbohu:
Thank you Matt.

PR #15, opened 2016-10-11
Pull request to add gene search functionality for gene network page

rzelayafavila:


cgreene:
This one is pretty angulary for me. I'm not going to be able to provide much (if any) useful feedback.

rzelayafavila:
@cgreene : Ok, gotcha - I'll remove you from the Assignees so that you don't get bugged by notifications (hopefully).

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+    .state( 'gene_search', {
+      url: '/gene_search',
+      views: {
+        "main": {
+          templateUrl: 'gene/gene-network.tpl.html',
+          controller: ['$scope', 'UserFactory',
+            function($scope, UserFactory) {
+              UserFactory.getPromise().$promise.then( function() {
+                $scope.userObj = UserFactory.getUser();
+              });
```
dongbohu: Do we need to provide a error function to then() too, or it is not necessary here?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+            delete selected_genes[geneObj.id];
+        },
+
+        return_genes: function () {
+            return selected_genes;
+        }
+    };
+})
+
+// Directive for whole gene search form
```
dongbohu: The indentation is not consistent in this section. Some blocks are using 2 spaces and the others are 4.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+        }
+      };
+
+      $scope.removeNotFound = function() {
+      // Function to get rid of all the queries that returned no results.
+          var searchResults = SearchResults.getSearchResults();
+          for (var key in searchResults) {
+            var results = searchResults[key];
+            if (results['found'].length === 0) {
+              SearchResults.remove(key);
```
dongbohu: Please note that when you delete a key **inside** for/in loop, the order of keys that will be listed by for/in loop may not be deterministic. In your case, the non-deterministic order of key doesn't seem to affect the correctness of your code, right?

interface/src/app/gene/genes.js
rzelayafavila: I think in this case it is not necessary because the UserFactory will return 'null' as the user object if no user is logged-in or something goes wrong with the request - so worst thing that can happen at this point is that $scope.userObj is set to null. What do you think @dongbohu ?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+        scope.loadingSearchResults = true;
+      });
+
+      scope.$on('results.searchResultsReturned', function() {
+        scope.loadingSearchResults = false;
+      });
+
+      //Watch for page changes and update
+      scope.$watch('currentPage', function() {
+        var begin = ((scope.currentPage-1)*scope.itemsPerPage);
```
dongbohu: A space character before and after "*" may make the code more readable.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+            if (results['found'][0]) {
+              var gene = results['found'][0];
+              SearchResults.remove(key);
+            }
+          }
+        }
+      };
+
+      $scope.removeNotFound = function() {
+      // Function to get rid of all the queries that returned no results.
```
dongbohu: Further indentation needed.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+              SearchResults.remove(key);
+            }
+          }
+      };
+    }],
+
+    link: function(scope, element, attr) {
+      scope.$on('results.update', function() {
+        scope.totalResults = SearchResults.size();
+        var begin = ((scope.currentPage-1)*scope.itemsPerPage);
```
dongbohu: Extra space characters before and after "-" and "*" maybe?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+      scope.$on('results.loadingSearchResults', function() {
+        scope.loadingSearchResults = true;
+      });
+
+      scope.$on('results.searchResultsReturned', function() {
+        scope.loadingSearchResults = false;
+      });
+
+      //Watch for page changes and update
+      scope.$watch('currentPage', function() {
```
dongbohu: Lines 203-205 seem identical to lines 188-190. If you want, you can define a separate function to reduce the redundancy. (since only three extra lines are involved, it is up to you.)

interface/src/app/gene/genes.js
rzelayafavila: Ah yes, good call, I will change this.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+      $scope.loadingSearchResults = false;
+
+      $scope.addAllNonAmbiguous = function() {
+      // Function to automatically add all genes that only have one search result
+        var searchResults = SearchResults.getSearchResults();
+        for (var key in searchResults) {
+          var results = searchResults[key];
+          if (results['found'].length <= 1) {
+            if (results['found'][0]) {
+              var gene = results['found'][0];
```
rzelayafavila: Mmm it is true - I will remove it.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+                    $rootScope.$broadcast( 'results.update' );
+                }
+                $rootScope.$broadcast( 'results.searchResultsReturned' );
+            });
+        }
+    };
+}])
+
+.factory( 'SelectedGenesFactory', function( ) {
+    var selected_genes = {};
```
dongbohu: It seems that all keys in **selected_genes** will be integers, is it better to define this variable as an array instead of regular object?

cgreene:
Seeing some style comments. Might be a good idea to enable codeclimate to check for these issues (indentation, etc). Needs a `.eslintrc` file in the root of the repository.

I'd encourage us to make as many checks that we agree on as possible automatic to avoid using person-time to find these issues :)

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+        scope: false, 
+        templateUrl: 'gene/previous-result-button.tpl.html'
+    };
+})
+
+// Directive for search buttonset, has buttons for handling
+// search results
+.directive('searchButtonset', function( SearchResults ) {
+    return {
+        controller: function( $scope ) {
```
dongbohu: Inline array annotation.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+            $scope.pageDict = {page: 1};
+            $scope.results = SearchResults.getQueryResults( $scope.query );
+            $scope.found = $scope.results.found;
+
+            var begin, end;
+            $scope.updatePage = function(page) {
+                begin = ((page-1)*3);
```
dongbohu: **begin = (page - 1) * 3;**

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+                end=begin+3;
```
dongbohu: **end = begin + 3;**

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+        controller: ['$scope', 'SelectedGenesFactory', '$state',
+          function($scope, SelectedGenesFactory, $state) {
+            $scope.selectedGenes = SelectedGenesFactory.return_genes();
+
+            $scope.sendToNetwork = function () {
+              var genes = [];
+              for (var gene in $scope.selectedGenes) {
+                genes.push(gene);
+              }
+              $state.go('gene_network', {'genes': genes});
```
rzelayafavila: So 1) Yes, I think we can pass an array as a parameter to the state. However, 2) $scope.selectedGenes is actually an object, where each key is the gene id and the value is the gene object, so I think I still need to build an array of gene ids here.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+    };
+})
+
+// Directive for button to get previous search results 
+.directive('previousResultButton', function( SearchResults ) {
+    return {
+        link: function(scope, element, attr) {
+            element.bind( "click", function() {
+                scope.page = scope.pageDict.page;
+                scope.page = scope.page - 1;
```
dongbohu: Can we merge lines 307-8 into one line like this?
```scope.page = scope.pageDict.page - 1;```

File: interface/src/app/gene/resource.js [hunk 15]
```diff
@@ -0,0 +1,22 @@
+angular.module( 'adage.gene.resource', ['ngResource'])
+
+.factory( 'Gene', ['$resource', function($resource) {
+  return $resource(
+    '/api/v0/gene/:id', {id: '@id'}, {
+
+      query: {
```
dongbohu: This is the default value of **query**, so it is not necessary, right?

interface/src/app/tribe_client/directives.js
```diff
@@ -77,4 +77,19 @@ angular.module('adage.tribe_client.directives', [
     };
 }])
 
+// Directive for whole gene search form
+.directive('genesetSearchForm', function( ) {
+    return {
+        controller: ['$scope', function($scope) {
```
dongbohu: This controller doesn't seem to do anything, can we remove it?

interface/src/app/gene/resource.js
rzelayafavila: For the most part - I guess the only difference is that isArray: is usually set to 'true'. However, I don't think we are using 'query' for now - I was including it in there in case we used it at some point, but I guess we can add it to the code if we need it, right?

interface/src/app/tribe_client/directives.js
rzelayafavila: Yes, for now, but I will reinsert it in the next pull request, when the geneset search is enabled.

cgreene: Code review isn't about making more work. But if it's not clear to someone that you're planning something, they aren't going to be able to know that when they review. :)

Even better - through a `TODO:` at the front of that comment and codeclimate will flag it for you automatically.

interface/src/app/gene/genes.js
rzelayafavila: Not really, because the selected_genes will hold the gene information as the value for each key (which will be the gene ID and that one will be an integer). I think it will be useful to keep the gene info. so that we can (for example) provide the standard gene name and a tooltip with the full name of the gene in this interface instead of just the gene ID.

rzelayafavila: @dongbohu do you think that is a good way to go about it?

interface/src/app/gene/genes.js
rzelayafavila: Ok, fixed it.

interface/src/app/gene/genes.js
dongbohu: Makes sense to me. Thank you for your reply.

interface/src/app/gene/genes.js
dongbohu: Yes, I forgot that **scope.selectedGenes** is actually an object. In my comment on line 85, I was actually asking whether it is possible to change **selected_genes** from object to an array.

interface/src/app/gene/genes.js
dongbohu: What I was thinking was actually setting **selected_genes** as an array whose index is the gene ID. I think you are right. An object type seems to be a better fit. 
(Anyway, in JavaScript, an array is internally an object, so no big deal.)

interface/src/app/gene/resource.js
dongbohu: Agree.

interface/src/app/tribe_client/directives.js
dongbohu: Either will be fine with me. Thanks.

interface/src/app/gene/genes.js
rzelayafavila: Right, exactly - in this case it does not matter, but it is definitely something to watch out for. Thanks for catching this!

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+              SearchResults.remove(key);
+            }
+          }
+        }
+      };
+
+      $scope.removeNotFound = function() {
+      // Function to get rid of all the queries that returned no results.
+          var searchResults = SearchResults.getSearchResults();
+          for (var key in searchResults) {
```
rzelayafavila: Ah, I see. Thanks for catching that!

interface/src/app/gene/genes.js
mhuyck: Good catch, @dongbohu. We could replace the anonymous function in both places with a variable holding the redundant implementation.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,377 @@
...
+        scope.resultsForPage = SearchResults.getQueries().slice(begin, end);
+      });
+
+      scope.$on('results.loadingSearchResults', function() {
+        scope.loadingSearchResults = true;
+      });
+
+      scope.$on('results.searchResultsReturned', function() {
+        scope.loadingSearchResults = false;
+      });
```
mhuyck: Okay, I see now where these signals are getting picked up. Maybe I'm missing something, but it feels non-Angularish to be using a `$broadcast` to set a variable like this. The `loadingSearchResults` variable should exist in one place as the primary model and any view that needs to access it should refer to that instance and let two-way binding handle the rest.

Another possibility is to use a chain of `$promise`s to achieve the sequence of updates required.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+      $scope.pageDict = {
+        page: 1
+      };
+      $scope.results = SearchResults.getQueryResults($scope.query);
+      $scope.found = $scope.results.found;
+
+      var begin, end;
+      $scope.updatePage = function(page) {
+        begin = (page - 1) * 3;
+        end = begin + 3;
```
mhuyck: The number `3` looks like a "magic number" here. What does it do?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+  }
+])
+
+// Directive for button to get more options, should get
+// next page of search results for this query from the server
+.directive('moreResultButton', ['SearchResults', function(SearchResults) {
+  return {
+    link: function(scope, element, attr) {
+      element.bind("click", function() {
+        scope.page = scope.pageDict.page + 1;
```
mhuyck: I'm confused about the difference between `scope.page` and `scope.pageDict.page`. Is there any reason this couldn't be simplified to `scope.page = scope.page + 1`? Maybe we need to track page in two separate variables under `scope`, but I don't follow why.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+        scope.$apply(function() {
+          scope.pageDict.page = scope.page;
+          scope.updatePage(scope.page);
```
mhuyck: What does `updatePage` do for us that hasn't already been handled with updates to `scope.page` and `scope.pageDict.page`? This doesn't feel Angular-ish.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+
+// Directive for button to get previous search results 
+.directive('previousResultButton', ['SearchResults', function(SearchResults) {
+  return {
+    link: function(scope, element, attr) {
+      element.bind("click", function() {
+        scope.page = scope.pageDict.page - 1;
+        scope.$apply(function() {
+          scope.pageDict.page = scope.page;
+          scope.updatePage(scope.page);
```
mhuyck: I have the same questions here as above with `moreResultButton`.

.eslintrc
```diff
@@ -0,0 +1,213 @@
+ecmaFeatures:
```
cgreene: @rzelayafavila : Once you extend google, do you need any of these, or are you now disabling things that are actually in accordance with google styles? For example, we're getting `no-empty-label` errors because that is being set, but doesn't exist anymore. Instead the new one is `no-labels`

cgreene: relevant google config link: 
https://github.com/google/eslint-config-google/blob/master/index.js

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+      $scope.searchResults = SearchResults.getSearchResults();
+      $scope.loadingSearchResults = false;
+
+      $scope.addAllNonAmbiguous = function() {
+        // Function to automatically add all genes that
+        // only have one search result.
+        var searchResults = SearchResults.getSearchResults();
+        for (var key in searchResults) {
+          var results = searchResults[key];
+          if (results['found'].length <= 1) {
```
dongbohu: JavaScript does not have **<==** or **>==** operators (surprise!). I think the reason behind it is that *<==** is equivalent to **NOT >**, and *>==** to **NOT <**. See more details here:

interface/src/app/gene/genes.js
rzelayafavila: @mhuyck : Aha, I see what you are saying. I thought this would be covered by the UserFactory setting the user to null by default (before calling the promise or anything else), but I see that the 'then' part of this code would not be executed, if, for example suddenly the server goes down (which would obviously break other things as well) or the tribe_client urls stopped working. However, the ng-if in the gene-network.tpl.html would resolve undefined (like null) to false, so it would also not show the geneset search form. I can, however, add on an error function (as @dongbohu was saying) to this to make sure the user is set to null if the promise totally breaks. Do you think this is a good approach?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+            return queries;
+        },
+        getQueryResults: function ( query ) {
+            return searchResults[query];
+        },
+        getSearchResults: function () {
+            return searchResults;
+        },
+        remove: function ( query ) { // Remove a query and its associated result
+            queries = queries.filter(function(el) { return el != query; });
```
rzelayafavila: Ok, changed that is now changed.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+        clear: function () { // Clear the service
+            queries = [];
+            searchResults = {};
+            $rootScope.$broadcast( 'results.update' );
+        },
+        size: function() {
+            return queries.length;
+        },
+        search: function ( qparams ) {
+          // Search for genes and add the results to the service
```
rzelayafavila: Changed.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+      $scope.currentPage = 1;
+      $scope.itemsPerPage = 10;
+      $scope.totalResults = 0;
+      $scope.maxSize = 10;
+      $scope.resultsForPage = [];
+      $scope.searchResults = SearchResults.getSearchResults();
+      $scope.loadingSearchResults = false;
+
+      $scope.addAllNonAmbiguous = function() {
+      // Function to automatically add all genes that only have one search result
```
rzelayafavila: Changed.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+    },
+    replace: true,
+    restrict: "E",
+    scope: true,
+    templateUrl: 'gene/search-result-table.tpl.html'
+    };
+}])
+
+// Directive for table containing search results
+.directive('selectedGenesPanel', function( ) {
```
rzelayafavila: Changed.

.eslintrc
rzelayafavila: Ok, done, and put up the info on setting up ESLint on Slack. Let me know if you have any other questions. Thanks for the PR with the fixed config, @cgreene!

interface/src/app/gene/genes.js
rzelayafavila: Fixed.

interface/src/app/gene/genes.js
rzelayafavila: Fixed.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,376 @@
...
+])
+
+// Directive for button to get more options, should get
+// next page of search results for this query from the server
+.directive('moreResultButton', function( SearchResults ) {
+    return {
+        link: function(scope, element, attr) {
+            element.bind( "click", function() {
+                scope.page = scope.pageDict.page;
+                scope.page = scope.page + 1;
```
rzelayafavila: Yep, fixed.

interface/src/app/gene/genes.js
rzelayafavila: Fixed.

interface/src/app/gene/resource.js
rzelayafavila: Changed.

interface/src/app/gene/gene-network.tpl.html
```diff
@@ -0,0 +1,10 @@
+<div>
+  <h1 class="page-header">
+    Gene Network
+  </h1>
+  <div class="well">
+    <geneset-search-form ng-if="userObj"></geneset-search-form>
```
rzelayafavila: Right - I was copying the behavior that the GIANT webserver uses for searching genes and/or Tribe genesets (the Tribe geneset section also only appears if the user is logged in using Tribe). I see what you mean though, so maybe it would be worth it to disable the section instead of removing it.

interface/src/app/gene/genes.js
mhuyck: Right, the `.then()` is called when the promise is fulfilled, but not if it's rejected. You could cover the other case by including a second callback function in `.then()`, by using a `.catch()` (as described in that promise write-up I found: https://developers.google.com/web/fundamentals/getting-started/primers/promises?hl=en ), or you could simply initialize `$scope.userObj` to `null` explicitly before the call to `UserFactory.getPromise()`. I think the last solution is the simplest, as it's a one-line fix and easy to follow.

interface/src/app/gene/gene-network.tpl.html
mhuyck: This isn't exactly a code issue, so it's up to you to decide if it's an issue worth addressing right now, later via an issue or maybe it's not actually an issue at all. I meant to be just asking the question while I had it on my mind. Until I see the actual screen in action, I can't give solid design review advice.

interface/src/app/gene/gene-search-form.tpl.html
```diff
@@ -0,0 +1,25 @@
...
+          <h4>Search for Genes</h4>
+          <form class="form">
+            <textarea name="genes" rows="7" ng-model="genesToAdd.query"
+              placeholder="Search genes by symbol or Entrez ID">
+            </textarea>
+            <span class="error">{{ errors.name }}</span>
+            <br>
+            <button class="btn btn-primary" ng-click="searchGenes()"
+              ng-disabled="loadingSearchResults">
+              Search   <i class="fa fa-search"></i>
```
rzelayafavila: Yes, thanks for catching that - having multiple spaces before the icon was just a typo - it'll be fixed in the next commit.
Yeah, for the indenting, that is what I chose to go with - I looked but didn't see anything in the Google style guide to discourage breaking the tag line like that (and the required indenting is 2 spaces).

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+          }
+          $scope.loadingSearchResults = true;
+
+          $rootScope.$broadcast('results.loadingSearchResults');
+          qparams = {
+            'query': $scope.genesToAdd.query
+          };
+
+          if ($scope.organism) {
+            qparams['organism'] = $scope.organism;
```
rzelayafavila: True. I just realized that we do not have multiple organism support yet for adage, so it does not make sense for this to be here yet. We can just add it if/when we add multi-organism support.

cgreene: We will... so if this is working you could:

`//TODO: adage doesn't currently support multiple organisms
$scope.organism = 'whatever indicates human'
`

rzelayafavila: Ok, added comments for that.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+// and remove entire row from list
+.directive('geneResultButton', ['SearchResults', 'SelectedGenesFactory',
+  function(SearchResults, SelectedGenesFactory) {
+    return {
+      restrict: "E",
+      link: function(scope, element, attr) {
+        element.bind("click", function() {
+          scope.$apply(function() {
+            SelectedGenesFactory.addGene(scope.gene);
+            SearchResults.remove(scope.query);
```
rzelayafavila: Right, once they add it here, it disappears from the search results.

interface/src/app/gene/gene-network.tpl.html
rzelayafavila: Ok, I will leave it as it is for the purposes of this PR, and if we decide that it would be more user friendly to change it, I think it will be a pretty easy change.

interface/src/app/gene/gene-search-form.tpl.html
```diff
@@ -0,0 +1,25 @@
...
+            </button>
+          </form>
+        </div>
+        <div class="col-xs-9">  
+          <search-result-table></search-result-table>
```
rzelayafavila: Ok, I think this works for now, but we can play around a little more with the layout and see if we want to change it afterwards.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,377 @@
+angular.module('adage.gene.search', [
```
rzelayafavila: Yes, I agree, I think that makes sense. I will change it to that.

interface/src/app/gene/genes.js
rzelayafavila: This is the number of gene results per 'page' in the button-set with the gene results. I will make a variable for this so that we can change it in the future if we want it to be something other than 3.

rzelayafavila:
Ok @mhuyck @dongbohu @cgreene : I think I have addressed all of the feedback for this pull request (let me know if I missed anything!) I'll keep thinking about the issue(s) @mhuyck brought up about making some parts of the code more Angularish and how to do this (I'll also open up an issue about this specifically in the gene/search.js file), but I think in the interest of time, we can go ahead and make sure everything else looks good and merge the PR.

interface/src/app/gene/search.js
```diff
@@ -0,0 +1,402 @@
...
+            }
+          }
+        };
+      }
+    ],
+
+    link: function(scope, element, attr) {
+      scope.$on('results.update', function() {
+        scope.totalResults = SearchResults.size();
+        var begin = ((scope.currentPage - 1) * scope.itemsPerPage);
```
dongbohu: Not sure whether you saw my previous comments. Lines 203-205 are identical to 218-220. Not sure whether it is worth the time to factor them out.

rzelayafavila: Ah yes - good catch, I will factor them out right away.

interface/src/app/tribe_client/directives.js
```diff
@@ -1,80 +1,90 @@
...
+              ]
+            });
 
-        }],
-        link: CommonFuncts.updateUser,
-        replace: true,
-        restrict: "E",
-        templateUrl: 'tribe_client/tribe-login-button.tpl.html'
+            modalInstance.result.then(function(success) {
+              if (success === true) {
```
dongbohu: Although I am a strong advocate of **===** operator, I am hesitant here: is **success** always a boolean? If it is an object, then this condition will be always false, because **===** doesn't involve any implicit type conversion.

dongbohu: Well, technically speaking, if an object defines its own **valueOf()** method that returns 1, then **obj == true** will be evaluated to true.

rzelayafavila:
@dongbohu and @mhuyck : Ok - I addressed those comments @dongbohu was talking about. Am I missing anything else in this PR?

.eslintrc
dongbohu: One more comment: you may want to rename this file into **.eslintrc.yaml** or **.eslintrc.yml**. According to:
http://eslint.org/docs/user-guide/configuring#configuration-file-formats
The name **.eslintrc** is deprecated. (I think it is a good idea, because without the extra extension, we have to guess what syntax this file is written in.)

rzelayafavila: Aha, good catch @dongbohu! I'll just change the name right now.

cgreene:
I commented on style because I wasn't able to comment well on the Angular. Codeclimate is now set up and configured, so 👍 from my point of view.

interface/src/app/gene/genes.js
mhuyck: I don't think you put in a fix for this one yet, did you?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,377 @@
...
+        clear: function () { // Clear the service
+            queries = [];
+            searchResults = {};
+            $rootScope.$broadcast( 'results.update' );
+        },
+        size: function() {
+            return queries.length;
+        },
+        search: function ( qparams ) {
+            // Search for genes and add the results to the service
```
mhuyck: Did I get this logic right?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,377 @@
...
+                      //if search term didn't already exist
+                      searchResults[ query ] = data[i]; //add it to the results
+                      queries.push(query);// add to the list of queries
+                    }
+                }
+                if (previousQueries != queries.length) {
+                    $rootScope.$broadcast( 'results.update' );
+                }
+                $rootScope.$broadcast( 'results.searchResultsReturned' );
+            });
```
mhuyck: This has not been addressed -- or am I missing where it was?

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+
+      $scope.addAllNonAmbiguous = function() {
+        // Function to automatically add all genes that
+        // only have one search result.
+        var searchResults = SearchResults.getSearchResults();
+        for (var key in searchResults) {
+          var results = searchResults[key];
+          if (results['found'].length <= 1) {
+            if (results['found'][0]) {
+              SearchResults.remove(key);
```
mhuyck: This `.remove` instead of `.add` is still confusing to me.

interface/src/app/gene/genes.js
rzelayafavila: Ok, fixed.

interface/src/app/gene/genes.js
```diff
@@ -0,0 +1,384 @@
...
+        });
+      },
+      replace: true,
+      templateUrl: 'gene/gene-result-button.tpl.html'
+    };
+  }
+])
+
+// Directive for button with a gene, should add gene
+// and remove entire row from list
```
mhuyck: Looks like you got this one.

interface/src/app/gene/genes.js
mhuyck: added to issue #17.

mhuyck:
I found a few places that seem to have been overlooked, expecting my comments to appear here but they aren't. I can see why you might have missed them, @rzelayafavila!

It appears that to see my comments you need to click the "View changes" button on the two reviews labeled "mhuyck requested changes." Not everything needs a code change, but I would at least like to understand the parts I noted just now.

interface/src/app/gene/genes.js
rzelayafavila: Yep! I made a few changes, but I will add that documentation in here.

interface/src/app/gene/genes.js
rzelayafavila: Hmm - well, the results need to be removed from the SearchResults once the user selects them. Oh, and this file is old (the new one is gene/search.js, as per our file name change discussion), which also has the ``SelectedGenesFactory.addGene(results.found[0]);`` line. Is this what is confusing?

mhuyck: Aha! Yep, I see that line now. So this version of the file does indeed have a bug, which is corrected via the update, right?

rzelayafavila: Yep, it was corrected before the name change.

interface/src/app/gene/genes.js
rzelayafavila: Ok, I pushed a couple of changes to address that.

rzelayafavila:
Ok @mhuyck I think I got the final few things you pointed out.

mhuyck:
Indeed you did. Glad it wasn't too tricky to work in that error handling. Thank you!

PR #18, opened 2016-10-19
Customize eslint rules based on google coding style, and integrate es…

dongbohu:
This PR includes:
* **eslint.google.js** (new file): a clone of Google's eslint rules at **eslint-config-google**. You do **NOT** need to review this file.
* **eslint.conf.js** (new file): includes a few coding style rules of our own. I added some comments in this file to explain what they mean. For more details, please check: http://eslint.org/docs/rules/
* **Gruntfile.js**: modified to integrate eslint into Grunt. Note that since most of the js files in our codebase don't conform to the rules yet, I didn't include **eslint** task into the default **grunt** command. So typing **grunt** won't run **eslint**. You have to type **grunt eslint** to launch it. (Once we clean up our codebase, I will add `eslint` task into `build` so that `grunt` command will launch it automatically.
* **package.json**: Removed **eslint** and **eslint-config-google** dependencies, added `grunt-eslint`.
* **build.config.js**: defined a variable for e2e test files.
* A few corrections of typos in **src/README.md** and **src/app/README.md**.  
* Since we are using grunt to take care of eslint, I removed **.eslintrc.yaml**.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -11,6 +11,8 @@ engines:
         javascript:
   eslint:
     enabled: true
+    config:
+      config: interface/eslint.conf.js
     channel: "eslint-3"
     exclude_paths:
     - "interface/Gruntfile.js"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -16,6 +16,7 @@ engines:
     channel: "eslint-3"
     exclude_paths:
     - "interface/Gruntfile.js"
+    - "interface/eslint.google.js"
   fixme:
     enabled: false
   radon:
--- a/interface/Gruntfile.js
+++ b/interface/Gruntfile.js
@@ -20,6 +20,7 @@ module.exports = function ( grunt ) {
   grunt.loadNpmTasks('grunt-html2js');
   grunt.loadNpmTasks('grunt-protractor-runner');
   grunt.loadNpmTasks('grunt-protractor-webdriver');
+  grunt.loadNpmTasks('grunt-eslint');
 
   /**
    * Load in our build configuration file.
@@ -323,6 +324,18 @@ module.exports = function ( grunt ) {
           src: [ '<%= app_files.coffeeunit %>' ]
         }
       }
+    },
+
+    /**
+     * eslint configuration.
+     */
+    eslint: {
+      options: {
+        configFile: 'eslint.conf.js'
+      },
+      src: ['<%= app_files.js %>'],
+      unit: ['<%= app_files.jsunit %>'],
+      e2e: ['<%= app_files.jse2e %>']
     },
 
     /**
--- a/interface/eslint.conf.js
+++ b/interface/eslint.conf.js
@@ -1,6 +1,6 @@
 /**
  * GreeneLab JavaScript ESLint rules, extends from Google coding guide at:
- * https://github.com/google/eslint-config-google/index.js (commit 40e3b15)
+ * https://github.com/google/eslint-config-google/index.js (commit 0037169)
  * (renamed to "eslint.google.js").
  */
 
--- a/interface/eslint.conf.js
+++ b/interface/eslint.conf.js
@@ -1,0 +1,31 @@
+/**
+ * GreeneLab JavaScript ESLint rules, extends from Google coding guide at:
+ * https://github.com/google/eslint-config-google/index.js (commit 40e3b15)
+ * (renamed to "eslint.google.js").
+ */
+
+module.exports = {
+  extends: './eslint.google.js',
+
+  globals: {
+    angular: true,
+    d3: true
+  },
+
+  // Customized rules.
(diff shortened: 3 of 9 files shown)


dongbohu:
About the new eslint config file **eslint.conf.js**, I avoided the default name **.eslintrc.js** because it would be hidden due to the leading dot character. If you prefer running **eslint** (instead of **grunt eslint**) on command line, you can either use **eslint -c eslint.conf.js** or set a shell alias (which is what I did). I also configured emacs editor so that the real-time check will load this file. I am sure the other editors are able to do something similar. 

That being said, if you guys prefer **.eslintrc.js**, I can change the name back.

cgreene:
Irony - it looks like google's file is failing its own lint checks:
`Found in interface/eslint.google.js by eslint`

Might want to make the lint files part of the exclude paths in codeclimate:
`     exclude_paths:`

dongbohu:
That is because of the trailing comma rule that we overwrote. Let me exclude this file.

dongbohu:
By the way, when eslint is completely integrated into Grunt, we probably should disable eslint in codeclimate to reduce some noise (and maintenance work).

dongbohu:
Added one more rule which could be controversial: always use **===** (instead of **==**) and **!==** (instead of **!=**). Please let me know if you object it. Surprisingly, Google's coding guide doesn't mention this rule at all.

interface/eslint.conf.js
```diff
@@ -20,12 +20,11 @@ module.exports = {
   // gNNN: Line number of the same option in ./eslint.google.js
   rules: {
     'curly': 2,                   // (g081) Always use curly braces.
+    'eqeqeq': [2, 'always'],      // (g086) Never allows '==' or '!=' operators.
```
cgreene: Agree that checking for this is a good idea. Comes up on PRs - nice to have automated :+1:

interface/Gruntfile.js
```diff
@@ -326,6 +327,18 @@ module.exports = function ( grunt ) {
     },
 
     /**
+     * eslint configuration.
+     */
+    eslint: {
```
rzelayafavila: Nice! It will be convenient to have grunt do this for us.

rzelayafavila:
Ok, this LGTM too! :+1:  As @cgreene commented, I was also wondering if vega should be added to the globals @mhuyck.

dongbohu:
One more thing I forgot to mention: **ngBoilerplace** includes another JavaScript linting tool called **jshint**. It has a few simple rules defined in **Gruntfile.js**. By default jshint is configured in ngbp so that if you start **grunt watch** task, any code change will trigger jshint automatically. I didn't integrate **eslint** into "grunt watch", because I don't think any of us is using "grunt watch". But if you really need it, please let me know and I will enable eslint in "grunt watch" task too.

dongbohu:
I reported a few minor issues in google's index.js and they have corrected them. So I synced our version too.

.codeclimate.yml
```diff
@@ -11,9 +11,12 @@ engines:
         javascript:
   eslint:
     enabled: true
+    config:
+      config: interface/eslint.conf.js
```
mhuyck: Is this the right `.yml` syntax here or a typo? Seems odd to have config twice like that.

interface/eslint.conf.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * GreeneLab JavaScript ESLint rules, extends from Google coding guide at:
+ * https://github.com/google/eslint-config-google/index.js (commit 0037169)
```
mhuyck: Thank you for noting the commit! 👍

interface/src/app/README.md
```diff
@@ -46,9 +46,8 @@ angular.module( 'ngBoilerplate', [
   'templates-app',
   'templates-common',
   'ngBoilerplate.home',
-  'ngBoilerplate.about'
-  'ui.router',
-  'ui.route'
```
mhuyck: Weird. Does `ui.route` even exist? Was this a typo in ngBoilerplate?

mhuyck:
Thank you, @cgreene and @rzelayafavila, for noting that we'll probably also need `vega` as a global. Let's leave it out here and I'll include it in a separate pull request. I'm kinda curious to see what ESLint says about it!

.codeclimate.yml
dongbohu: It does seem weird. I copied the syntax from:
https://docs.codeclimate.com/docs/eslint
The first **config** may include other options, such as "extensions", the second **config** is used to specify the configuration file's path.

interface/src/app/README.md
mhuyck: Sharp eyes!

.codeclimate.yml
mhuyck: That explains it then. Thanks!

PR #21, opened 2016-10-21
Enforce a rule that does not allow whitespace character inside "( )"

dongbohu:
A minor PR with a new linting rule.


dongbohu:
Codeclimate will probably report some new errors.

mhuyck:
Yep. Multiple errors per line... looks like codeclimate reports a separate error for each "offending" paren.

PR #2, opened 2016-09-01
Add .gitignore, remove .hgignore

dongbohu:
This simple PR includes a **.gitignore** file that is based on the original **.hgignore**.

PR #22, opened 2016-10-21
Adding basic functionality for geneset search (from Tribe genesets)

rzelayafavila:


interface/src/app/gene/gene-network.tpl.html
```diff
@@ -3,8 +3,19 @@ <h1 class="page-header">
     Gene Network
   </h1>
   <div class="well">
-    <geneset-search-form ng-if="userObj"></geneset-search-form>
```
cgreene: Did you accidentally convert from this approach into the other where elements are included directly in the page? This one feels more reusable.

interface/src/app/gene/search.js
```diff
@@ -10,12 +10,30 @@ angular.module('adage.gene.search', [
       views: {
         "main": {
           templateUrl: 'gene/gene-network.tpl.html',
-          controller: ['$scope', 'UserFactory',
-            function($scope, UserFactory) {
+          controller: ['$scope', 'UserFactory', '$uibModal',
```
cgreene: What is this `$uibModal`? I probably just don't understand enough angular.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,88 @@
...
+    getQuery: function() {
+      return query;
+    },
+    getGenesets: function() {
+      return genesets;
+    },
+    clear: function() {
+      query = {};
+      genesets = [];
+      $rootScope.$broadcast('genesets.update');
```
cgreene: I now know that this isn't very angularish from what I've heard from @mhuyck.

interface/src/app/tribe_client/tribe_client.js
```diff
@@ -1,42 +1,45 @@
...
-    'adage.tribe_client.directives',
+  'adage.tribe_client.resource',
+  'adage.tribe_client.directives',
+  'adage.tribe_client.genesets'
 ])
 
-.factory( 'UserFactory', ['User', function( User ) {
-    var promise = null;
-    var user = null;
+.factory('UserFactory', ['User', function(User) {
```
cgreene: Everything below this point looked like whitespace changes if I read it correctly.

interface/src/app/gene/gene-network.tpl.html
rzelayafavila: I agree that having one directive instead of a bunch of code is more reusable. I must say, though, that the code that I added underneath is not for the geneset search form itself, but just to go to the geneset search form. Currently, this means opening up a modal where the geneset search form will be displayed. Do you think this approach makes sense?

interface/src/app/tribe_client/genesets.js
rzelayafavila: True - I will find once and for all a way to do this in an Angular way.

interface/src/app/tribe_client/tribe_client.js
rzelayafavila: Yep, exactly, re-formatting thanks to our new JS standard. I almost did not include this file in the pull request, but decided to do so because of the injecting of 'adage.tribe_client.genesets' above, which is necessary for this pull request.

interface/src/app/gene/search.js
```diff
@@ -10,12 +10,30 @@ angular.module('adage.gene.search', [
...
               });
+
+              $scope.searchTribe = function() {
+                var modalInstance = $uibModal.open({
+                  templateUrl: 'tribe_client/geneset-search-modal.tpl.html',
+                  controller: ['$scope', function($scope) {
+                    // TODO: Right now, we are hard-coding this organism
+                    // as Pseudomonas (since it is the only one currently
+                    // supported by ADAGE). However, as we incorporate
+                    // multi-species support, this organism will havt to
```
dongbohu: Typo: ... will **have** to ...

interface/src/app/gene/search.js
```diff
@@ -10,12 +10,30 @@ angular.module('adage.gene.search', [
...
+                    // be obtained from the ML model. This is the same as
+                    // the issue in geneSearchForm (also with $scope.organism).
+                    $scope.organism = 'Pseudomonas aeruginosa';
+                  }]
+                });
+
+                modalInstance.result.then(function(standards) {
```
dongbohu: Maybe you want to add some "TODO" comments inside this blank function body?

interface/src/app/gene/gene-network.tpl.html
mhuyck: I always prefer modeless over modal, if possible. This reduces clicks and makes the features more readily accessible to the user. Sometimes modal "dialog boxes" are unavoidable, but in most cases you can design a non-modal way of doing it that will be better.

In this particular case, what if the page has a few scattered terms the user wants to copy and paste into the search box? They can't do that if the input field is inaccessible in a modal.

interface/src/app/tribe_client/geneset-result-table.tpl.html
```diff
@@ -0,0 +1,14 @@
+<div>
+    <table class="table table-striped" name="searchResults">
+        <tr><th>Title</th><th>Species</th><th>Creator</th></tr>
```
mhuyck: Minor quibble: if the table header were formatted the same as the table rows below, it would be easier to see which heading goes with which column.

interface/src/app/tribe_client/genesets.js
mhuyck: It should be possible to have Angular do this cross-scope communication for you if you bind `genesets` to a relevant `$scope` in the DOM. Anyone who is currently configured to "receive" the `$broadcast` will automatically update via two-way binding.

This means a change to the `scope` parameter in the relevant `.directive`. The parent view that invokes the directive will also need to be updated to pass in the variable. I think of this like parameters to a function, only the "function" happens to be a visible widget. A place where I have implemented something like this is the `results` parameter in `interface/src/app/analyze/search/search.js`, with invoking parent in `.../analyze/analyze.tpl.html`. See if that example would fit this code.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,88 @@
...
+    },
+    totalResults: function() {
```
mhuyck: I agree that `totalResults` sounds a bit like you're going to get all (the *totality*) of the GeneSet results. In addition, "total" could sound like a verb here, as if it's a calculation, but this is just an accessor. Perhaps a minor rework would allow this to be called `getResultCount()` and the corresponding attribute could be called `resultCount`?

mhuyck: Oh... another refinement: `resultCount` should probably be initialized to `null` so that calling `getResultCount()` before `query()` will return `null` (meaning no query yet) as opposed to `0`, which could mean either "no query yet" *or* "query returned 0 results".

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,88 @@
...
+        search['organism__scientific_name'] = $scope.organism;
+        GenesetSearch.query(search);
+      };
+    }],
+    replace: true,
+    restrict: "E",
+    scope: {
+      limit: '=limit',
+      organism: '='
+    },
```
mhuyck: Awesome. You are already using an isolate scope, so this is where you add a parameter for your `results` to get passed back to the parent... ☆

interface/src/app/tribe_client/geneset-search-modal.tpl.html
```diff
@@ -0,0 +1,15 @@
+<div class="modal-header">
+    <geneset-search-form organism="organism"><geneset-search-form>
```
mhuyck: ☆ ...and this is where you will need to bind the new `results` parameter (see `.directive`) to the parent's `scope`.

mhuyck: Only we want this not to be in a modal, but a part of the view. That change will also simplify managing the scope (which is tricky to get right in modals).

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,88 @@
...
+    totalResults: function() {
+      return totalResults;
+    },
+    query: function(searchParams) {
+      query = searchParams;
+      Genesets.query(query, function(data) {
+        totalResults = data.meta.total_count;
+        genesets = data.objects;
+        $rootScope.$broadcast('genesets.update');
```
mhuyck: Once `genesets` is bound to the parent's scope via the `.directive()` modifications I suggested, this `$broadcast` can be eliminated.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,88 @@
...
+      $scope.genesets = GenesetSearch.getGenesets();
+
+      var query = GenesetSearch.getQuery();
+      query['limit'] = $scope.limit;
+      query['show_tip'] = true;
+      query['organism__scientific_name'] = $scope.organism;
+      GenesetSearch.query(query);
+    }],
+    link: function(scope, element, attr) {
+      scope.$on('genesets.update', function() {
```
mhuyck: Once `genesets` is bound to the parent's scope via the `.directive()` modifications I suggested, this `$on` listener can be eliminated. Perhaps you can also get your count of `totalResults` from something like `scope.genesets.length` so we don't have a proliferation of parameters. If you do need access to `totalResults()`, you can also set `scope.totalResults = GenesetSearch.totalResults;` directly in your `controller` above and it will work.

mhuyck: Oh, it will work, but with a slight difference: where you reference `scope.totalResults` in your page would need to become a function call: `scope.totalResults()`.

interface/src/app/tribe_client/resource.js
```diff
@@ -1,12 +1,33 @@
...
+}])
+
+.factory('Genesets', ['$resource', function($resource) {
+  return $resource(
+    'https://tribe.greenelab.com/api/v1/geneset/:creator/:slug/', {
+      creator: '@creator',
+      slug: '@slug'
+    },
+
+    {
```
mhuyck: You have two different styles here: above, at line 16, it says

`param1, {`

where `{` is the beginning of `param2`, and here you have

```
},

{
```

where `}` is the end of `param2` and `{` is the beginning of `param3`. Maybe ESLint is trying to enforce something here, but if so I think it's doing it wrong. The style should be consistent.

interface/src/app/tribe_client/tribe_client.js
mhuyck: Thanks for noting. I skipped the rest of this file.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Ok, thanks @mhuyck! This makes sense. I also saw the comments in the directives below, where these genesets could be passed between the relevant scopes and the code you mentioned in the analyze module, and am working on restructuring this code.

dongbohu:
@rzelayafavila: I have a few more thoughts on this PR:

(1) To make the gene search web UI consistent with the gene network interface, maybe you can show the gene's standard name (if available) in the matched gene results bar; if standard name is not available, show systematic name instead. (This is what I did to label the genes in the network.)

(2) Since you are working on **search.js** now, can you remove the following lines from search.js?
```
.state('gene_network', {
    // This is a placeholder for the Gene Network page that
    // Dongbo is working on, which may or may not live in this file.
    // For the moment, it is just the state where users go to after
    // they have selected their genes.
    url: '/gene_network/',
    data: {
      pageTitle: 'Gene Network'
    }
  });
```
My original plan was to have my gene network code share the same module with your gene search code. But since the gene search module is much more complex than I thought, I don't think it is a good idea to couple my code with yours (at least not now, since both parts of the code are in active development).

(3) In the following code of **search.js**:
```
// Directive for table containing search results
.directive('selectedGenesPanel', function() {
  return {
    controller: ['$scope', 'SelectedGenesFactory', '$state',
      function($scope, SelectedGenesFactory, $state) {
        $scope.selectedGenes = SelectedGenesFactory.returnGenes();

        $scope.sendToNetwork = function() {
          var genes = [];
          for (var gene in $scope.selectedGenes) {
            genes.push(gene);
          }
          $state.go('gene_network', {
            'genes': genes
          });
        };
      }
    ],
    replace: true,
    restrict: "E",
    scope: true,
    templateUrl: 'gene/selected-genes-panel.tpl.html'
  };
})
```
Suppose a user selects two genes whose IDs are 123, 456, then **sendToNetwork()** function is supposed to direct the page to: 
**/gene_network/?genes=123,456**
but it actually directed to the following page:
**/gene_network/?genes=123&genes=456**
Can you fix it too? 

Thanks.

rzelayafavila:
Hi @dongbohu : Thanks for your comments. I think most of those things do not pertain quite to this pull request though. They are mostly on the side of doing the gene search instead of the geneset search. I am almost done making the necessary changes to finish up this pull request, so I can make another pull request for the improvements you are talking about in the gene search code once this pull request is merged (to keep all changes relating to different features in different pull requests).

However, your point #2 is very doable in this pull request - I will just go ahead and delete that state from the ``gene/search.js`` file and handle the other changes in the future pull request regarding gene search.

dongbohu:
@rzelayafavila: No problem. Thank you!

interface/src/app/tribe_client/geneset-search-form.tpl.html
```diff
@@ -1,22 +1,14 @@
...
+    <h4>
+      Have a Tribe gene set you want to use?
+    </h4>
+  </div>
+
+  <div class="panel-body">
+    <label for="title">Search Tribe gene sets</label>
+    <geneset-search-bar organism="organism" limit="limit" genesets="genesets"></geneset-search-bar>
+    <hr>
+    <gene-set-table limit="limit" genesets="genesets"></gene-set-table>
```
mhuyck: One more quibble, which should be an easy fix: here you have `gene-set-` but everywhere else you have `geneset-`. Shouldn't the naming be consistent?

rzelayafavila:
Ok @mhuyck, @dongbohu - I made the changes referring to the last comment, about the name of the old geneSetTable directive.

However, I am actually still struggling a little bit with the directives and the passed scope. It looks like the genesets *are* being passed and shared between the different directives, but they only update after the *second* time that the 'Search' button is clicked. Therefore, a user needs to enter a query into the box, and then click the 'Search' button twice for the search results to update. I am looking into this, and my hunch is that I must add some kind of watcher, or make a $digest cycle run after the first 'Search' click so that the genesets update between directives, but would you all have any ideas about this? Thanks!

mhuyck:
Hmm. That is odd, and definitely not the behavior I see with the `Sample`/`Experiment` search I implemented. I'll take another look at the code...

interface/src/app/tribe_client/geneset-search-bar.tpl.html
```diff
@@ -0,0 +1,12 @@
+<div>
+  <form class="form-inline">
+    <label>Organism:</label>
+    <select ng-disabled="true">
+      <option ng-selected="true" value="organism">{{organism}}</option>  
+    </select>
+
+    <input ng-model="search.query" placeholder="Search text">
+
+    <button class="btn btn-default" ng-click="searchGenesets(search)">Search</button>
```
mhuyck: This line invokes `searchGenesets(search)`, but have a look at the implementation of that function... ☆

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,97 @@
...
+// Search box for users to enter their search text into
+// and go search Tribe genesets.
+.directive('genesetSearchBar', ['GenesetSearch', function(GenesetSearch) {
+  return {
+    controller: ['$scope', function($scope) {
+      $scope.search = {};
+      $scope.search['organism__scientific_name'] = $scope.organism;
+      $scope.search['limit'] = $scope.limit;
+
+      $scope.searchGenesets = function() {
```
mhuyck: ☆ ...here. See how this function takes no parameter? What you are passing in gets ignored.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,97 @@
...
+        GenesetSearch.query($scope.search);
```
mhuyck: You might think this line would make up for that, but you have an isolate scope, which means the `genesetSearchBar` cannot see the `.search` you set in the template (parent scope) because you haven't passed it in.

I believe this is why it's not quite working right. Sorry I missed that in review earlier!

mhuyck: This might not be perfectly correct advice, but I believe it's on track. Have a look here at what's going on with scopes in more detail: https://docs.angularjs.org/guide/directive#isolating-the-scope-of-a-directive

rzelayafavila: Ahh, I see what is going on @mhuyck. It looks like the $scope.search is getting set correctly, since it is getting it from the same template it is bound to, not the parent one (I think I should definitely remove the ``search`` parameter from the searchGenesets() call in the template though). I think what is happening though is that the genesets aren't being returned by the query until *after* I call the GenesetSearch.getGenesets() (therefore, at the time when I do, this still returns an empty list). I should wait for the query to finish, and *then * ask to get the genesets back from the factory. I'll check to see if this approach works.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,97 @@
...
+      query = searchParams;
+      Genesets.query(query, function(data) {
+        resultCount = data.meta.total_count;
+        genesets = data.objects;
+      });
+    }
+  };
+}])
+
+.directive('genesetSearchForm', ['GenesetSearch', function(GenesetSearch) {
```
mhuyck: 'GenesetSearch' no longer needs to be injected here. You aren't using it in the controller and it's not referenced in the template.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,97 @@
...
+      limit: '=',
+      organism: '=',
+      genesets: '='
+    },
+    templateUrl: 'tribe_client/geneset-search-bar.tpl.html'
+  };
+}])
+
+// Directive for table containing search results
+.directive('genesetResultTable', ['GenesetSearch', function(GenesetSearch) {
```
mhuyck: `GenesetSearch` does not need to be referenced here, either. I don't see it used in the controller code or in the template.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Yep, I'll remove it from here and from genesetResultTable. Thanks!

interface/src/app/tribe_client/genesets.js
mhuyck: Okay... sounds like you're on the right track. My explanation seemed a bit off since it couldn't explain why a *second* click worked.

If you're still stuck, I think I've figured out how I can pull a copy of the change sets from your fork to work with on my development instance so I can trace through it myself. I'm heading off to an afternoon with violin lesson and pediatrician appointment, so can't follow up until the evening, probably. Please let me know if I can do any more to help!

rzelayafavila:
Ok, it works! It lives!! What I ended up doing was basically removing the GenesetSearchFactory in its entirety (since it really was not offering us any more functionality than the Genesets resource now that we no longer use the $broadcast system), and just waiting for the query to finish before assigning that to the $scope.genesets. Thanks @mhuyck for your help! Let me know if I got all of those last changes you mentioned!

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,97 @@
...
+.directive('genesetSearchBar', ['GenesetSearch', function(GenesetSearch) {
+  return {
+    controller: ['$scope', function($scope) {
+      $scope.search = {};
+      $scope.search['organism__scientific_name'] = $scope.organism;
+      $scope.search['limit'] = $scope.limit;
+
+      $scope.searchGenesets = function() {
+        GenesetSearch.query($scope.search);
+        $scope.genesets = GenesetSearch.getGenesets();
```
mhuyck: Do this assignment in the code that executes when the promise is fulfilled and you should be right as rain.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,97 @@
+angular.module('adage.tribe_client.genesets', [
+  'adage.tribe_client.resource'
+])
+
+// Geneset Search Factory
+// Allows for retreiving for and interacting with search results for genesets.
+.factory('GenesetSearch', ['Genesets', function(Genesets) {
+  var genesets = [];
```
mhuyck: Wait a second.. I may have a simplification for you here.

The trouble we're having is from passing around copies of `genesets`, right? The *Angularish* solution should be to simply reference the One True `genesets` wherever you need to, and two-way binding will handle the rest.

The problem here is that the One True `genesets` is locked up in this closure variable inside the function that implements the `GenesetSearch` service. What if, instead, these `var`s were moved into the object that gets returned (line 12 below)? Then you can bind `GenesetSearch.genesets` directly wherever it's needed.

For an example of how I've done that, refer to the `SampleBin` implementation in `interface/src/app/analysis/sampleBin.js`.

mhuyck: Hah. I did indeed just now see your "it lives!" message. In part, GitHub was being sluggish to show your replies. The other part is stepping away for the afternoon of kid activities I mentioned just before I dashed off this "wait a second..." note.

Glad you've sorted it out! I'm looking over what you did now.

interface/src/app/tribe_client/genesets.js
```diff
@@ -2,38 +2,7 @@ angular.module('adage.tribe_client.genesets', [
...
-      query = searchParams;
-      Genesets.query(query, function(data) {
-        resultCount = data.meta.total_count;
-        genesets = data.objects;
-      });
-    }
-  };
-}])
-
-.directive('genesetSearchForm', ['GenesetSearch', function(GenesetSearch) {
```
mhuyck: Good call on just taking out the `GenesetSearch` service entirely. I guess it could have been a clue for me that I was pointing out all the places you were no longer using it... :-)

mhuyck:
Okay, this approach totally makes sense now. You know you did something right when you fixed a bug by removing a big chunk of code you don't need. Very satisfying.

I have one more area to ask about, now that I can see how things are fitting together better.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+}])
+
+// Directive for table containing search results
+.directive('genesetResultTable', [function() {
+  return {
+    replace: true,
+    restrict: 'E',
+    scope: {
+      limit: '=',
```
mhuyck: Also: I see the `limit` getting passed in to the search above and here in the `genesetResultTable` directive, but have you tested what happens if your list of geneset results exceeds this limit? Did you mean to implement pagination in this table?

If pagination is planned but not implemented yet, I'd call that out of scope for this PR, but a TODO comment in the code would help clarify the intent.

interface/src/app/tribe_client/genesets.js
```diff
@@ -0,0 +1,64 @@
+angular.module('adage.tribe_client.genesets', [
+  'adage.tribe_client.resource'
+])
+
+.directive('genesetSearchForm', [function() {
+  return {
+    controller: ['$scope', function($scope) {
+      // Number of results (in this case genesets) to come back
+      // in each response page.
+      $scope.limit = 10;
```
rzelayafavila: Ah good catch. I agree - I think this should probably get set in the genesetSearchForm directive controller, since that is where all the business logic for this directive is handled. I will remove the setting of this ``limit`` parameter wherever this directive gets called.

interface/src/app/tribe_client/genesets.js
rzelayafavila: True, I put this in here exactly for the case when the resultCount was larger than the limit, and so that we could implement pagination down the road. I will add a TODO comment for this.

rzelayafavila:
Ok @mhuyck - I have made the couple more changes regarding the 'limit' parameter. Let me know how it looks!

mhuyck:
Looks good. Thanks for all the hard work on this!

PR #24, opened 2016-10-26
Gene-Gene Network Web UI

dongbohu:
This is a relatively heavy pull request. It includes all files related to the new gene-gene network web UI. Right now there are still a few minor issues related to Aaron Wang's **d3.network**. Instead of waiting for him to revise it, I decide to put my current code in PR first.

You may notice a few **FIXME** tags in some files. These are issues that I am still working on. If you happen to know how to solve them, I would love to hear your opinion. 

All revisions in **app.spec.js** are because of eslint coding styles. You don't have to review it.

I put all gene-gene network related files in **gene/network/** to avoid possible file overwriting, because @rzelayafavila is working in the same directory right now. In the long run, it may be a good idea to put all gene-search related code into a sub-directory too.

### DIFF
--- a/interface/build.config.js
+++ b/interface/build.config.js
@@ -74,6 +74,10 @@ module.exports = {
       'vendor/angular-sanitize/angular-sanitize.js',
       'vendor/angular-ui-router/release/angular-ui-router.js',
       'vendor/ng-sortable/dist/ng-sortable.min.js',
+      'vendor/jquery/dist/jquery.min.js',              // why no effect?
+      'vendor/jquery-ui/ui/minified/jquery-ui.min.js', // why not effect?
+      'vendor/angular-ui-slider/src/slider.js',
+      'vendor/d3-tip/index.js',
       'vendor/placeholders/angular-placeholders-0.0.1-SNAPSHOT.min.js'
     ],
     css: [
--- a/interface/build.config.js
+++ b/interface/build.config.js
@@ -75,7 +75,7 @@ module.exports = {
       'vendor/angular-ui-router/release/angular-ui-router.js',
       'vendor/ng-sortable/dist/ng-sortable.min.js',
       'vendor/jquery/dist/jquery.min.js',              // why no effect?
-      'vendor/jquery-ui/ui/minified/jquery-ui.min.js', // why not effect?
+      'vendor/jquery-ui/ui/minified/jquery-ui.min.js', // why no effect?
       'vendor/angular-ui-slider/src/slider.js',
       'vendor/d3-tip/index.js',
       'vendor/placeholders/angular-placeholders-0.0.1-SNAPSHOT.min.js'
--- a/interface/build.config.js
+++ b/interface/build.config.js
@@ -64,6 +64,10 @@ module.exports = {
    */
   vendor_files: {
     js: [
+      // For some unknown reason(s), jquery and jquery-ui library files MUST be
+      // put at the beginning of this array; Otherwise the ui-slider wouldn't
+      // show up on the web UI, and we got the following error in the console:
+      // "TypeError: d.slider is not a function".
       'vendor/jquery/dist/jquery.min.js',
       'vendor/jquery-ui/ui/minified/jquery-ui.min.js',
       'vendor/d3/d3.js',
--- a/interface/build.config.js
+++ b/interface/build.config.js
@@ -64,12 +64,6 @@ module.exports = {
    */
   vendor_files: {
     js: [
-      // For some unknown reason(s), jquery and jquery-ui library files MUST be
-      // put at the beginning of this array; Otherwise the ui-slider wouldn't
-      // show up on the web UI, and we got the following error in the console:
-      // "TypeError: d.slider is not a function".
-      'vendor/jquery/dist/jquery.min.js',
-      'vendor/jquery-ui/ui/minified/jquery-ui.min.js',
       'vendor/d3/d3.js',
       'vendor/vega/vega.js',
       'vendor/angular/angular.js',
@@ -80,11 +74,12 @@ module.exports = {
       'vendor/angular-sanitize/angular-sanitize.js',
       'vendor/angular-ui-router/release/angular-ui-router.js',
       'vendor/ng-sortable/dist/ng-sortable.min.js',
-      'vendor/angular-ui-slider/src/slider.js',
+      'vendor/angularjs-slider/dist/rzslider.min.js',
       'vendor/d3-tip/index.js',
       'vendor/placeholders/angular-placeholders-0.0.1-SNAPSHOT.min.js'
     ],
     css: [
+      'vendor/angularjs-slider/dist/rzslider.min.css'
     ],
     assets: [
     ]
--- a/interface/build.config.js
+++ b/interface/build.config.js
@@ -64,6 +64,8 @@ module.exports = {
    */
   vendor_files: {
     js: [
+      'vendor/jquery/dist/jquery.min.js',
+      'vendor/jquery-ui/ui/minified/jquery-ui.min.js',
       'vendor/d3/d3.js',
       'vendor/vega/vega.js',
       'vendor/angular/angular.js',
@@ -74,8 +76,6 @@ module.exports = {
       'vendor/angular-sanitize/angular-sanitize.js',
       'vendor/angular-ui-router/release/angular-ui-router.js',
       'vendor/ng-sortable/dist/ng-sortable.min.js',
-      'vendor/jquery/dist/jquery.min.js',              // why no effect?
-      'vendor/jquery-ui/ui/minified/jquery-ui.min.js', // why no effect?
       'vendor/angular-ui-slider/src/slider.js',
       'vendor/d3-tip/index.js',
       'vendor/placeholders/angular-placeholders-0.0.1-SNAPSHOT.min.js'
--- a/interface/src/app/app.spec.js
+++ b/interface/src/app/app.spec.js
@@ -1,17 +1,17 @@
-describe( 'AppCtrl', function() {
-  describe( 'isCurrentUrl', function() {
+describe('AppCtrl', function() {
+  describe('isCurrentUrl', function() {
     var AppCtrl, $location, $scope;
 
-    beforeEach( module( 'adage' ) );
+    beforeEach(module('adage'));
 
-    beforeEach( inject( function( $controller, _$location_, $rootScope ) {
+    beforeEach(inject(function($controller, _$location_, $rootScope) {
       $location = _$location_;
       $scope = $rootScope.$new();
-      AppCtrl = $controller( 'AppCtrl', { $location: $location, $scope: $scope });
+      AppCtrl = $controller('AppCtrl', {$location: $location, $scope: $scope});
     }));
 
-    it( 'should pass a dummy test', inject( function() {
-      expect( AppCtrl ).toBeTruthy();
+    it('should pass a dummy test', inject(function() {
+      expect(AppCtrl).toBeTruthy();
     }));
   });
 });
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -37,6 +37,21 @@ angular.module('adage.gene.network', [
       var rawMinWeight = -1.0;
       var rawMaxWeight = 1.0;
       var scaledMaxWeight = rawMaxWeight - rawMinWeight;
+      var setEdgeColor = d3.scale.linear()
+          .domain([0, scaledMaxWeight / 2.0, scaledMaxWeight])
+          .range(['green', 'orange', 'red']);
+
+      var network = d3.network()  // Initialize the network.
+          .minEdge(0)
+          .maxEdge(rawMaxWeight - rawMinWeight)
+          .edgeColor(setEdgeColor)
+          .geneText(function(d) {
+            return d.label;
+          })
+          .legendStart(rawMinWeight)
+          .legendEnd(rawMaxWeight)
+          .legendText('Correlation');
+
       var self = this;
       // The following properties of "self" will be available to HTML.
       self.minCorrelation = -1.0;
@@ -57,35 +72,8 @@ angular.module('adage.gene.network', [
       }
 
       var urlGenes = $stateParams.genes.split(',');
-      var setEdgeColor = d3.scale.linear()
-          .domain([0, scaledMaxWeight / 2.0, scaledMaxWeight])
-          .range(['green', 'orange', 'red']);
-
       var genes = [];
       var edges = [];
-      var network = d3.network()  // Initialize the network.
-          .minEdge(0)
-          .maxEdge(rawMaxWeight - rawMinWeight)
-          .edgeColor(setEdgeColor)
-          .geneText(function(d) {
-            return d.label;
-          })
-          .legendStart(rawMinWeight)
-          .legendEnd(rawMaxWeight)
-          .legendText('Correlation');
-
-      // Initialize the SVG.
-      var svg = d3.select('#chart').append('svg');
-      var minSvgSize = 600;  // Minimum size of svg.
-      var maxSvgSize = 1280; // Maximum size of svg.
-
-      // Initialize tips on gene and edge.
-      var geneTip = d3.tip()
-          .attr('class', 'gene-tip')
-          .offset([-20, 0]);
-      var edgeTip = d3.tip()
-          .attr('class', 'edge-tip')
-          .offset([-20, 0]);
 
       /**
        * Find index of input gene ID in genes array.
@@ -119,7 +107,7 @@ angular.module('adage.gene.network', [
             geneIn.query = true;
           }
           geneIn.label = (geneIn.standard_name ? geneIn.standard_name
-                           : geneIn.systematic_name);
+                          : geneIn.systematic_name);
           genes.push(geneIn);
           idx = genes.length - 1;
         }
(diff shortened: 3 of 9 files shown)


interface/src/index.html
```diff
@@ -15,12 +15,29 @@
...
+    </script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.4/jquery-ui.min.js">
+    </script>
+
     <!-- compiled JavaScript --><% scripts.forEach( function ( file ) { %>
     <script type="text/javascript" src="/static/<%= file %>"></script><% }); %>
 
+    <!-- d3.network library, written by Aaron Wang (aaronkw) -->
+    <script src="https://rawgit.com/dongbohu/d3-network/master/d3.network.js"
+            charset="utf-8"></script>
```
mhuyck: It should be possible to include this in the `build.confi.js` under `vendor_files`. That way the code will get minified and bundled up with the rest of our code. Did you try that?

mhuyck:
Just a couple of quick notes so far, responding to your `FIXME` comment. I'll have to set aside some time to go over the bulk of this because it is going to take some time.

interface/src/index.html
```diff
@@ -15,12 +15,29 @@
...
+
+    <!-- Required CSS for ui-slider -->
+    <link rel="stylesheet"
+          href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.4/themes/smoothness/jquery-ui.css">
+
     <!-- compiled CSS --><% styles.forEach( function ( file ) { %>
     <link rel="stylesheet" type="text/css" href="/static/<%= file %>" /><% }); %>
 
+    <!-- Required js for ui-slider
+         FIXME: Why adding these two libraries in build.config.js does not work?
```
dongbohu: Yes, they are for the ui-slider widget. I tried both minified and non-minified version in **build.config.js**, but the slider never showed up.

interface/src/index.html
dongbohu: Because **d3.network** in active development now, so I didn't include it in **build.config.js**. If I did, every time d3.network changes, I will have to copy it to **vendor/** directory.

interface/src/index.html
mhuyck: It is tricky to get jQuery and Angular to place nicely together. The creator of this slider and other widgets out there do it, but I feel it should be avoided, if possible. Are there any sliders you could use that don't require jQuery?

mhuyck: play nicely, I mean. Autocorrect error.

interface/src/index.html
mhuyck: Okay with me, but we will need to fix that before deployment. Can you add a FIXME note, so we don't forget?

Another option is to include a step in the build script that pulls and incorporates the latest version automatically. That is essentially what happens in the vendor section, but those others are more formally published with version info, etc.

interface/src/index.html
dongbohu: I found a few slider libraries, and this one is in one of the coponents in **angular-ui** collection:
https://github.com/angular-ui/ui-slider
I feel that this probably will work best with Angular.

dongbohu:
I fixed the **jquery** and **jquery-ui** library dependency issue. The key is these two static library files must be put at the beginning of **js** array in **build.config.js**. I don't know exactly why.

interface/build.config.js
```diff
@@ -64,6 +64,12 @@ module.exports = {
    */
   vendor_files: {
     js: [
+      // For some unknown reason(s), jquery and jquery-ui library files MUST be
+      // put at the beginning of this array; Otherwise the ui-slider wouldn't
+      // show up on the web UI, and we got the following error in the console:
```
rzelayafavila: Hmmm - is this because they are also loaded onto the webpage in this order and they are dependencies for ui-slider?

interface/src/app/app.spec.js
```diff
@@ -1,17 +1,17 @@
...
-  describe( 'isCurrentUrl', function() {
+describe('AppCtrl', function() {
+  describe('isCurrentUrl', function() {
     var AppCtrl, $location, $scope;
 
-    beforeEach( module( 'adage' ) );
+    beforeEach(module('adage'));
 
-    beforeEach( inject( function( $controller, _$location_, $rootScope ) {
+    beforeEach(inject(function($controller, _$location_, $rootScope) {
```
rzelayafavila: Interesting - I had never noticed that the _$location_ in the parameters of this function in this test file had those underscores. I wonder why that is special about the $location service that makes it require that?

interface/build.config.js
dongbohu: Yes, they are dependencies for ui-slider, so I put them right before ui-slider library, but that is not enough. I tried a few locations for these two libraries, and seems that they have to be before angular.js too.

interface/src/app/app.spec.js
dongbohu: I think **_$location_** is a mock of **$location** in AngularJS.

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,262 @@
...
+])
+
+.config(function config($stateProvider) {
+  $stateProvider.state('gene_network', {
+    url: '/gene_network/?genes&mlmodel',
+    views: {
+      main: {
+        templateUrl: 'gene/network/network.tpl.html',
+        controller: 'GeneNetworkCtrl',
+        controllerAs: 'ctrl'
```
rzelayafavila: I have been thinking about this 'controllerAs' syntax, and it is definitely great that you are using it here, since it is very much considered a best practice. I wonder if we should change the rest of our codebase at some point to use 'controllerAs' syntax instead of '$scope'?

interface/build.config.js
rzelayafavila: Ah gotcha.

interface/src/app/gene/network/network.js
dongbohu: I like it because the code is easier to follow and we don't need $scope as a dependency any more.  We probably should create an issue to change to this syntax in the long run.

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,262 @@
...
+
+.controller('GeneNetworkCtrl',
+   ['$stateParams', 'EdgeService', 'NodeService', '$log',
+    function GeneNetworkController($stateParams, EdgeService, NodeService,
+                                   $log) {
+      var rawMinWeight = -1.0;
+      var rawMaxWeight = 1.0;
+      var scaledMaxWeight = rawMaxWeight - rawMinWeight;
+      var self = this;
+      // The following properties of "self" will be available to HTMl.
```
rzelayafavila: Totally minor, nitpicky typo: is that last "l" in HTML supposed to be capitalized?

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,262 @@
...
+      self.minCorrelation = -1.0;
+      self.maxNodeNum = Number.MAX_SAFE_INTEGER;
+      self.statusMessage = 'Connecting to the server ...';
+      self.sliderOptions = {  // Slider configuration.
+        range: 'max',
```
rzelayafavila: What does range 'max' do?

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,262 @@
...
+        stop: function(event, ui) {
```
rzelayafavila: What is this 'stop' function doing?

interface/src/app/gene/network/network.js
dongbohu: Thanks. I will correct it.

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,262 @@
...
+      }
+
+      var urlGenes = $stateParams.genes.split(',');
+      var setEdgeColor = d3.scale.linear()
+          .domain([0, scaledMaxWeight / 2.0, scaledMaxWeight])
+          .range(['green', 'orange', 'red']);
+
+      var genes = [];
+      var edges = [];
+      var network = d3.network()  // Initialize the network.
```
rzelayafavila: So I guess all of this code is using d3 and not vega at all for the visualization right? Also I guess we don't need to include d3 as one of the dependencies for this module at the top of this file since it is not an angular dependency?

interface/src/app/gene/network/network.js
dongbohu: This **range** option in **ui-slider** controls which section is referred to by **.ui-slider-range** in network.less. It is subtle to explain but I will try. As you can see, **network.less* includes these lines:
```
 .ui-slider-range {
    background: DarkGray;
  }
```
When **range** is 'max', the color from the current slider to the ending point will have a background of DarkGray. On the other hand, if **range is 'min', then the color from the starting point to the current slider position will be DarkGray. See the page on my local web server:
http://165.123.67.202/#/gene_network/?genes=6546
and move the slider back and forth, see whether my explanation makes sense.

interface/src/app/gene/network/network.js
dongbohu: It specifies a callback function that will be called when a user stops moving the slider.

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,262 @@
...
+        stop: function(event, ui) {
+          network.filter(self.minCorrelation - rawMinWeight,
+                         self.maxNodeNum).draw();
+        }
+      };
+
+      // Do nothing if no genes are specified in URL.
+      if (!$stateParams.genes || !$stateParams.genes.split(',').length) {
+        self.statusMessage = 'No genes are specified.';
+        return;
```
dongbohu: As far as I understand, controller in Angular is like a callback function that monitors user's action on the web UI. In this controller, the **return** statement will end this function call, and the web UI will show statusMessage.

interface/src/app/gene/network/network.js
rzelayafavila: Ah I see, thanks!

interface/src/app/gene/network/network.js
rzelayafavila: Ah, I see

interface/src/app/gene/network/network.tpl.html
```diff
@@ -0,0 +1,15 @@
+<h1>ADAGE: Gene-Gene Network</h1>
+
+<div ng-hide="ctrl.statusMessage">
```
rzelayafavila: So if statusMessage is not undefined or not false then this will show? What if the statusMessage is a blank string?

rzelayafavila:
Ok, I have one more question and comment, but other than that it looks pretty good to me @dongbohu!

interface/src/app/gene/network/network.tpl.html
dongbohu: If **statusMessage** is a blank string, this **div** will show up. As you may have noticed, when the gene netowork data has been retrieved from the backend successfully,  **statusMessage** will be set to a blank string.

dongbohu: As for the question on **ng-if**, it would be handy if Angular had both ""ng-if** and **ng-else**, but the latter is not supported, and the gene network should be shown only when **statusMessage** is a blank string. (Of course I can use two **ng-if**s to achieve the same effect.)

dongbohu: I put more thoughts on this **ng-if** and **ng-hide** question and added a comment to Issue #25 (which I closed yesterday). @rzelayafavila, please let me know whether it makes sense to you.

rzelayafavila: Ok, yep, these make sense. Thanks!

interface/src/index.html
mhuyck: I started trying to write out the reasons we are trying to avoid a jQuery dependency, but this article does an excellent job of explaining my thoughts on trying to combine jQuery with AngularJS code:  http://tech.zumba.com/2014/08/02/angularjs-forget-jquery/

I would like for us to think of jQuery as a "legacy" dependency. The details are complex because jQuery is indeed an excellent tool and even Angular embeds a stripped-down version they call jQuery-lite: https://docs.angularjs.org/misc/faq#does-angular-use-the-jquery-library- . At a fundamental level, however, what we are trying to do by using AngularJS in our system architecture is to get away from this "old" way of doing things.

Some people (including the developer of `ui-slider`, who lists "AngularUI team" in the LICENSE file) continue to use jQuery as part of their Angular projects. That is fine for them, but not for us. I should also point out that the README for `ui-slider` says "This is WIP so expect breaking changes", which is another good reason to avoid that component entirely apart from its jQuery dependency. The developer is practically advertising that it's not ready for production.

It is possible to implement a slider without requiring jQuery, and there are several examples I found via Google. The one that looked most full featured and actively supported was http://angular-slider.github.io/angularjs-slider/ . Source code here: https://github.com/angular-slider/angularjs-slider .

Please have a look at that and see if it would meet our needs for the slider here.

interface/build.config.js
mhuyck: Angular is compatible with jQuery. Part of it is implemented with a stripped-down jQuery they call jQuery-lite, and, in fact, Angular itself will automatically switch to using the jQuery you supply instead of jQuery-lite. I do not count this as a feature.

The possibility of having Angular itself change behavior with a change in jQuery is unsettling. I suspect this "adaptation" to the supplied jQuery is responsible for the sensitivity to JavaScript module loading order. Angular and jQuery will confuse each other unless they are loaded together exactly the right way.

interface/src/index.html
dongbohu: Okay, thanks @mhuyck. Let me try the new slider library.

interface/src/app/app.spec.js
```diff
@@ -1,17 +1,17 @@
...
       $location = _$location_;
       $scope = $rootScope.$new();
-      AppCtrl = $controller( 'AppCtrl', { $location: $location, $scope: $scope });
+      AppCtrl = $controller('AppCtrl', {$location: $location, $scope: $scope});
     }));
 
-    it( 'should pass a dummy test', inject( function() {
-      expect( AppCtrl ).toBeTruthy();
+    it('should pass a dummy test', inject(function() {
+      expect(AppCtrl).toBeTruthy();
```
mhuyck: I wish there were a BeautifyJS feature that could write our tests for us!

interface/src/index.html
mhuyck: Thank you! I hate asking you to change code that is working, but I really do feel that it would become a problem for maintenance down the road.

One other note about jQuery and Angular that I can't find now but saw in my reading is that I think you only get to include it once, so if one of your modules requires version A.8 of jQuery and another requires version A.7 but isn't yet compatible with A.8 then you are really stuck.

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,272 @@
...
+angular.module('adage.gene.network', [
+  'ui.router',
+  'placeholders',
+  'ui.bootstrap',
+  'rzModule'
+])
+
+.config(function config($stateProvider) {
+  $stateProvider.state('gene_network', {
+    url: '/gene_network/?genes&mlmodel',
```
mhuyck: I didn't know that `$stateProvider` handled parameters this way. Cool!

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,272 @@
...
+          .edgeColor(setEdgeColor)
+          .geneText(function(d) {
+            return d.label;
+          })
+          .legendStart(rawMinWeight)
+          .legendEnd(rawMaxWeight)
+          .legendText('Correlation');
+
+      var self = this;
+      // The following properties of "self" will be available to HTML.
```
mhuyck: Interesting. So does `this` give you the DOM element to which the controller is bound? Not sure I see how this works. We usually use an explicit reference to `$scope` in this type of situation, but maybe this works just as well.

interface/src/app/gene/network/network.js
mhuyck: My understanding is that it acts like an initializer for the control, so the code below will only execute if the control is somehow re-initialized with something in `$stateParams.genes`. Is that consistent with what you're seeing, @dongbohu?

interface/src/app/gene/network/network.js
dongbohu: This is a new Angular feature called `controllerAs` syntax. (I think it was introduced since angular 1.2). As far as I understand, it has two advantages compared to $scope:
(1) It removes $scope dependency.
(2) The HTML code will use **controllerTag.variableName** (**controllerTag** is defined on line 18) to access the properties of **self**, it shows explicitly that **variableName** is from the controller, thus the code is easier to understand.

interface/src/app/gene/network/network.js
dongbohu: @mhuyck: Yes.

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,272 @@
...
+      /**
+       * Draw gene-gene network.
+       * @param {void} null;
+       * @return {void}.
+       */
+      function drawNetwork() {
+        // Calculate svg size.
+        var minSvgSize = 600;  // Minimum size of svg.
+        var maxSvgSize = 1280; // Maximum size of svg.
+        var svgSize = genes.length * 10;
```
mhuyck: Nice. Is this how you auto-scale the view now?

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,272 @@
...
+        if (svgSize < minSvgSize) {
+          svgSize = minSvgSize;
+        } else if (svgSize > maxSvgSize) {
+          svgSize = maxSvgSize;
+        }
```
mhuyck: Have you tested how many genes you can handle before the `maxSvgSize` is too small to navigate the network? It'll be good to keep these limits in mind as we work through real-world example data.

interface/src/app/gene/network/network.js
dongbohu: Yes. This might be an overly simple way to auto-scale the size of network svg. (Ideally I would like to have two more slider bars or edit boxes, one for the width and the other for height, and allow users to set the width and height to whatever size they like. Maybe we can add that kind of feature later.)

interface/src/app/gene/network/network.js
```diff
@@ -0,0 +1,272 @@
...
+            }
+          );
+        }
+
+        /**
+         * Callback function to hide edge tips.
+         * @param {edge_data} data;
+         * @return {void}.
+         * ---------------------------------------------------------------------
+         * Since the edge tip is shown asynchronously, when edgeTip.hide() is
```
mhuyck: Can you add a FIXME label to this? I'm not sure off the top of my head how to fix this either, but we can locate this note more easily with that label.

mhuyck: For what it's worth: I needed a similar sort of workaround so the transitions would work right in the sample search results.

interface/src/app/gene/network/network.js
dongbohu: The key is actually not the number of genes in the network but the number of edges. When the total number of edges is less than 300, the pre-defined maximum size seems okay. On the other hand, when a network includes so many edges, even if the size is larger, it is not easy for a user to see details. So he/she will probably have to use the minimum correlation slider bar to filter out some edges.

interface/src/app/gene/network/network.js
dongbohu: Sure. Good idea!

interface/src/app/gene/network/network.tpl.html
mhuyck: Nice debugging on issue #25! Messing around with the DOM gets extra tricky when you're integrating Angular with external libraries such as `d3` and `vega`.

interface/src/app/gene/network/network.js
mhuyck: I like it. The `$scope` syntax can get very confusing. Might have to start using this approach.

dongbohu: I like it very much too. Many angular tutorials recommend using `controllerAs` to replace `$scope`, if possible.

mhuyck:
Looks awesome. Thank you for clarifying those points @dongbohu!

PR #27, opened 2016-11-01
Improvements on gene search module

rzelayafavila:
Mostly moving all files to a gene/search folder, writing a function to get either the gene standard_name or systematic_name in lieu of it, and sending the selected genes to the gene Network state as a string instead of an array. Not too heavy of a pull request.


interface/src/app/gene/search/search.js
```diff
@@ -264,18 +264,21 @@ angular.module('adage.gene.search', [
           for (var gene in $scope.selectedGenes) {
```
cgreene: This looks like a for loop over elements of an array. This seems like the case where we want to use the traditional loop over elements to avoid snagging a property.
http://stackoverflow.com/questions/1963102/what-does-the-jslint-error-body-of-a-for-in-should-be-wrapped-in-an-if-statemen

interface/src/app/gene/utils.js
```diff
@@ -14,6 +14,17 @@ angular.module('adage.gene.utils', [
...
       scope.updatePage(scope.page);
+    },
+
+    getGeneLabel: function(gene) {
+      // Returns standard_name if it exists and systematic_name otherwise.
+      var geneLabel = gene.systematic_name;
+
+      if (gene.standard_name) {
+        geneLabel = gene.standard_name;
+      } else {}
```
cgreene: Is this empty else required by our style?

cgreene:
I had a few quick comments.

interface/src/app/gene/utils.js
rzelayafavila: Ah good catch, I guess not. For some reason I thought it was required but I see that it isn't, and this piece of code is probably simple enough that adding the empty ``else {}`` does not add any clarity. I will take it out.

interface/src/app/gene/search/search.js
rzelayafavila: Mmmm, totally - I agree. In fact, there are a couple other places in the code where I could either use this other approach, or wrap statements in the for loop inside an ``if`` statement with a ``hasOwnProperty()`` check, so I will go ahead and do this.

cgreene: I think @mhuyck was concerned about the performance of hasOwnProperty, so I didn't comment on the other spots. I only commented on the one that appeared to be an array - that should not cause a performance issue and avoids the danger of iterating over non-numeric properties.

dongbohu: @rzelayafavila: This **for-in without filter** issue is actually caught by eslint, because our current eslint rules do not allow it. Here is what I get when running eslint on this file:
```
  202:13  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  216:13  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  264:11  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in

✖ 3 problems (3 errors, 0 warnings)
```

rzelayafavila: Aha, I see - so given what @cgreene and @mhuyck were saying about the the performance of ``hasOwnProperty``, what should we do about these other cases? Should we wrap in an ``if`` statement with ``hasOwnProperty``? Should we try to loop some other way?

rzelayafavila: I mean for the cases in which we are not looping over an array. I know that for an array we should use the good old ``for (var i = 0; i < this.length; i++) {``.

dongbohu: I can think of two solutions:

(1) Use **hasOwnProperty()** method to filter out the inherited enumerable properties. (This solution is recommended by Douglas Crockford, the author of **jslint** and the popular book "JavaScript: the good parts"). Besides the possible performance issue, another downside of this solution is that even **hasOwnProperty()** method may be overwritten by a user and thus not do what is expected.

(2) Another solution is to create a dummy object without any prototype:
```dummyObj = Object.create(null);```
then add user-defined properties to this dummy object. You will be sure that this dummy object does not inherit any properties from anywhere.
The downside of this solution is that in many cases, you may have to add new properties to an object that already exists, thus you don't have the luxury to work from your dummy object.

I prefer the first solution.

**Comment by cgreene** [comment]
I prefer the first solution. Are there documented reports of issues with
performance in the situations that we're likely to encounter?

**Comment by dongbohu** [comment]
I have never seen any formal documentation or benchmark about the performance hit of calling **hasOwnProperty()**, I think I saw it from some javascript blog that was mentioned briefly. Maybe @mhuyck has a better resource?

**Comment by rzelayafavila** [comment]
Aha! I think I have found a good solution. Calling Object.keys(), and then iterating over that as we would over any other array. Apparently it is more optimal than checking hasOwnProperty every time: https://leftshift.io/4-javascript-optimisations-you-should-know  @dongbohu @cgreene @mhuyck

**Comment by rzelayafavila** [comment]
I also did some googling and didn't find any benchmark about the performance hit, other than people saying that it would slow down performance a bit to run the check every time.

**Comment by rzelayafavila** [comment]
I just talked to @mhuyck over slack and I know he is about to post some thoughts on this, though.

**Comment by dongbohu** [comment]
Calling **Object.keys()** method seems to be a good idea. I forgot this method (which was added since ES5, I believe).

File: interface/src/app/gene/search/search.js (hunk 1 above)
Comment (mhuyck) [inline interface/src/app/gene/search/search.js]: I should clarify my concern about `hasOwnProperty()`. I do not think it has any inherent performance trouble, in and of itself. The issue I was trying to highlight is that it would add unnecessary steps in any loop that could have performance implications, especially if it is an unnecessary step.

JavaScript has been around a long time and it is now used far more extensively than originally planned. Many of the warnings about its use stem from things you need to be careful about when interacting with the DOM and external libraries, and I count this *for-in without filter* rule among them.

In this context, `$scope.selectedGenes` is created by our own code in `search.js` (line 105):

```
var selectedGenes = {};
```

In a case like this, there is nothing to be concerned about. The use of `{}` for creating an empty Object is the preferred way to do this (because it is more readable and compact than `var selectedGenes = new Object;`. While it is technically true that some library we are using could muck around with the prototype on `Object`, that practice is deeply frowned upon these days and if we find any library we use doing that we should look for an alternative.

With all that said, if the loop is well bounded and not performance-critical (i.e. you're not looping over 600 `Activity` records per sample as I am on the heatmap) then there is no performance concern adding the `hasOwnProperty` filter. It will burn a few extra cycles that do nothing, but for a loop with only a few hundred iterations nobody will notice. (Unless that loop is then put inside another loop for any reason.)

dongbohu: One last comment on **search.js**: Considering the number of lines in this file, it may be worth the time to divide it into a few separate files (such as *controller.js*, *services.js*, etc). It is totally up to you. If you want, you can do it in another PR. 

Believe it or not, eslint even allows us to set an upper limit on how many lines a certain *.js file should have. It is disabled in google's eslint rules right now. The default upper limit is 300:
http://eslint.org/docs/rules/max-lines

Everything else in this PR seems good to me. Thanks @rzelayafavila.

cgreene:
@rzelayafavila : actual performance tests indicate no difference in the two syntax versions of for:
http://jsfiddle.net/kNLda/27/

rzelayafavila:
@dongbohu: In terms of separating search.js into smaller files, I was weighing the options and almost did that for this PR, but I think I might do that in a new pull request (pretty lightweight one) once this gets merged. That way it will be easy in the new PR to see that it is just about separating the code. Do you think that makes sense?

dongbohu:
Sure. Thanks.

rzelayafavila:
Ok, I am about to merge this PR. @dongbohu, @mhuyck, @cgreene: Everything LGTY?

dongbohu:
Yes, looks good to me.

rzelayafavila:
Ok, thanks everyone!

PR #28, opened 2016-11-02
Separating gene/search/search.js into different files.

rzelayafavila:
Moving all directives in interface/src/app/gene/search/search.js to a new file called interface/src/app/gene/search/directives.js and all factories to a new file called interface/src/app/gene/search/services.js. This looks like a very large pull request, but it is basically just breaking up a large file, as per @dongbohu's recommendations on the last pull request.


rzelayafavila:
I want to make sure I am separating the code into modules and then importing it in the most Angular way.

cgreene:
Not an area that I can review. Removed myself for this reason.

rzelayafavila:
@cgreene : Ok, thanks.

interface/src/app/gene/search/directives.js
```diff
@@ -0,0 +1,321 @@
+angular.module('adage.gene.search.directives', [
```
dongbohu: If you replace lines 1-4 into:
```angular.module('adage.gene.search')```
I believe it will get the reference of the module **adage.gene.search** instead of creating a new one. Then you can add directorives or services to it.

Comment (dongbohu) [inline interface/src/app/gene/search/directives.js]: By the way, when all these files are in the same module, you can even divide **directives.js** further if you want.

Comment (rzelayafavila) [inline interface/src/app/gene/search/directives.js]: Ah yes, you're totally right, I had totally forgotten about that syntax. As we talked about in person this morning though, it looks like grunt-uglify is minifying and loading files in alphabetical order, so that can cause a problem if the file that defines the module is not the first alphabetically. I will look more into solutions to that this morning, and if I don't find anything, we can go with the differently named modules.

File: interface/src/app/gene/searchMany/search.js [hunk 2]
```diff
@@ -1,8 +1,108 @@
-angular.module('adage.gene.search.directives', [
-  'adage.gene.search.services',
-  'adage.gene.utils'
+angular.module('adage.gene.search', [
```
cgreene: The filename `search.js` and module name `search` for this seems a bit weird since, if I understand correctly, this is only part of the type of search that you plan to support. Not sure how file naming is usually done but maybe `search_many.js` or something might be more appropriate?

cgreene: @rzelayafavila : you changed this even before my comment came through. Wow!

rzelayafavila: Oh - just added new commits right after those where that is addressed.

rzelayafavila: Let me know if those address that issue

cgreene: Yep! :+1:

rzelayafavila: Oh ok, nice!

rzelayafavila:
Ok @dongbohu, @mhuyck, @cgreene : I moved a lot of code around so that the previous gene.search module was broken up into: gene.searchMany, gene.searchFew and gene.selected, which made more functional sense. Also, I noticed that I could get rid of the ``link`` function (and an annoying ``$apply`` in there) and moved that code into the controllers instead for a few directives, since that is what is considered a best practice. Let me know how it looks!

mhuyck:
This looks like exactly the sort of *functional* division of labor I was thinking about. I'm just giving a quick skim so I can stay focused on what I'm doing, but on the surface this looks like it does what I was trying to describe yesterday. Thank you!

I'm not sure what the exact distinction between `searchFew` and `searchMany` actually *is*, but if I would understand by looking at live pages then there is no need to say more. Breaking out `selected` also makes perfect sense to me.

dongbohu:
I just realized that it is possible to use `controllerAs` syntax in the definition of a directive too. Since most of your directives' template HTMLs are pretty short, it is up to you whether you want to use **controllerAs** or not.

dongbohu:
Everything else seems good to me.

rzelayafavila:
@dongbohu : Ok, I was thinking it over, and I think I might wait on that until a future PR (since I think it would be changing more than a few lines of code) to keep this one on the topic of separating the gene search module.

dongbohu:
Sure. It is up to you.

PR #30, opened 2016-11-04
Add .eslintignore to ignore some js files

dongbohu:
I created a **.eslintignore** file, which is used to ignore some JavaScript files when you run `eslint` from command line or when your IDE runs it on the background. The ignored files either belong to other packages (such as ngbp) or are generated by `grunt` compilation.

### DIFF
--- a/interface/.eslintignore
+++ b/interface/.eslintignore
@@ -1,0 +1,7 @@
+build.config.js
+eslint.google.js
+Gruntfile.js
+build/*
+karma/*
+node_modules/*
+vendor/*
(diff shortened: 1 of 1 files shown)


interface/.eslintignore
```diff
@@ -0,0 +1,7 @@
+build.config.js
+eslint.google.js
+Gruntfile.js
+build/*
```
rzelayafavila: Nice! Should bin/* also be here since build/* is in here?

dongbohu: Good catch. I forgot that **bin** also includes js files.

rzelayafavila:
Ok, LGTM! :+1:

PR #32, opened 2016-11-07
Cache activity data

mhuyck:
Initial caching implementation for activity data (see `sampleBin.rebuildHeatmapActivity()`), camelCase and other ESLint compliance.

We need a similar caching ability for Sample metadata, but that will be handled in a separate pull request once this request is reviewed and any feedback incorporated.

### DIFF
--- a/interface/src/app/analyze/analysis/sampleBin.js
+++ b/interface/src/app/analyze/analysis/sampleBin.js
@@ -27,13 +27,21 @@ function($log, Sample, Activity) {
             ' already in the sample list; ignoring.');
       } else {
         SampleBin.samples.push(+id);
+        // TODO need to report query progress and errors somehow
+        // TODO this should use caching and promises to fetch sample data
+        // 1. check cache for existing Sample data. Hit = return
+        // 2. if cache miss: 
+        // 2a. getSampleDetails & append a promise to SampleBin.promises
+        // 2b. getActivityForSample & append a promise to SampleBin.promises
+        // ... now can use Promise.all(SampleBin.promises).then() to trigger
+        // a redraw of the heatmap when ready
       }
     },
 
     remove_sample: function(id) {
       var pos = SampleBin.samples.indexOf(+id);
       SampleBin.samples.splice(pos, 1);
-      // TODO need to report query progress and errors somehow
+      // TODO now need to mutate SampleBin.heatmapData.activity rather than re-retrieve data
       Activity.get({sample__in: SampleBin.samples.join()},
         // success callback
         function(responseObject, responseHeaders) {
--- a/interface/src/app/analyze/analysis/sampleBin.js
+++ b/interface/src/app/analyze/analysis/sampleBin.js
@@ -13,12 +13,14 @@ angular.module('adage.analyze.sampleBin', [
   return $resource('/api/v0/activity/');
 }])
 
-.factory('SampleBin', ['$log', 'Sample', 'Activity',
-function($log, Sample, Activity) {
+.factory('SampleBin', ['$log', '$cacheFactory', 'Sample', 'Activity',
+function($log, $cacheFactory, Sample, Activity) {
   var SampleBin = {
     samples: [],
     heatmapData: {},
     sampleData: {},
+    sampleCache: $cacheFactory('sample'),
+    activityCache: $cacheFactory('activity'),
 
     add_sample: function(id) {
       if (SampleBin.samples.indexOf(+id) !== -1) {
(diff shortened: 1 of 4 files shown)


dongbohu:
My second thought: It might be a good idea to use **SampleBin** directly in the method definitions, since sometimes **this** is ambiguous.

mhuyck:
Thank you for bringing `this` up, @dongbohu. I have a great amount of confusion about when it's safe to use `this` and when it isn't, in JavaScript. I'm reading up on the situation now.

interface/src/app/analyze/analyze.tpl.html
```diff
@@ -23,8 +23,8 @@ <h1 class="page-header">
             </div>
             <div class="media-body">
               <button type="button" class="btn"
-              ng-class="sb.has_item(search_item)?'disabled':''"
-              ng-click="sb.add_item(search_item)">
+              ng-class="sb.hasItem(search_item)?'disabled':''"
```
rzelayafavila: :+1: on the change to camelCase. I was wondering if the line with ``sb.hasItem(search_item)?'disabled':''`` might be to much logic for the template and might be better to move into the controller though?

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -13,68 +13,58 @@ angular.module('adage.analyze.sampleBin', [
...
+        SampleBin.addExperiment(searchItem.related_items);
       }
     },
 
-    has_item: function(search_item) {
-      if (search_item.item_type === 'sample') {
-        if (SampleBin.samples.indexOf(+search_item.pk) !== -1) {
+    hasItem: function(searchItem) {
+      if (searchItem.item_type === 'sample') {
+        if (SampleBin.samples.indexOf(+searchItem.pk) !== -1) {
```
rzelayafavila: Is the ``+`` operator here turning  searchItem.pk into an int? Isn't it already supposed to be an int though?

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -133,7 +168,7 @@ function SampleBinCtrl($scope, $log, $uibModal, Sample, SampleBin) {
   $scope.sb = SampleBin;
 
   $scope.show = function() {
-    var modalInstance = $uibModal.open({
```
mhuyck: Right. ESLint picked up on the fact that I was declaring the `var` and never using it. True to my philosophy of avoiding modals, this one will go away soon as well. Just trying to keep the pull requests focused by doing one thing at a time.

interface/src/app/analyze/analyze.tpl.html
mhuyck: In this case, I'd say all of the logic is in the `hasItem()` call and I'm just making use of it to specify the class. I like keeping the class name here in the HTML, if possible, rather than putting that into the controller logic. Does that separation make sense? Not sure I'm perfectly consistent on that, but this is what I eventually landed on as I worked through this code the first time.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: Right. It's here as a succinct way to make sure it's an `int`. There is a possibility that `searchItem.pk` winds up a `str` -- I think I saw that happen in debugging, actually.

rzelayafavila: Oh, I just realized - is it because the REST API returns all of these values as strings, and we then have to convert them to ints manually?

rzelayafavila: Ah ok, gotcha

mhuyck:
The following two resources have been most informative on the use of `this` in JavaScript:

I started here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this , but it mostly made no sense. I read a few other things, then came to StackOverflow here: http://stackoverflow.com/questions/3127429/how-does-the-this-keyword-work , which seems to be the most informative explanation I've seen to date.

After lunch, I'll turn back to the code and see if I can make sense of the right way to use `this` in the context of the `sampleBin` code. I suspect I ended up doing things the way I did because of problems with callbacks, where I was getting the `this` reference messed up.

mhuyck:
Thank you for the suggestion to use `this`, Dongbo. It's a little tricky knowing when it's safe to use in JavaScript, but very helpful to have dug into it so I now understand!

dongbohu:
@mhuyck: I spent some time reading the two resources that you listed (both of them are so long that I didn't read them word by word) and other books on the usage of **this**. It seems to be an open question whether `this` should be used in your current scenario. I can definitely see the advantage of using the object name directly in the methods now, because you don't have to worry about any callbacks in the future that may call these methods in whatever context. 

Many other people recommend reassignment of `this` to another variable such as `self` (or `that` or `me`), then use this new variable name in callbacks. This pattern (arguably) makes the code easier to follow.

At the end of my reading I feel that this is more like a preference instead of strict rule. So whatever way you prefer is okay with me. I should thank you too for giving me a better understanding of **this**.

interface/src/app/analyze/analyze.tpl.html
rzelayafavila: Ah I see. Ok, I think that makes sense!

PR #33, opened 2016-11-08
Changes on the backend for the new Node page

dongbohu:
This PR includes the following changes:

1. api.py: Set `full` flag to true on `gene` attribute of **ParticipationResource** class; Add a new **MLModelResource** class (based on `MLModel` in **model.py**). This change will allow the Node page to access the corresponding machine learning model's name, which will be shown on frontend.
 
2. url.py: Enable **MLModelResource**.

### DIFF
--- a/adage/adage/urls.py
+++ b/adage/adage/urls.py
@@ -19,9 +19,10 @@ from django.contrib import admin
 from tastypie.api import Api
 from organisms.api import OrganismResource
 from genes.api import GeneResource
-from analyze.api import SearchResource, ExperimentResource,\
-    AnnotationTypeResource, SampleResource, NodeResource, ActivityResource,\
-    EdgeResource, ParticipationResource
+from analyze.api import (SearchResource, ExperimentResource,
+                         AnnotationTypeResource, SampleResource,
+                         MLModelResource, NodeResource, ActivityResource,
+                         EdgeResource, ParticipationResource)
 
 v0_api = Api(api_name='v0')
 v0_api.register(SearchResource())
@@ -30,6 +31,7 @@ v0_api.register(AnnotationTypeResource())
 v0_api.register(SampleResource())
 v0_api.register(OrganismResource())
 v0_api.register(GeneResource())
+v0_api.register(MLModelResource())
 v0_api.register(NodeResource())
 v0_api.register(ActivityResource())
 v0_api.register(EdgeResource())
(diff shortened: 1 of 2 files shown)


adage/analyze/api.py
```diff
@@ -342,7 +350,7 @@ def apply_filters(self, request, applicable_filters):
 
 class ParticipationResource(ModelResource):
     node = fields.IntegerField(attribute='node_id', null=False)
```
cgreene: Should node also be a foreign key? Not sure what the implications would be but the parallel construction might be nice.

rzelayafavila:
Also, along the same lines of @cgreene's comment about the Foreign Key: Should the mlmodel in EdgeResource, line 299, also be a ForeignKey now that we have a MLModelResource? (For some reason, GitHub didn't let me post this comment right at line 299 - I guess because it is too far away from any of the places where code changed).

adage/analyze/api.py
mhuyck: Good point. The key relationships are captured and enforced at the level of `models.py` but not here in `api.py`. If the attributes are explicitly declared to TastyPie as `ForeignKey`s, we get the ability to access those related objects "for free."

The only reason I can see not to do this is if we don't want to expose those models directly via an API, but I think there is no harm in doing that. Shouldn't we make all of our relationships explicit via TastyPie?

adage/adage/urls.py
```diff
@@ -19,9 +19,10 @@
 from tastypie.api import Api
 from organisms.api import OrganismResource
 from genes.api import GeneResource
-from analyze.api import SearchResource, ExperimentResource,\
-    AnnotationTypeResource, SampleResource, NodeResource, ActivityResource,\
-    EdgeResource, ParticipationResource
```
mhuyck: Thank you for ridding us of the backslash curse!

adage/analyze/api.py
cgreene: @dongbohu : this rationale makes total sense to me. Can you put in a quick comment noting why there is a difference? Basically mentioning the performance and that we don't, at this time, intend to do a query that returns all nodes given an input gene.

dongbohu: @cgreene: Sure.

mhuyck: Performance is an excellent reason. I also agree wholeheartedly with the rationale and the suggestion to comment about deliberately not using `ForeignKey`.

PR #36, opened 2016-11-15
Implementation of "heavy genes" section on Node page

dongbohu:
This PR includes all frontend changes for the implementation of **heavy genes** section on **Node** page.  I also created a few blank "upcoming" sections on **node.tpl.html**. They are temporary placeholders that will be eventually replaced by the implementation code.

### DIFF
--- a/interface/src/app/node/heavy_genes.tpl.html
+++ b/interface/src/app/node/heavy_genes.tpl.html
@@ -1,11 +1,9 @@
 <div ng-if="!queryStatus">
   <h4>
-    <div style="float: left">
-      <b>High-weight Genes: </b>{{genes.length}}
-    </div>
-    <div ng-if="genes.length > 0">
-      &nbsp({{hypoPercentage}}% Hypothetical)
-    </div>
+    <b>High-weight Genes: </b>{{genes.length}}
+    <span ng-if="genes.length > 0">
+      &nbsp;({{hypoPercentage}}% Hypothetical)
+    </span>
   </h4>
 
   <table class="table table-striped">
--- a/interface/src/app/node/heavy_genes.tpl.html
+++ b/interface/src/app/node/heavy_genes.tpl.html
@@ -1,0 +1,30 @@
+<div ng-if="!queryStatus">
+  <h4>
+    <div style="float: left">
+      <b>High-weight Genes: </b>{{genes.length}}
+    </div>
+    <div ng-if="genes.length > 0">
+      &nbsp({{hypoPercentage}}% Hypothetical)
+    </div>
+  </h4>
+
+  <table class="table table-striped">
+    <thead>
+      <tr>
+        <th>Systematic Name</th>
+        <th>Standard Name</th>
+        <th>Description</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr ng-repeat="gene in genes | orderBy: 'sysName'">
+        <td><b>{{gene.sysName}}</b></td>
+        <td>{{gene.stdName}}</td>
+        <td>{{gene.desc}}</td>
+      </tr>
+    </tbody>
+  </table>
+
+</div>
+
+<div ng-bind="queryStatus"></div>
--- a/interface/src/app/node/node.js
+++ b/interface/src/app/node/node.js
@@ -1,0 +1,98 @@
+/**
+ * "adage.gene.network" module.
+ */
+
+angular.module('adage.node', [
+  'ui.router',
+  'placeholders',
+  'ui.bootstrap'
+])
+
+.config(function config($stateProvider) {
+  $stateProvider.state('node', {
+    url: '/node/{id:int}',
+    views: {
+      main: {
+        templateUrl: 'node/node.tpl.html',
+        controller: 'NodeCtrl as ctrl'
+      }
+    },
+    data: {pageTitle: 'Node Information'}
+  });
+})
+
+.factory('NodeInfo', ['$resource', '$stateParams',
+  function($resource) {
+    return $resource('/api/v0/node/:id');
+  }
+])
+
+.factory('HeavyGenes', ['$resource', function($resource) {
+  return $resource('/api/v0/participation/');
+}])
+
+.controller('NodeCtrl', ['NodeInfo', '$stateParams', '$log',
+  function NodeController(NodeInfo, $stateParams, $log) {
+    var self = this;
+    if (!$stateParams.id) {
+      self.statusMessage = 'Please specify node ID in the URL.';
+      return;
+    }
+    self.id = $stateParams.id;
+    self.statusMessage = 'Connecting to the server ...';
+    NodeInfo.get(
+      {id: self.id},
+      function success(response) {
+        self.name = response.name;
+        self.mlmodel = response.mlmodel.title;
+        self.statusMessage = '';
+      },
+      function error(err) {
+        $log.error('Failed to get node information: ');
+        self.statusMessage = 'Failed to get node information from server';
+      }
+    );
+  }
+])
+
+.directive('highWeightGenes', ['HeavyGenes', '$log',
+  function(HeavyGenes, $log) {
+    return {
+      templateUrl: 'node/heavy_genes.tpl.html',
+      scope: {
+        nodeId: '='
+      },
+      link: function($scope) {
+        $scope.queryStatus = 'Connecting to the server ...';
+        HeavyGenes.get(
+          {node: $scope.nodeId, limit: 0},
+          function success(response) {
+            $scope.genes = [];
+            var i = 0, n = response.objects.length;
+            var sysName, stdName, desc;
+            var numHypo = 0;
+            for (; i < n; ++i) {
+              sysName = response.objects[i].gene.systematic_name;
+              stdName= response.objects[i].gene.standard_name;
+              desc = response.objects[i].gene.description;
+              if (desc.toLowerCase() === 'hypothetical protein') {
+                ++numHypo;
+              }
+              $scope.genes.push(
+                {sysName: sysName, stdName: stdName, desc: desc});
+            }
+            $scope.hypoPercentage = Math.round(numHypo / n * 100);
+            $scope.queryStatus = '';
+          },
+          function error(err) {
+            $log.error('Failed to get high weight genes: ' + err);
+            $scope.queryStatus = 'Failed to get high weight genes from server';
+          }
+        );
+      }
+    };
+  }]
+)
+
+// TODO: Directives for the other setions
+;
--- a/interface/src/app/node/node.js
+++ b/interface/src/app/node/node.js
@@ -1,5 +1,5 @@
 /**
- * "adage.gene.node" module.
+ * "adage.node" module.
  */
 
 angular.module('adage.node', [
--- a/interface/src/app/node/node.js
+++ b/interface/src/app/node/node.js
@@ -1,14 +1,13 @@
 /**
- * "adage.gene.network" module.
+ * "adage.gene.node" module.
  */
 
 angular.module('adage.node', [
   'ui.router',
-  'placeholders',
   'ui.bootstrap'
 ])
 
-.config(function config($stateProvider) {
+.config(['$stateProvider', function($stateProvider) {
   $stateProvider.state('node', {
     url: '/node/{id:int}',
     views: {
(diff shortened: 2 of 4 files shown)


dongbohu:
The two new issues reported by codeclimate are trivial. I think we can ignore them now.

dongbohu:
This is an example of what the implemented web UI looks like:
http://165.123.67.202/#/node/10999

interface/src/app/node/heavy_genes.tpl.html
```diff
@@ -0,0 +1,30 @@
+<div ng-if="!queryStatus">
+  <h4>
+    <div style="float: left">
+      <b>High-weight Genes: </b>{{genes.length}}
+    </div>
+    <div ng-if="genes.length > 0">
+      &nbsp({{hypoPercentage}}% Hypothetical)
```
mhuyck: Missing semicolon in nbsp HTML entity. Should be: `&nbsp;`

dongbohu: Thanks. I am surprised that the web browser accepted it.

mhuyck: Also: since this is really a continuation of the `<h4>` above rather than a semantically different section, you could incorporate it directly using `<span>` tags, which are treated in-line rather than `<div>` which, by default is handled as a block element.

I would simplify lines 2-9 as follows:
```
<h4>
  <b>High-weight Genes</b>: {{genes.length}}
  <span ng-if="genes.length > 0">&nbsp;({{hypoPercentage}}% Hypothetical)</span>
</h4>
```

This has the added benefit of eliminating the need for `style="float: left"` on line 3.

dongbohu: Good point. I forgot that I could use **span** tag.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * "adage.gene.network" module.
```
mhuyck: Copy & paste error? I think this comment should say `adage.node`, right?

dongbohu: Yes, that is a dumb typo.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+angular.module('adage.node', [
+  'ui.router',
+  'placeholders',
```
mhuyck: The `angular-placeholders` module is handy when mocking up pages (http://joshdmiller.github.io/angular-placeholders/ ) but I don't see any place you're using them here so this dependency can be removed.

dongbohu: Good to know. Thanks. (I had to confess that I never really understood what **angular-placeholders** module  exactly does.)

interface/src/app/node/heavy_genes.tpl.html
mhuyck: Browsers tend to be forgiving of typos. Most of the web would not work if they weren't! It is possible to run an HTML validator, too, which would point out such errors. GitHub highlighted the `&` in red because it was an error, in fact.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+  'ui.bootstrap'
+])
+
+.config(function config($stateProvider) {
```
mhuyck: This should have dependency injection for `$stateProvider`.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+      main: {
+        templateUrl: 'node/node.tpl.html',
+        controller: 'NodeCtrl as ctrl'
+      }
+    },
+    data: {pageTitle: 'Node Information'}
+  });
+})
+
+.factory('NodeInfo', ['$resource', '$stateParams',
```
mhuyck: It appears `$stateParams` is not used, so it should be removed.

dongbohu: Good catch. My original version used this parameter but I removed it later.

interface/src/app/node/node.js
dongbohu: Sure. Thanks.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+  function($resource) {
+    return $resource('/api/v0/node/:id');
+  }
+])
+
+.factory('HeavyGenes', ['$resource', function($resource) {
+  return $resource('/api/v0/participation/');
```
mhuyck: Maybe a dumb question, but I want to be sure I fully understand...

Participation tells us what genes participate in what nodes, right? Here you are defining an Angular function to query the `participation` endpoint, but calling it `HeavyGenes`. Does the API only provide access to high-weight, or heavy genes (via a filter?), or could I also get low-weight genes via this resource?

I am strongly in favor of having the naming precisely match what the feature does. If the `participation` API is filtered to only high-weight genes, then let's rename it. If the API is general, let's name the Angular service to match the back-end.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+}])
+
+.controller('NodeCtrl', ['NodeInfo', '$stateParams', '$log',
+  function NodeController(NodeInfo, $stateParams, $log) {
+    var self = this;
+    if (!$stateParams.id) {
+      self.statusMessage = 'Please specify node ID in the URL.';
```
mhuyck: This is nice error reporting. Thank you!

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+    self.id = $stateParams.id;
```
mhuyck: I am pondering the use of `self` vs. `this` here. I know you need a closure variable in the callbacks below, but wouldn't it be slightly more clear to use `this` in a spot like this line and a more descriptive closure variable that is clearly labeled for callbacks where needed?

This is mainly a question because I haven't formed a firm opinion yet.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+      },
+      function error(err) {
+        $log.error('Failed to get node information: ');
+        self.statusMessage = 'Failed to get node information from server';
+      }
+    );
+  }
+])
+
+.directive('highWeightGenes', ['HeavyGenes', '$log',
```
mhuyck: For consistency, we should probably inject `$scope` here as well.

interface/src/app/node/node.js
dongbohu: As you can see from the code, **HeavyGenes** service retrieves all records whose node ID is the given value from **Participation** table on backend. There is no any filter involved because **Paticipation** table only have two columns: node ID and gene ID. Are you suggesting that this service be named **ParticipationService** or something like that?

interface/src/app/node/node.js
dongbohu: **self** is used here to denote that all of its properties will be accessible from the HTML that is under control of this controller. This usage is more like a coding style that is bundled with **controllerAs** syntax. Does this make sense?

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+        HeavyGenes.get(
+          {node: $scope.nodeId, limit: 0},
+          function success(response) {
+            $scope.genes = [];
+            var i = 0, n = response.objects.length;
+            var sysName, stdName, desc;
+            var numHypo = 0;
+            for (; i < n; ++i) {
+              sysName = response.objects[i].gene.systematic_name;
+              stdName= response.objects[i].gene.standard_name;
```
dongbohu: My fault. Thanks. (I should blame eslint for not being able to catch it).

interface/src/app/node/node.js
mhuyck: Right, I didn't see any filter here, but I wasn't sure if the `participation` back-end did any filtering and didn't dig into the code for that -- hence the caveat that this may be a dumb question.

Yes, if the back-end also does no filtering, I'd suggest that this service use the same name as the API. If the back-end *does* do filtering, then I'd suggest that the API name be changed to reflect that.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+      scope: {
+        nodeId: '='
+      },
+      link: function($scope) {
+        $scope.queryStatus = 'Connecting to the server ...';
+        HeavyGenes.get(
+          {node: $scope.nodeId, limit: 0},
+          function success(response) {
+            $scope.genes = [];
+            var i = 0, n = response.objects.length;
```
dongbohu: This is actually my personal preference to fight with the infamous **variable hoisting** in JavaScript. I am sure you know that a line like this:
```
var foo = 123;
for (var i = 0; i < 10; ++i) { 
  // ... 
}
```
is equivalent to:
```
var foo = 123;
var i;
for (i = 0; i < 10; ++i) { 
  // ... 
}
```
So I put the variable declarations and definitions at the beginning to try to emphasize all declared variables in this function.

interface/src/app/node/node.js
mhuyck: I was surprised that ESLint didn't pick that up too!

interface/src/app/node/node.js
```diff
@@ -0,0 +1,98 @@
...
+            var sysName, stdName, desc;
+            var numHypo = 0;
+            for (; i < n; ++i) {
+              sysName = response.objects[i].gene.systematic_name;
+              stdName= response.objects[i].gene.standard_name;
+              desc = response.objects[i].gene.description;
+              if (desc.toLowerCase() === 'hypothetical protein') {
```
dongbohu: As far as I know, only a gene's description field can tell whether it is hypothetical or not. Deb, Georgia and Casey had some discussions on this topic in issue #11. I just used the most straightforward way to implement it.

mhuyck: Okay, that is what I was afraid of. So this is yet another example of Bioinformatics being messy.

@cgreene: it looks as if this is the only way to distinguish predicted genes from actual genes, but it strikes me that coding this into our UI ties us to a brittle encoding. Should we flag this as an issue to follow up sometime down the line? Maybe it's something we can fix with the data refinery?

cgreene: @mhuyck / @dongbohu I agree that this is the only way that we have available to us right now, and it is brittle and messy. A refinery approach would be great. I agree that going with this strategy for now and flagging it for future improvements makes sense.

interface/src/app/node/node.tpl.html
```diff
@@ -0,0 +1,35 @@
+<h1>ADAGE: Node Information</h1>
+
+<div ng-if="!ctrl.statusMessage">
```
dongbohu: Yes, if **statusMessage** is not blank, the whole page (including all sub-sections) will be hidden, because every sub-section depends on a valid node. When **statusMessage** is non-blank, the retrieval of the node's basic information (node name, machine learning model) must have failed, so it doesn't make sense to show any other information that is directly related to this node.

On the other hand, each section has (or will have) its own status. For example, in **Heavy Weight Gens** section, a **queryStatus** property is defined to indicate this section's query status. If it is not blank, only the corresponding section will be hidden, and it won't affect any other sections.

mhuyck: Okay, so my gut was wrong on this one... carry on!

interface/src/app/node/node.js
mhuyck: Is there somewhere to capture notes about the refinery? We could make a note about this and link to issue #11 as well.

cgreene: Not yet - probably best to capture it in a new issue in this repository. I'd do it outside of #11, since I expect that will close when we get a first draft of the page done.

interface/src/app/node/node.js
dongbohu: Yes, when AngularJS executes link function (behind the scene), it calls the function in the following fashion:
```
link($scope, DOM_element, attributes_of_DOM_element, current_controller, transclude)
```
That is why sometimes link is defined like this:
```
link: function($scope, $element, $attrs, ctrl, $transclude)
```
(But this kind of usage seems rare.)

interface/src/app/node/node.js
dongbohu: Sure. I will rename the service.

dongbohu:
@rzelayafavila and @mhuyck: It is possible to enforce a space character before and after infix operators (including **=**):
http://eslint.org/docs/rules/space-infix-ops
I have tested this option:
```
'space-infix-ops': 2
```
If you agree, I can add it into our own eslint rules.

interface/src/app/node/node.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * "adage.gene.node" module.
```
mhuyck: Shouldn't this match the name below (`adage.node`)?

dongbohu: Yes, this is yet another dumb error I made.

dongbohu:
I am going to file a separate PR that includes the new eslint rule and a few other trivial changes.

cgreene:
:+1:

mhuyck:
Looks great, @dongbohu. You got everything I found.

dongbohu:
@mhuyck: Thank you very much for pointing out the dumb errors. I will try my best to avoid them in future PRs.

mhuyck:
I wouldn't call those "dumb" errors. Part of the value of code review is fresh eyes to spot things we all tend to overlook as we get familiar with the code. Proof reading is important for good prose writing for exactly the same reasons!

PR #37, opened 2016-11-17
Heatmap sample sorting

mhuyck:
This pull request makes the heatmap sample sort order follow the manually-sorted sample list.

### DIFF
--- a/interface/src/app/analyze/analysis/analysis.js
+++ b/interface/src/app/analyze/analysis/analysis.js
@@ -48,6 +48,9 @@ AnnotationType, SampleBin) {
         ]
       }, {
         // this dataset "streamed" in via ngVega from heatmapData
+        'name': 'samples'
+      }, {
+        // this dataset "streamed" in via ngVega from heatmapData
         'name': 'sample_objects'
       }, {
         // compute minimum normalized value for each node (across samples)
@@ -183,8 +186,9 @@ AnnotationType, SampleBin) {
   };
 
   // populate sample details
-  for (var i = 0; i < SampleBin.samples.length; i++) {
-    SampleBin.getSampleDetails(SampleBin.samples[i]);
+  // FIXME implement this loop as a method on SampleBin? (prob. with caching)
+  for (var i = 0; i < SampleBin.heatmapData.samples.length; i++) {
+    SampleBin.getSampleDetails(SampleBin.heatmapData.samples[i]);
   }
 
   SampleBin.getActivityForSampleList($scope.analysis);
--- a/interface/src/app/analyze/analysis/analysis.js
+++ b/interface/src/app/analyze/analysis/analysis.js
@@ -80,10 +80,10 @@ AnnotationType, SampleBin) {
             'expr': 'datum.normval - datum.node_summary.min_normval'
           }, {
             'type': 'lookup',
-            'on': 'sample_objects',
-            'onKey': 'id',
+            'on': 'samples',
+            'onKey': 'data',
             'keys': ['sample'],
-            'as': ['sample_object']
+            'as': ['sample_order']
           }
         ]
       }
@@ -98,7 +98,11 @@ AnnotationType, SampleBin) {
       }, {
         'name': 'samples',
         'type': 'ordinal',
-        'domain': {'data': 'activity_normalized', 'field': 'sample'},
+        'domain': {
+          'data': 'activity_normalized',
+          'field': 'sample_order._id',
+          'sort': true
+        },
         'range': 'height'
       }, {
         'name': 'minvals',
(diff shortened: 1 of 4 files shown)


dongbohu:
I have finished reviewing this PR. The changes on **analysis.js** are mainly grouping data in Vega. I don't have much to say about it. One question about the changes on **sampleBin.js**: what is the advantage of putting `samples` array in **heatmapData**? Is it a container into which you will dump more data in the future?

mhuyck:
Hmm I thought I made a comment about that. Ah, not a comment in the code but on the change set when I checked in. The `samples` array tracks the order of the samples. The `heatmapData` is a specially structured object with all data we need to pass to Vega. Vega cannot access any data that aren't available in this structure, so in order to have the visualization make use of the sample order I needed to move the `samples` array inside it.

mhuyck:
The changes in `analysis.js` can be thought of as "merging" the newly-introduced samples dataset into the `activity_normalized` dataset, which is what ultimately feeds the rendering instructions in the "marks" section of the json-format Vega "spec."

I also had to reconstruct the labels on the y-axis. These had been generated for me before, but it seems that the axis labels must be drawn from the same data used for computing the y-coordinate position of the marks, which is what I had to change (to an internal `_id`, not meaningful to a user) to get the sort order we want.

dongbohu:
Thank you for the clarification. I should have read your commit comments more carefully.

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -124,21 +131,11 @@ AnnotationType, SampleBin) {
...
-        // "properties": {
-        //   "labels": {
-        //     // "text": {"field": "sample_object.ml_data_source"}
-        //     "text": {"template": "test
-        // {{datum.data.sample_object.ml_data_source}}"}
-        //   }
-        // }
-      }
-    ],
+    // 'axes': [
```
rzelayafavila: Are we not going to use this 'axes' value (because of the new method of building the axes labels)? Will we ever need these lines again?

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -95,7 +98,11 @@ AnnotationType, SampleBin) {
       }, {
         'name': 'samples',
         'type': 'ordinal',
-        'domain': {'data': 'activity_normalized', 'field': 'sample'},
+        'domain': {
+          'data': 'activity_normalized',
+          'field': 'sample_order._id',
```
rzelayafavila: What does this sample_order._id field give us? Is it an index-like value telling us where each sample is supposed to go in the ordered set?

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -163,11 +160,28 @@ AnnotationType, SampleBin) {
...
+        'type': 'text',
+        'from': {'data': 'samples'},
+        'properties': {
+          'update': {
+            'text': {'field': 'data'},
+            'x': {'value': -5},
+            'y': {
+              'scale': 'samples',
+              'field': '_id',
+              'offset': 18
```
rzelayafavila: 'offset' here and 'value' above in line 174 seem a little bit like magic numbers here. Where are they coming from and how do we decide on their values?

rzelayafavila:
Ok, I made some comments but overall I think it looks good!

interface/src/app/analyze/analysis/analysis.js
mhuyck: Exactly. Vega can use a list of matching objects or just a plain list as a data source. The way it uses a plain list is by converting it to a list of objects. In either case, Vega captures the original list order by adding an `_id` field on to each object. For more details, see: https://github.com/vega/vega/wiki/Data

interface/src/app/analyze/analysis/analysis.js
mhuyck: Good question. Are you wondering if I should just strip them out? I spent so long trying to get 'axes' to work that I thought I'd leave the "useless" comment here to remind us not to go down that path again.

interface/src/app/analyze/analysis/analysis.js
rzelayafavila: Ok, cool!

mhuyck:
Okay, @rzelayafavila, I think I've covered everything you brought up. Let me know if you agree!

interface/src/app/analyze/analysis/analysis.js
rzelayafavila: Oh I see - should we then put a small comment before this code block saying that? Maybe something like - "This was the old way of constructing the axes and labels - stay away from this approach as we now use this much better way"?

interface/src/app/analyze/analysis/analysis.js
rzelayafavila: Ok!

rzelayafavila:
Ok @mhuyck - I just posted one more place where a comment might be helpful, but it is not totally necessary. I think it is pretty much done though!

mhuyck:
Oh, good call. I'll make that more clear.

rzelayafavila:
Ok, looks great to me @mhuyck ! 👍

PR #39, opened 2016-11-18
Rewriting the whole gene.searchMany module to make it better comply with Angular's best practices, and make each part of it more loosely coupled with the other parts. 

rzelayafavila:
This included getting rid of a lot of JS code and some templates too.

### DIFF
--- a/interface/src/app/gene/gene-network.tpl.html
+++ b/interface/src/app/gene/gene-network.tpl.html
@@ -3,7 +3,7 @@
     Gene Network
   </h1>
   <div class="well">
-    <gene-search-form></gene-search-form>
+    <gene-search-panel organism="organism"></gene-search-panel>
     <geneset-search-form ng-if="userObj" organism="organism">
     </geneset-search-form>
     <selected-genes-panel></selected-genes-panel>
--- a/interface/src/app/gene/searchMany/gene-search-form.tpl.html
+++ b/interface/src/app/gene/searchMany/gene-search-form.tpl.html
@@ -1,29 +1,11 @@
-<div class="row">
-  <div class="col-xs-12">
-    <div class="panel panel-default">
-      <div class="panel-heading">
-        <h3>
-          Search for Genes
-        </h3>
-      </div>
-      <div class="panel-body">
-        <div class="col-xs-3">
-          <form class="form">
-            <textarea name="genes" rows="7" ng-model="genesToAdd.query"
-              placeholder="Search genes by symbol or Entrez ID">
-            </textarea>
-            <p class="bg-danger">{{ errors }}</p>
-            <br>
-            <button class="btn btn-primary" ng-click="searchGenes()"
-              ng-disabled="loadingSearchResults">
-              Search <i class="fa fa-search"></i>
-            </button>
-          </form>
-        </div>
-        <div class="col-xs-9">  
-          <search-result-table></search-result-table>
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
+<form class="form">
+  <textarea name="genes" rows="7" ng-model="geneQueries"
+    placeholder="Search genes by symbol or Entrez ID">
+  </textarea>
+  <p class="bg-danger">{{ errors }}</p>
+  <br>
+  <button class="btn btn-primary" ng-click="searchGenes()"
+    ng-disabled="loadingSearchResults">
+    Search <i class="fa fa-search"></i>
+  </button>
+</form>
--- a/interface/src/app/gene/searchMany/gene-search-panel.tpl.html
+++ b/interface/src/app/gene/searchMany/gene-search-panel.tpl.html
@@ -1,0 +1,17 @@
+<div class="panel panel-default">
+  <div class="panel-heading">
+    <h3>
+      Search for Genes
+    </h3>
+  </div>
+  <div class="panel-body">
+    <div class="col-xs-3">
+      <gene-search-form search-results="searchResults" queries="queries">
+      </gene-search-form>
+    </div>
+    <div class="col-xs-9">  
+      <search-result-table search-results="searchResults" queries="queries">
+      </search-result-table>
+    </div>
+  </div>
+</div>
--- a/interface/src/app/gene/searchMany/search-buttonset.tpl.html
+++ b/interface/src/app/gene/searchMany/search-buttonset.tpl.html
@@ -2,14 +2,16 @@
   <button class="btn btn-default" ng-click="removeGene()">
     <i class="fa fa-minus-square"></i> None
   </button>
+
   <button class="btn btn-default" ng-if="previousButtonPages"
     ng-click="previousGenePage()">
       <i class="fa fa-angle-double-left"></i> Previous
   </button>
 
-  <gene-result-button query="{{query}}" search-results="searchResults"
-    queries="queries" ng-repeat="gene in buttonPageGenes" gene="gene">
-  </gene-result-button>
+  <button class="btn btn-default" uib-tooltip="{{ gene.description }}"
+    ng-repeat="gene in buttonPageGenes" ng-click="addToSelectedGenes(gene)">
+      {{ getGeneLabel(gene) }}
+  </button>
 
   <button class="btn btn-default" ng-click="nextGenePage()"
     ng-if="additionalButtonPages">
--- a/interface/src/app/gene/searchMany/search-buttonset.tpl.html
+++ b/interface/src/app/gene/searchMany/search-buttonset.tpl.html
@@ -1,7 +1,18 @@
 <div class="btn-group">
-    <no-result-button query="{{query}}"></no-result-button>
-    <previous-result-button ng-if="previousPages"></previous-result-button>
-    <gene-result-button query="{{query}}" ng-repeat="gene in pageGenes">
-    </gene-result-button>
-    <more-result-button ng-if="additionalPages"></more-result-button> 
+  <button class="btn btn-default" ng-click="removeGene()">
+    <i class="fa fa-minus-square"></i> None
+  </button>
+  <button class="btn btn-default" ng-if="previousButtonPages"
+    ng-click="previousGenePage()">
+      <i class="fa fa-angle-double-left"></i> Previous
+  </button>
+
+  <gene-result-button query="{{query}}" search-results="searchResults"
+    queries="queries" ng-repeat="gene in buttonPageGenes" gene="gene">
+  </gene-result-button>
+
+  <button class="btn btn-default" ng-click="nextGenePage()"
+    ng-if="additionalButtonPages">
+      More <i class="fa fa-angle-double-right"></i>
+  </button>
 </div>
--- a/interface/src/app/gene/searchMany/searchMany.js
+++ b/interface/src/app/gene/searchMany/searchMany.js
@@ -36,226 +36,151 @@ angular.module('adage.gene.searchMany', [
   ;
 })
 
-.factory('SearchResults', ['$rootScope', 'Gene', function($rootScope, Gene) {
-  var queries = [];
-  var searchResults = {};
-  var searchSuccess = null;
 
+// Directive for whole gene search form
+.directive('geneSearchPanel', [function() {
   return {
-    getQueries: function() {
-      return queries;
+    controller: ['$scope', function($scope) {
+      $scope.loadingSearchResults = false;
+
+      $scope.queries = [];
+      $scope.searchResults = {};
+    }],
+    restrict: 'E',
+    scope: {
+      organism: '='
     },
-    getQueryResults: function(query) {
-      return searchResults[query];
-    },
-    getSearchResults: function() {
-      return searchResults;
-    },
-    getSearchSuccess: function() {
-      return searchSuccess;
-    },
-    remove: function(query) { // Remove a query and its associated result
-      queries = queries.filter(function(el) {
-        return el !== query;
-      });
-      delete searchResults[query];
-      $rootScope.$broadcast('results.update');
-    },
-    clear: function() { // Clear the service
-      queries = [];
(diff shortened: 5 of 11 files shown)


rzelayafavila:
P.S. - Since so much changed, I would recommend taking a look at the file diffs using the 'Split' view (on the upper-right area of the 'Files changed' tab).

cgreene:
Wow! A net reduction of at least 100 lines of javascript (especially considering the added comments)!

cgreene:
Can you launch a version of this so that reviewers can see it? I know that was helpful for @mhuyck when looking at @dongbohu's PR.

rzelayafavila:
@cgreene: Sure! I will post a url of it here in a sec.

rzelayafavila:
Ok, @cgreene @dongbohu @mhuyck: http://165.123.67.154:8000/#/gene_search is where it's at.

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -36,226 +36,151 @@ angular.module('adage.gene.searchMany', [
...
-      $rootScope.$broadcast('results.update');
-    },
-    clear: function() { // Clear the service
-      queries = [];
-      searchResults = {};
-      $rootScope.$broadcast('results.update');
-    },
-    size: function() {
-      return queries.length;
+    controller: ['$scope', function($scope) {
```
dongbohu: Here you can use either a controller or a link function to set variables in the directive's scope. The Angular book you have (**AngularJS: Up and Running**) has a good discussion on which one should be used under certain scenario (see **page 216**). Since you are not doing any inter-directive communication here, link function is probably a better choice.

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -36,226 +36,151 @@ angular.module('adage.gene.searchMany', [
...
 
 
 // Directive for whole gene search form
-.directive('geneSearchForm', ['SearchResults', function(SearchResults) {
+.directive('geneSearchForm', [function() {
   return {
-    controller: ['$scope', '$rootScope', 'SearchResults',
-      function($scope, $rootScope, SearchResults) {
-        $scope.loadingSearchResults = false;
+    controller: ['$scope', 'Gene', function($scope, Gene) {
```
dongbohu: It is worth considering whether all directives in this file should use link functions instead of controllers. (I didn't realize the difference until I wrote my own directive recently.)

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -36,226 +36,151 @@ angular.module('adage.gene.searchMany', [
...
-            // search endpoint, this will return *all* gene objects it finds,
-            // without filtering for organism. At the point when we add
-            // genes for more than one species to the database, we will
-            // need to set this $scope.organism to whatever organism is in
-            // the ML model (or genes for many organisms will be returned).
-            qparams['organism'] = $scope.organism;
-          }
-          SearchResults.search(qparams);
-        };
+        $scope.loadingSearchResults = true;
```
dongbohu: Trivial question: **loadingSearchResult** is a boolean flag. Can we give it a better name?

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -264,94 +189,80 @@ angular.module('adage.gene.searchMany', [
...
         };
       }
     ],
+    scope: {
+      query: '@',
+      gene: '=',
+      searchResults: '=',
+      queries: '='
+    },
     replace: true,
```
dongbohu: Not really related to this PR: The keyword **replace** has been deprecated since Angular 1.3. You may want to avoid it.

interface/src/app/gene/searchMany/searchMany.js
mhuyck: Can you suggest something you think would be better, @dongbohu? I like that it is very descriptive, personally.

interface/src/app/gene/searchMany/searchMany.js
mhuyck: Good to know. It appears this change is somewhat controversial: http://stackoverflow.com/questions/24194972/why-is-replace-deprecated-in-angularjs

interface/src/app/gene/gene-network.tpl.html
```diff
@@ -3,7 +3,7 @@ <h1 class="page-header">
     Gene Network
   </h1>
   <div class="well">
-    <gene-search-form></gene-search-form>
+    <gene-search-panel organism="organism"></gene-search-panel>
```
mhuyck: Interesting. I just noticed that the template corresponding to this directive is named `gene-search-form.tpl.html` rather than `geneSearchForm.tpl.html`. I didn't know that either style could work! The tutorials I used when learning Angular all had this `camelCase.tpl.html` template file name style mixed with `<camel-case>...` directive name style.

Since you have the hyphenated style of naming everywhere, I guess it works! Wondering if we should standardize on one approach vs. the other. What are your thoughts, @dongbohu and @rzelayafavila?

interface/src/app/gene/searchMany/gene-search-form.tpl.html
```diff
@@ -1,29 +1,11 @@
...
+  <textarea name="genes" rows="7" ng-model="geneQueries"
+    placeholder="Search genes by symbol or Entrez ID">
+  </textarea>
+  <p class="bg-danger">{{ errors }}</p>
+  <br>
+  <button class="btn btn-primary" ng-click="searchGenes()"
+    ng-disabled="loadingSearchResults">
+    Search <i class="fa fa-search"></i>
+  </button>
+</form>
```
mhuyck: I really like how this is now a tightly-focused "widget". Nice work!

interface/src/app/gene/searchMany/gene-search-form.tpl.html
```diff
@@ -1,29 +1,11 @@
...
-        </div>
-        <div class="col-xs-9">  
-          <search-result-table></search-result-table>
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
+<form class="form">
+  <textarea name="genes" rows="7" ng-model="geneQueries"
```
mhuyck: Should this "magic number" 7 be a parameter that can be specified in the `<gene-search-form>` directive (but default to 7)? That way the `gene-search-panel`, which is responsible for layout of this widget, can specify another "size," if needed.

interface/src/app/gene/searchMany/gene-search-form.tpl.html
```diff
@@ -1,29 +1,11 @@
...
+    placeholder="Search genes by symbol or Entrez ID">
```
mhuyck: Similar to the above comment, I like to make the placeholder text a parameter that can be specified. Just a suggestion.

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
```diff
@@ -0,0 +1,17 @@
...
+    <div class="col-xs-3">
+      <gene-search-form search-results="searchResults" queries="queries">
+      </gene-search-form>
+    </div>
+    <div class="col-xs-9">  
+      <search-result-table search-results="searchResults" queries="queries">
+      </search-result-table>
+    </div>
+  </div>
+</div>
```
mhuyck: I really like how this is now tightly focused solely on layout of these two widgets. We should strive for this clean separation of responsibilities wherever we can. Even the naming (`panel` here vs. `form`, above) is helpfully clear.

interface/src/app/gene/searchMany/search-buttonset.tpl.html
```diff
@@ -1,7 +1,18 @@
...
+  </button>
+
+  <gene-result-button query="{{query}}" search-results="searchResults"
+    queries="queries" ng-repeat="gene in buttonPageGenes" gene="gene">
+  </gene-result-button>
+
+  <button class="btn btn-default" ng-click="nextGenePage()"
+    ng-if="additionalButtonPages">
+      More <i class="fa fa-angle-double-right"></i>
+  </button>
```
mhuyck: I take it all of these buttons always appear together as a button set. If so, I think this strikes a great balance between reusable module and the "granularity" of re-use.

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -36,226 +36,151 @@ angular.module('adage.gene.searchMany', [
...
-.directive('geneSearchForm', ['SearchResults', function(SearchResults) {
+.directive('geneSearchForm', [function() {
   return {
-    controller: ['$scope', '$rootScope', 'SearchResults',
-      function($scope, $rootScope, SearchResults) {
-        $scope.loadingSearchResults = false;
+    controller: ['$scope', 'Gene', function($scope, Gene) {
+      $scope.errors = null;
+
+      $scope.searchGenes = function() {
```
mhuyck: Agree that this belongs in the `form` widget, not in the `panel` layout. Good separation of concerns.

interface/src/app/gene/searchMany/searchMany.js
dongbohu: Good to know. I heard it was deprecated but never dug deeper.

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -36,226 +36,151 @@ angular.module('adage.gene.searchMany', [
...
-        });
+          }
+        };
 
-        scope.$on('results.loadingSearchResults', function() {
-          scope.loadingSearchResults = true;
+        $scope.$watchCollection('searchResults', function() {
+          $scope.queries = Object.keys($scope.searchResults);
+          $scope.totalResults = $scope.queries.length;
+          $scope.resultsForPage = $scope.updatePageGenes();
```
mhuyck: I had forgotten about `$watchCollection`. This seems like a reasonable way to keep this section updated.

interface/src/app/gene/gene-network.tpl.html
dongbohu: As far as I know, Angular only cares about whether the file specified by **templateUrl** exists and can be loaded, it doesn't really enforce any filename rules. The suffix "**.tpl.html**" is actually enforced by ngBoilerplace in **build.config.js** (lines 31-32). So I don't think we need to stick to a certain format, unless you guys want to.

mhuyck:
I really like that Code Climate is keeping us DRY, but it sure doesn't seem to handle boilerplate-related repetition very well. Maybe there is a tuning parameter (mass?) we can adjust to make it a little less obsessive.

cgreene:
@mhuyck : agree that it might be customizable, but we might carefully consider this scholarly contribution before spending a lot of time dedicated to improving it:
https://xkcd.com/1205/

I usually just take a quick peek and say: codeclimate's nuts in this case.

rzelayafavila:
@dongbohu: Thank you - I am just checking these issues and making sure the code is as good as it can be for this PR (and create the appropriate issues for outstanding things) - I will probably merge pretty soon after I finish this.

interface/src/app/gene/searchMany/searchMany.js
rzelayafavila: I agree - I had totally forgotten about it until this PR, but it is a great functionality that Angular provides.

interface/src/app/gene/searchMany/searchMany.js
mhuyck: Hah! What's wrong with `foobar`? :-)  I once had a classmate who liked to use the number `17` as a variable in his calculus problems.

On this particular variable name, I'd say that `loadingSearchResults` is a good description of state. A function that did that would be called something like `loadSearchResults`. A subtle distinction, perhaps, but the difference is very clear to me.

Something in between such as `resultsLoading` would be shorter but just as clear, in my opinion.

dongbohu:
About @mhuyck's question on codeclimate, I actually customized a duplication threshold on python programs to 50, (which is part of the reason why codeclimate is not so paranoid on python programs), but didn't do it for JavaScript programs. I will add that parameter to make codeclimate less noisy.

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
rzelayafavila: Glad this makes it more clear!

interface/src/app/gene/searchMany/gene-search-form.tpl.html
rzelayafavila: True - I agree that both can/maybe should be a parameter. I will make an issue for this.

interface/src/app/gene/gene-network.tpl.html
rzelayafavila: Yep, I think (as @dongbohu said) you can specify the location & name of the template file in whatever format (but build.config.js will care about only including template files that end in 'tpl.html', as we have it currently configured). I think the only other thing that needs to follow the strict hyphens <-> camelCase connection is the name of the directive itself (camelCase) and the name of the element it will replace in the html template (hyphens). I am totally fine if we decide to standardize on a specific format for template names though.

interface/src/app/gene/searchMany/searchMany.js
rzelayafavila: Ok, I think I will keep this variable name for now, but keep an eye out and see if it is confusing to anyone else, and if so I will change it. I will also keep this in mind when naming more variables in the future.

rzelayafavila:
Ok, I think I addressed all the concerns and filed the pertaining issue! I decided not to move the code in the directives' ``controller`` function to the ``link`` function since I think this code interacts with the functionality of parent- and sibling- directives. I also removed the one directive that I would have done this code move for, geneResultButton. This was also the directive that included the ``replace: true`` line, so we don't need to worry about that anymore either. AND the code got even shorter!!

PR #43, opened 2016-11-29
Backend Change for "Experiments with Activity Range" feature on Node Page

dongbohu:
This PR includes changes in the following files:

* `api.py`: added a new **node** filter in **ExperimentResource**;
* `tests.py`: created a test case for the filter, also removed a few occurrences of trailing backslashes;
* `models.py`:  corrected a typo in **Activity** model.

### DIFF
--- a/adage/analyze/models.py
+++ b/adage/analyze/models.py
@@ -173,7 +173,7 @@ class Activity(models.Model):
     value = models.FloatField()
 
     def __unicode__(self):
-        return "Sample %s at Node %s with value %f" % (self.sample.sample,
+        return "Sample %s at Node %s with value %f" % (self.sample.name,
                                                        self.node.name,
                                                        self.value)
 
--- a/adage/analyze/tests.py
+++ b/adage/analyze/tests.py
@@ -805,11 +805,11 @@ class APIResourceTestCase(ResourceTestCaseMixin, TestCase):
         # Build relationship between the first sample and the last experiment:
         Sample.objects.first().experiments.add(Experiment.objects.last())
 
-        # Confirm that a random node is related to both the experiment that is
+        # Confirm that a random node is related to both the experiment that was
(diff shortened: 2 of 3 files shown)


dongbohu:
I tweaked the filter's implementation a few times and the current one is much shorted than it was.

adage/analyze/models.py
```diff
@@ -173,7 +173,7 @@ class Activity(models.Model):
     value = models.FloatField()
 
     def __unicode__(self):
-        return "Sample %s at Node %s with value %f" % (self.sample.sample,
+        return "Sample %s at Node %s with value %f" % (self.sample.name,
```
mhuyck: Nice catch. The old way with `.sample` should work, but give the result from calling `__unicode__`, I believe.

adage/analyze/tests.py
```diff
@@ -381,18 +381,18 @@ def setUp(self):
...
         self.create_activities(self.node_counter)
 
-        self.activityURI = self.baseURI + "activity/" + \
-            str(self.random_object(Activity).id) + "/"
-        self.activity_sample_URI = self.baseURI + "activity/?sample=" + \
-            str(self.random_object(Sample).id)
+        self.activityURI = (self.baseURI + "activity/" +
+                            str(self.random_object(Activity).id) + "/")
+        self.activity_sample_URI = (self.baseURI + "activity/?sample=" +
+                                    str(self.random_object(Sample).id))
```
mhuyck: More `\` elimination. Huzzah! 🎉

dongbohu:
Thank you Matt. I will take care of the issues you pointed out.

dongbohu:
I also removed `format=json` from the other test cases. It makes the code cleaner. Thank you @mhuyck!

mhuyck:
LGTM. Thanks for noticing the other `format=json` lines... I thought we had already dealt with that so I didn't even bother to look!

PR #45, opened 2016-12-08
Heatmap clustering

mhuyck:
This update implements heatmap clustering and, since node clustering takes a while, restores status updates for the user (temporarily removed when code was reorganized to cache activity data) so they understand that the browser is not freezing up during lengthy operations.

### DIFF
--- a/interface/package.json
+++ b/interface/package.json
@@ -14,7 +14,9 @@
   "engines": {
     "node": ">=4.0"
   },
-  "dependencies": {},
+  "dependencies": {
+    "hclusterjs": "^1.3.0"
+  },
   "devDependencies": {
     "grunt": "^1.0.1",
     "grunt-bump": "^0.8.0",
--- a/interface/src/app/analyze/analysis/analysis.js
+++ b/interface/src/app/analyze/analysis/analysis.js
@@ -48,6 +48,9 @@ AnnotationType, SampleBin) {
         'name': 'samples'
       }, {
         // this dataset "streamed" in via ngVega from heatmapData
+        'name': 'nodeOrder'
+      }, {
+        // this dataset "streamed" in via ngVega from heatmapData
         'name': 'sample_objects'
       }, {
         // compute minimum normalized value for each node (across samples)
@@ -61,7 +64,8 @@ AnnotationType, SampleBin) {
           }
         ]
       }, {
-        // now subtract the minimum sample value from each node
+        // now subtract the minimum sample value from each node and
+        // lookup the order for drawing samples and nodes
         'name': 'activity_normalized',
         'source': 'activity',
         'transform': [
@@ -81,6 +85,12 @@ AnnotationType, SampleBin) {
             'onKey': 'data',
             'keys': ['sample'],
             'as': ['sample_order']
+          }, {
+            'type': 'lookup',
+            'on': 'nodeOrder',
+            'onKey': 'data',
+            'keys': ['node'],
+            'as': ['node_order']
           }
         ]
       }
@@ -90,7 +100,11 @@ AnnotationType, SampleBin) {
       {
         'name': 'nodes',
         'type': 'ordinal',
-        'domain': {'data': 'activity_normalized', 'field': 'node'},
+        'domain': {
+          'data': 'activity_normalized',
+          'field': 'node_order._id',
+          'sort': true
+        },
         'range': 'width'
       }, {
         'name': 'samples',
@@ -158,7 +172,7 @@ AnnotationType, SampleBin) {
         'from': {'data': 'activity_normalized'},
         'properties': {
           'enter': {
-            'x': {'scale': 'nodes', 'field': 'node'},
+            'x': {'scale': 'nodes', 'field': 'node_order._id'},
             'width': {'scale': 'nodes', 'band': true},
             'y': {'scale': 'samples', 'field': 'sample_order._id'},
             'height': {'scale': 'samples', 'band': true},
--- a/interface/src/app/analyze/analysis/analysis.js
+++ b/interface/src/app/analyze/analysis/analysis.js
@@ -19,8 +19,27 @@ angular.module('adage.analyze.analysis', [
 'Activity', 'AnnotationType', 'SampleBin',
 function AnalysisCtrl($scope, $log, $location, $q, Sample, Activity,
 AnnotationType, SampleBin) {
+  $scope.analysis = {
+    status: '',
+    // TODO these exampleCols are temporarily hard-coded until a column chooser
+    // feature can be added
+    exampleCols: [
+      {'typename': 'genotype'},
+      {'typename': 'medium'},
+      {'typename': 'strain'}
+    ]
+  };
+
   // give our templates a way to access the SampleBin service
   $scope.sb = SampleBin;
+
+  // wrap some SampleBin features to implement status updates
+  $scope.clusterNodes = function() {
+    $scope.analysis.status = 'clustering nodes (this will take a minute)';
+    SampleBin.clusterNodes().then(function() {
+      $scope.analysis.status = '';
+    });
+  };
 
   // these options are important for making ngSortable work with tables
   $scope.sortableOptions = {
@@ -199,17 +218,6 @@ AnnotationType, SampleBin) {
           }
         }
       }
-    ]
-  };
-
-  $scope.analysis = {
-    status: '',
-    // TODO these exampleCols are temporarily hard-coded until a column chooser
-    // feature can be added
-    exampleCols: [
-      {'typename': 'genotype'},
-      {'typename': 'medium'},
-      {'typename': 'strain'}
     ]
   };
 
--- a/interface/src/app/analyze/analysis/analysis.js
+++ b/interface/src/app/analyze/analysis/analysis.js
@@ -6,6 +6,7 @@
 angular.module('adage.analyze.analysis', [
   'adage.analyze.sampleBin',
   'adage.analyze.sample',
+  'statusBar',
   'ngVega',
   'ngResource'
 ])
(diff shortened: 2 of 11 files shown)


mhuyck:
@dongbohu, @rzelayafavila: Just noticed the new GitHub "reviewers" feature, so moved you from "Assignees" to the Reviewers section.

interface/package.json
```diff
@@ -14,7 +14,9 @@
   "engines": {
     "node": ">=4.0"
   },
-  "dependencies": {},
+  "dependencies": {
+    "hclusterjs": "^1.3.0"
```
dongbohu: It is possible to specify a git repository in **bower.json** for installation. I am not sure whether we should do it to have hclusterjs package in `vendors/` directory instead of in `node_modules/`.

dongbohu: https://github.com/bower/spec/blob/master/json.md#dependencies

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -187,21 +221,20 @@ AnnotationType, SampleBin) {
...
   // FIXME implement this loop as a method on SampleBin? (prob. with caching)
+  $scope.analysis.status = 'Retrieving sample details';
+  var pArrSamples = [];
   for (var i = 0; i < SampleBin.heatmapData.samples.length; i++) {
-    SampleBin.getSampleDetails(SampleBin.heatmapData.samples[i]);
+    pArrSamples.push(
+      SampleBin.getSampleDetails(SampleBin.heatmapData.samples[i])
+    );
   }
+  $q.all(pArrSamples).then(function() {
```
dongbohu: I like your way of using `$q.all()` to handle the promises. On the other hand, do you think it will be worth the time to add a new API on the backend to retrieve all data of samples that are related to a certain node? Then the frontend will be simpler. (See my latest PR, which adds a new API to get the data of all samples that are related to a given experiment.)

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -75,48 +78,116 @@ function($log, $cacheFactory, $q, Sample, Activity) {
...
+      // reformat data from heatmapData.activity to a form that can be used
+      // by hcluster.js: need a separate array of objects for each node
+      var firstSampleNodes = this.activityCache.get(
+        this.heatmapData.samples[0]
+      );
+      var retval = firstSampleNodes.map(function(val, i, arr) {
+        var nodeObject = {
+          'id': val.node
+        };
+        nodeObject.activity = this.heatmapData.samples.map(
```
dongbohu: Can you add more comments here? I found the map inside map hard to follow.

interface/src/common/statusBar/statusBar.js
```diff
@@ -0,0 +1,12 @@
+angular.module('statusBar', [])
+
+.directive('statusBar', function() {
+  return {
+    restrict: 'E',
+    scope: {
+      msgText: '='
```
dongbohu: I like this `<staus-bar>` gadget. Since **msgText** is read-only, we can use **@** instead of **=** here (then you have change the html above into:
```
<status-bar msgText="{{sample.status}}"></status-bar>
```

Another way is use the new **component()** function, which was introduced since Angular 1.5 for simple directives.
https://docs.angularjs.org/guide/component
(I am not saying that **component** is better than **directive** here, but I think it is good to know this option.)

interface/src/common/statusBar/statusBar.js
```diff
@@ -0,0 +1,12 @@
+angular.module('statusBar', [])
```
mhuyck: Right: I realized we had several places that could benefit from having a "status bar" done in a standardized way, but it's not adage-server specific, so I put it into `interface/src/common`.

The camelCase naming follows the convention throughout ngBoilerplate of using camelCase in JavaScript source and hyphenated-names in HTML. I have done nothing special to make the translation between the two work, so this appears to be a feature of Angular and, I assume, the preferred way to do it. Please correct me if I'm wrong!

interface/src/app/analyze/sample/sampleDetail.tpl.html
```diff
@@ -10,10 +10,7 @@
     sample.related_experiments[0].accession, sample.results.ml_data_source)
   }}">{{sample.results.ml_data_source}}</a></small>
 </h3>
-<p ng-if="sample.status" class="alert alert-warning" role="alert">
-  Status: {{sample.status}}<br>
-  If this message persists, please try refreshing your browser.
-</p>
+<status-bar msg-text="sample.status"></status-bar>
```
mhuyck: Thank you! If there are other places that can use this, we should probably standardize across our codebase.

interface/package.json
mhuyck: I believe this is the first NPM module we are deploying into our front-end code rather than using as a tool operating in our build process. I think this means it belongs in `dependencies` rather than `dev-dependencies`.

Would there be some benefit from changing things around to have it appear in `vendors/` instead?

interface/src/app/analyze/analysis/analysis.js
mhuyck: I thought about this for a while. Retrieving samples independently allows us to do caching at the level of samples. This means that if a user adds or removes a sample from the heatmap we can rebuild it much more quickly than we could otherwise.

In addition, although I haven't implemented it yet, we can ultimately begin fetching the analysis data into the cache as soon as an individual sample is added to the `sampleBin` (rather than waiting until the `analysis` view is opened, as we do now). This will make the apparent performance of the UI even snappier.

No need for a new API, by the way. The `Activity` API can already filter on `sample_id__in`. Before I implemented the cache this is the way it was done, in fact. Without caching, it *is* more efficient to get all of the activity data in a single call, but with the cache, I think this way is better.

Does that approach make sense?

interface/src/common/statusBar/statusBar.js
mhuyck: I didn't know about `component()` yet... thank you for pointing it out! I'll have a look and see if it makes sense. If not, I agree with your idea of using `@` rather than `=`.

- [ ] look into using `component()` here

interface/src/app/analyze/analysis/analysis.js
mhuyck: Whoops. I should have looked more closely at the context here... The relevant API is `Sample`, not `Activity`. But the logic I'm getting at is the same. Caching is not yet implemented for this `Sample` data, but once I have generalized that code, the UI will cache this sample data as well as the activity data so the same logic applies.

interface/src/common/statusBar/statusBar.js
mhuyck: As usual, StackOverflow provides a good explanation. Apparently the hyphenated-names are sometimes called kebab-case. LOL!

http://stackoverflow.com/questions/33460274/angularjs-directive-naming-conventions

Where I probably saw this was in the Angular tutorial here: https://docs.angularjs.org/tutorial/step_00#what-is-the-code-doing-

interface/package.json
dongbohu: There is probably no any other benefits except the consistency. I am okay with hosting this js library in `node_modules/`.

interface/src/app/analyze/analysis/analysis.js
dongbohu: Sure, thank you for your explanation.

interface/src/common/statusBar/statusBar.js
dongbohu: If you don't have time now, don't worry about the option of `component()` function. I am reading Angular's official doc recently and feel that it is an interesting feature to share with you guys.

interface/package.json
mhuyck: Okay, sounds good. Keep in mind that the `vendors/` and `node_modules/` directories exist only for the build and are not used by the deployed code. The minifying step used for the final "packaging" of the files will merge this file along with the others pulled from `vendors/` into our own customize `adage-ver#.js` file (currently `interface/bin/assets/adage-0.2.1a.js`) so, in the end, this file does actually get treated the same as the others in `vendors/`.

dongbohu:
@mhuyck: Please note that the PR that I merged yesterday updated **build.config.js** too. You may want to update yours before merging. Otherwise you will have to resolve it later.

mhuyck:
Thank you for the heads up @dongbohu! I have made it a practice to do a rebase just before merging, so I'll do that here too.

interface/src/common/statusBar/statusBar.js
mhuyck: The component approach is totally the way to go here. I don't want to drag out this PR right now, though, so I'm going to track this as an issue. There are a bunch of other places that could benefit from this simplification in our code, too, so I'm going to suggest we tackle those all at once.

Thank you so much for calling my attention to that feature of Angular, @dongbohu!

dongbohu:
@mhuyck: I accidentally sent my last comment under the new issue that you created. In case you didn't see it, here it is:

Sounds good to me. I only have two minor concerns about this PR:
(1) It may be better to use @ in the scope (instead of =) when possible.
(2) Add some comments on the map inside map function calls.
After you address them, please feel free to merge it.

interface/src/common/statusBar/statusBar.js
mhuyck: I just tried converting from `msgText: '='` to use `'@'` instead. Whether or not I also make the corresponding change you mention to the HTML, the changes break the whole `statusBar`. I am not sure why yet, but the message text does not appear (the uninterpreted variable name appears instead) and the status bar does not disappear when it should either.

I don't have time to debug this now, so I'm reverting to using `'='` since that works and I'm making a note to follow up in the issue I just created.

mhuyck:
@dongbohu: see the updates I've pushed for (2). I attempted the approach you suggested in (1) but it broke the `statusBar` so I reverted it. I've made a note to follow up on that in the issue about converting to Component.

mhuyck:
@dongbohu Please let me know if my updated comments on the nested Array.prototype.map() code help clarify things.

PR #46, opened 2016-12-09
Backend change (again) for new features on Node page

dongbohu:
Here is another PR on adage-server backend. It includes:
* **api.py**: implemented a new filter **experiment** in `SampleResource` class;
* **tests.py**: added a new test case for this filter, also refactored the test case `test_node_filter_in_experiment()`.

### DIFF
--- a/adage/analyze/api.py
+++ b/adage/analyze/api.py
@@ -142,6 +142,9 @@ class SampleResource(ModelResource):
         allowed_methods = ['get']
         experiments_allowed_methods = allowed_methods
         annotations_allowed_methods = allowed_methods
+        filtering = {
+            'experiment': ('exact', ),  # Implemented in apply_filters()
+        }
 
     def prepend_urls(self):
         return [
@@ -233,6 +236,20 @@ class SampleResource(ModelResource):
         response['Content-Disposition'] = (
             'attachment; filename="sample_annotations.tsv"')
         return response
+
+    def apply_filters(self, request, applicable_filters):
+        """
+        Implementation of "experiment" filter, which allows the API to
+        get all samples that are related to a given experiment (whose
+        primary key is "accession").
+        """
+        object_list = super(SampleResource, self).apply_filters(
+            request, applicable_filters)
+        experiment = request.GET.get('experiment', None)
+        if experiment:
+            samples = Experiment.objects.get(pk=experiment).sample_set.all()
+            object_list = object_list.filter(pk__in=samples)
+        return object_list
 
 
 class MLModelResource(ModelResource):
--- a/adage/analyze/api.py
+++ b/adage/analyze/api.py
@@ -247,7 +247,12 @@ class SampleResource(ModelResource):
             request, applicable_filters)
         experiment = request.GET.get('experiment', None)
         if experiment:
-            samples = Experiment.objects.get(pk=experiment).sample_set.all()
+            try:
+                e = Experiment.objects.get(pk=experiment)
+            except ObjectDoesNotExist:
+                # Return an empty set if the experiment is not found.
+                return Sample.objects.none()
+            samples = e.sample_set.all()
             object_list = object_list.filter(pk__in=samples)
         return object_list
 
--- a/adage/analyze/tests.py
+++ b/adage/analyze/tests.py
@@ -789,30 +789,67 @@ class APIResourceTestCase(ResourceTestCaseMixin, TestCase):
 
         self.call_non_get_API(uri)  # Test non-get methods too.
 
+    def create_extra_experiments(self):
+        """
+        Generate a few more experiements, one of which is returned
+        (and will be used by test cases later).
+        """
+        # Create a new experiment:
+        exp2 = Experiment.objects.create(
+            accession="z1", name="z2", description="z3")
+        # Build relationship between the first sample and exp2:
+        Sample.objects.first().experiments.add(exp2)
+        # Create a few more random experiements that are not related to
+        # any samples:
+        factory.create(Experiment, 3)
+        return exp2
+
     def test_node_filter_in_experiment(self):
         """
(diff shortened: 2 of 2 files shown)


dongbohu:
One of the reasons why I changed **test_node_filter_in_experiment** in **tests.py** is because the primary key in Experiment table is "accession" (which is a string), so the last experiment may not necessarily be one of the random experiments generated by `fixtureless`.

dongbohu:
Just realized that another approach to get all samples related to a certain experiment is to change this line in `ExperimentResource` class:
``` 
sample_set = fields.ManyToManyField('analyze.api.SampleResource', 'sample_set')
```
into:
```
sample_set = fields.ManyToManyField('analyze.api.SampleResource', 'sample_set', full=True)
```
But depending on how big the sample and sample annotation tables are, this database query could be slow due to the implicit table join between Sample and SampleAnnotation.

dongbohu:
If performance is not a problem, then this new API will also make @mhuyck's **Experiment/Sample** frontend more compact, because both the experiment data and all related sample data (instead of just the sample data's URI) will be available to the frontend after one GET.

dongbohu:
Never mind. After some tests, I found that adding `full=True` in **ExperimentResource** is probably not a good idea, because it will slow down all APIs in this class. The current implementation seems to have a better balance between performance and the # of API calls needed.

rzelayafavila:
Ok - so this is more of a clarification question: I see that this implements a way for us to get all of the samples (using the sample API endpoint) that are associated with a certain experiment, right? And it looks like the way we ask for the desired experiment is using the Experiment's PK, which is its accession number (and therefore a string instead of an auto-incrementing integer)?

rzelayafavila:
Also @dongbohu - I agree, I can see how doing an extra table join by setting ``full=True`` in the ExperimentResource's sample_set would lead to a hit in performance, so I agree this is not the ideal way to go.

dongbohu:
@rzelayafavila: Your clarification is exactly right. Note that the **Experiment** model is not using auto-incrementing integer as the primary key.

adage/analyze/api.py
```diff
@@ -234,6 +237,20 @@ def get_annotations(request=None, **kwargs):
...
+        """
+        Implementation of "experiment" filter, which allows the API to
+        get all samples that are related to a given experiment (whose
+        primary key is "accession").
+        """
+        object_list = super(SampleResource, self).apply_filters(
+            request, applicable_filters)
+        experiment = request.GET.get('experiment', None)
+        if experiment:
+            samples = Experiment.objects.get(pk=experiment).sample_set.all()
```
mhuyck: What happens if the experiment key passed in turns out to be invalid? Do we need error handling in here, or does TastyPie do something smart for us?

adage/analyze/tests.py
```diff
@@ -789,30 +789,67 @@ def test_heavy_genes(self):
...
+        related_samples = self.deserialize(resp)['objects']
+        self.assertEqual(len(related_samples), Sample.objects.count())
+
+        # Confirm that only the first sample is related to exp2.
+        filter_uri = "%ssample/?experiment=%s&limit=0" % (
+            self.baseURI, exp2.accession)
+        resp = self.api_client.get(filter_uri)
+        related_samples = self.deserialize(resp)['objects']
+        self.assertEqual(len(related_samples), 1)
+        self.assertEqual(related_samples[0]['id'], Sample.objects.first().id)
```
mhuyck: nice tests!

adage/analyze/tests.py
```diff
@@ -789,30 +789,67 @@ def test_heavy_genes(self):
 
         self.call_non_get_API(uri)  # Test non-get methods too.
 
+    def create_extra_experiments(self):
+        """
+        Generate a few more experiements, one of which is returned
+        (and will be used by test cases later).
```
rzelayafavila: Specifically, this function will create 3 random experiments in the Database using fixtureless and one experiment that is manually created, and will be returned, right? Maybe we should put this in the docstring?

adage/analyze/tests.py
```diff
@@ -789,30 +789,67 @@ def test_heavy_genes(self):
...
+        """
+        # Create a new experiment:
+        exp2 = Experiment.objects.create(
```
rzelayafavila: Why is this called exp2, if it is the first one being created?

adage/analyze/tests.py
dongbohu: The first experiment in the database was actually created from `ModelsTestCase.experiment_data` in setup(), `exp2` is created here with its `accession` key after `accession` key in `ModelsTestCase.experiment_data`. Three more random experiments are then created by fixtureless. (Because these three experiments are random, their `accession` keys may be before or after accession` key in `ModelsTestCase.experiment_data`. They are created to increase the "noise" and ensure that the two tests that follow can survive the noise. )  Only `exp2` will be returned. Does this make sense?

adage/analyze/tests.py
dongbohu: Please see my reply to your first question.

dongbohu:
@rzelayafavila and @mhuyck: Thanks to both of you for your feedback. I have added the error handling code. Please take a look and let me know what you think.

adage/analyze/api.py
```diff
@@ -234,6 +237,25 @@ def get_annotations(request=None, **kwargs):
...
+        experiment = request.GET.get('experiment', None)
+        if experiment:
+            try:
+                e = Experiment.objects.get(pk=experiment)
+            except ObjectDoesNotExist:
+                # Return an empty set if the experiment is not found.
+                return Sample.objects.none()
+            samples = e.sample_set.all()
+            object_list = object_list.filter(pk__in=samples)
+        return object_list
```
mhuyck: Looks like a clean, Pythonish solution to me. Nice!

PR #48, opened 2016-12-14
Gene autocomplete search feature

rzelayafavila:
Includes Angular code and templates that make up the gene autocomplete search feature in the front-end.

### DIFF
--- a/interface/src/app/gene/searchMany/gene-search-panel.tpl.html
+++ b/interface/src/app/gene/searchMany/gene-search-panel.tpl.html
@@ -14,4 +14,9 @@
       </search-result-table>
     </div>
   </div>
+  <div class="panel-footer">
+    <button type="button" class="btn btn-link" ng-click="switchToFew()">
+      Switch to searching with autocomplete.
+    </button>
+  </div>
 </div>
--- a/interface/src/app/gene/searchMany/searchMany.js
+++ b/interface/src/app/gene/searchMany/searchMany.js
@@ -11,31 +11,34 @@ angular.module('adage.gene.searchMany', [
       views: {
         'main': {
           templateUrl: 'gene/gene-network.tpl.html',
-          controller: ['$scope', 'UserFactory',
-            function($scope, UserFactory) {
-              $scope.userObj = null;
-              UserFactory.getPromise().$promise.then(function() {
-                $scope.userObj = UserFactory.getUser();
-              });
+          controller: ['UserFactory', function(UserFactory) {
+            var self = this;
+            self.userObj = null;
+            UserFactory.getPromise().$promise.then(function() {
+              self.userObj = UserFactory.getUser();
+            });
 
-              // TODO: Right now, we are hard-coding this organism
-              // as Pseudomonas (since it is the only one currently
-              // supported by ADAGE). However, as we incorporate
-              // multi-species support, this organism will have to
-              // be obtained from the ML model. This is the same as
-              // the issue in geneSearchForm (also with $scope.organism).
-              $scope.organism = 'Pseudomonas aeruginosa';
+            // TODO: Right now, we are hard-coding this organism
+            // as Pseudomonas (since it is the only one currently
+            // supported by ADAGE). However, as we incorporate
+            // multi-species support, this organism will have to
+            // be obtained from the ML model. This is the same as
+            // the issue in geneSearchForm (also with $scope.organism).
+            self.organism = 'Pseudomonas aeruginosa';
 
-              // '$scope.autocomplete.autocomlete' holds a boolean value,
-              // of whether or not the user wants to use autocomplete
-              // search to look up a few genes. If not, then the autocomplete
-              // search panel will disappear, and the panel to search many
-              // genes will appear. We need to send the value itself inside
-              // an object so that it gets sent properly back and forth
-              // through the corresponding $scopes.
-              $scope.autocomplete = {autocomplete: true};
-            }
-          ]
+            // '$scope.autocomplete' holds a boolean value, of whether or
+            // not the user wants to use autocomplete search to llok up a
+            // few genes. If not, then the autocomplete search panel will
+            // disappear, and the panel to search many genes will appear.
+            // *Note: This boolean value wasn't getting properly propagated
+            // through the scopes of this state and the child directives
+            // if it was just placed in the $scope object. It needed to be
+            // placed either inside a new object (which was then placed in
+            // the $scope object), or making it part of the controller
+            // instance object.
+            self.autocomplete = true;
+          }],
+          controllerAs: 'searchCtrl'
         }
       },
       data: {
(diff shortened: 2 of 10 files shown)


rzelayafavila:
Hmmm, I see that codeclimate is complaining about some code repetition. In this case, I do agree that I could refactor this function and put it in the CommonGeneFuncts service in gene/utils.js, so I will do this.

dongbohu:
It will be very helpful if you can provide a demo page where we can see how the web UI looks like. Thanks.

rzelayafavila:
@dongbohu sure: http://165.123.67.154:8000/#/gene_search

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
```diff
@@ -14,4 +14,9 @@
       </search-result-table>
     </div>
   </div>
+  <div class="panel-footer">
+    <button type="button" class="btn btn-link" ng-click="switchToFew()">
+      Switch to searching with autocomplete.
+    </button>
```
dongbohu: Do you thin kit makes more sense to put this button at the end of the whole page (instead of putting it in the middle of "Search for Genes" section and "Selected Genes" section)?

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -25,6 +25,15 @@ angular.module('adage.gene.searchMany', [
...
               $scope.organism = 'Pseudomonas aeruginosa';
+
+              // '$scope.autocomplete.autocomlete' holds a boolean value,
+              // of whether or not the user wants to use autocomplete
+              // search to look up a few genes. If not, then the autocomplete
+              // search panel will disappear, and the panel to search many
+              // genes will appear. We need to send the value itself inside
+              // an object so that it gets sent properly back and forth
+              // through the corresponding $scopes.
+              $scope.autocomplete = {autocomplete: true};
```
dongbohu: Is there any advantage of defining `$scope.autocomplete` as an object instead of just a boolean?

dongbohu: Never mind, just noticed that your comments answered my question. I am still surprised that `autocomplete` has to be an object in order to allow the proper communication between this directive and its parent.

dongbohu: I did a test and found that the change on a boolean flag that is shared by the parent scope and its directive child scope can be detected correctly. What was the error you got when defining `autocomplete` as a boolean?

dongbohu:
I finished the reviewing and had two (minor) questions. Thanks @rzelayafavila.

interface/src/app/gene/searchMany/searchMany.js
dongbohu: Attached is a tarball file of my test. It is based on an example in Chapter 11 of **AngularJS: Up & Running" book. If you click the buttons **toggle dhu** on index.html, it will toggle a boolean variable and show/hide a line **hello from dhu**. I tested the directive with both link function and controller.
[test.tar.gz](https://github.com/greenelab/adage-server/files/652890/test.tar.gz)

rzelayafavila: Thanks for that @dongbohu - I tried out the example you sent, and I noticed that it uses controllerAs syntax, so I switched the code in here to also use controllerAs syntax instead of $scope - this worked! I am still scratching my head though, as to why this works, just like putting the boolean inside a new object, as opposed to just binding a boolean flag to the $scope object. Following the Pragmatic Programmer's mantra, I feel like this is a bug until I fully understand it, so I am diving into the documentation about this. Do you have any insights on this?

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
rzelayafavila: Possibly - I am going to switch stuff around and see if there is a layout that is more intuitive.

interface/src/app/gene/searchMany/searchMany.js
rzelayafavila: Aha! Thank you @dongbohu for the documentation links you provided on Slack - this is all much more clear now. This doc (https://github.com/angular/angular.js/wiki/Understanding-Scopes) and its link to the video of Misko Hevery on best practices (http://www.youtube.com/watch?v=ZhfUv0spHCY&feature=youtu.be&t=30m) were particularly useful in understanding how I was actually creating a new property in the directive's childScope instead of referring to the desired property in the parentScope and changing it from true to false and back. With this I am now following the best practice of "always have a '.' in your ng-models" (see video). This is something that we should keep in mind going forward and should change in the existing code whenever we see it happening.

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
rzelayafavila: Ok, I played around with this layout and showed it to Jackie for a UX test, and I think it probably makes sense to keep this button/link here for consistency with the autocomplete view (which has it in the same place). Also, it is right below the "Search" button, so that might catch the users' eyes before they click Search.

rzelayafavila:
Ok @dongbohu : I have gotten the bug that we didn't understand at first thanks to your documentation, and the placing of the button/link I think is resolved. Let me know if you see anything else! Thanks!

interface/src/app/gene/searchMany/searchMany.js
rzelayafavila: @dongbohu: Yes, I agree - In fact, the code is already using the controllerAs syntax, I changed it to using that syntax last week because I also thought that that was the best way to go. I think GitHub is just not showing those updates in this specific view, but they will show if you look at the Files changed. Does this look good to you as it is now?

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -11,22 +11,34 @@ angular.module('adage.gene.searchMany', [
...
+            // TODO: Right now, we are hard-coding this organism
+            // as Pseudomonas (since it is the only one currently
+            // supported by ADAGE). However, as we incorporate
+            // multi-species support, this organism will have to
+            // be obtained from the ML model. This is the same as
+            // the issue in geneSearchForm (also with $scope.organism).
+            self.organism = 'Pseudomonas aeruginosa';
+
+            // '$scope.autocomplete' holds a boolean value, of whether or
+            // not the user wants to use autocomplete search to llok up a
```
dongbohu: Typo: "llok" should be "look".
About the bug, one interesting thing is that if you use **ng-show** or **ng-hide** instead of **ng-if**, you won't have this problem, because ng-show/ng-hide doesn't generate a new scope.

rzelayafavila: Ok, thanks, changed it. True that makes sense, but now it is good since we are following the controllerAs syntax.

rzelayafavila:
Ok, thanks @dongbohu.

PR #49, opened 2016-12-20
New feature on Node Page: Experiments with Highest Range

dongbohu:
This is a relatively large PR. It includes the implementation of new **Experiments with Highest Range** feature on Node Page, and a minor change on gene network code. The visual tool that is embeded in the feature  to show the range of activity values for each experiment is using vega-lite, which may not be straightforward to understand. Please let me know if you have any questions.

The web UI can be tested at: http://165.123.67.202/#/node/10998
(Please feel free to change the node ID **10998** to test other nodes.)


interface/src/app/node/node.js
```diff
@@ -92,5 +92,204 @@ angular.module('adage.node', [
...
+
+      // Get activities that are related to the current node:
+      $http.get('/api/v0/activity/?limit=0&node=' + $scope.nodeId)
+        .then(function success(response) {
+          var sampleID;
+          for (var i = 0; i < response.data.objects.length; ++i) {
+            sampleID = response.data.objects[i].sample;
+            $scope.activities[sampleID] = response.data.objects[i].value;
+          }
+          return $http.get('/api/v0/experiment/?limit=0&node=' + $scope.nodeId);
```
cgreene: Can you pass parameters to `$http.get` instead of explicitly building a URL? Might make it a bit friendlier to read.

interface/src/app/node/node.js
```diff
@@ -92,5 +92,204 @@ angular.module('adage.node', [
...
+    templateUrl: 'node/high_range_exp.tpl.html',
+    restrict: 'E',
+    scope: {
+      nodeId: '@'
+    },
+    link: function($scope) {
+      $scope.queryStatus = 'Connecting to the server ...';
+      $scope.activities = {};
+      $scope.experiments = [];
+      $scope.top20Mode = true;
```
cgreene: Can this number be passed in as a parameter to tag `<high-range-exp node-id="{{ctrl.id}}">`? I am imagining that we might want to re-use this in other places with potentially fewer or more experiments shown by default.

interface/src/app/node/node.js
```diff
@@ -92,5 +92,204 @@ angular.module('adage.node', [
...
+        }); // end of $http.get().then() chaining.
+
+      // Event handler when user clicks "+" or "-" icon in front of
+      // an experiment.
+      $scope.toggleExpansion = function(exp) {
+        exp.isExpanded = !exp.isExpanded;
+        // Get samples data if they are not available yet.
+        if (exp.isExpanded && !exp.samples) {
+          exp.sampleStatus = 'Connecting to the server ...';
+          var sampleURI = '/api/v0/sample/?experiment=' + exp.accession;
```
cgreene: Question on building URL vs allowing `$http.get` to use params.

interface/src/app/node/node.js
```diff
@@ -92,5 +92,223 @@ angular.module('adage.node', [
   }]
 )
 
+.directive('highRangeExp', ['$http', '$log', function($http, $log) {
+  return {
+    templateUrl: 'node/high_range_exp.tpl.html',
```
rzelayafavila: Hi @dongbohu: Is this template file for this directive included in the pull request? I can't seem to find it.

dongbohu:
I forgot to add this file in the PR. Will do it either tonight or tomorrow morning.

rzelayafavila:
@dongbohu ok, thanks!

interface/bower.json
```diff
@@ -21,10 +21,13 @@
     "ng-sortable": "~1.3.6",
     "ng-vega": "^1.3.2",
     "d3-tip": "~0.7.1",
-    "angularjs-slider": "^5.8.1"
+    "angularjs-slider": "^5.8.1",
+    "vega-lite": "^1.3.1",
```
rzelayafavila: This more of a clarification point for me: So we need the vega-lite library even though we have already loaded vega itself?

interface/src/app/gene/network/network.js
```diff
@@ -198,7 +198,8 @@ angular.module('adage.gene.network', [
               edgeTip.show(data, target);
             },
             function error(err) {
-              $log.error('Failed to get node info for gene edge: ' + err);
+              $log.error(
+                'Failed to get node info for gene edge: ' + err.statusText);
```
rzelayafavila: Nice! I agree that we should use a '.' in our data binding on the template, especially given what we have recently learned about scopes.

interface/src/app/node/high_range_exp.tpl.html
```diff
@@ -0,0 +1,48 @@
...
+    <b>Experiments with the highest range: </b>{{experiments.length}}
+  </h4>
+  <button class="btn btn-primary" ng-click="showTop()"
+          ng-hide="topMode">Show Top {{topNum}}</button>
+  <button class="btn btn-primary" ng-click="showAll()"
+          ng-show="topMode">Show All</button>
+
+  <br />
+  (Click
+  <i class="fa fa-plus-square fa-lg" aria-hidden="true"></i> or
```
rzelayafavila: As I was playing around with the interface, I was wondering if these icons (on the table below) should be actual buttons so that users intuitively know that they can click on them.

interface/src/app/node/node.js
```diff
@@ -92,5 +92,223 @@ angular.module('adage.node', [
...
+    restrict: 'E',
+    scope: {
+      nodeId: '@',
+      inputTopNum: '@topExp'
+    },
+    link: function($scope) {
+      $scope.queryStatus = 'Connecting to the server ...';
```
rzelayafavila: Are we using ``$scope`` instead of 'controllerAs' syntax here because this code is in the ``link`` function?

interface/src/app/node/node.js
rzelayafavila: So, it looks like ``$http.get()`` can take a second argument after the url, which is a ``config`` object {}. In this object, you can specify these querystring parameters inside a ``params`` key instead of adding them to the end of the url: https://docs.angularjs.org/api/ng/service/$http#usage

rzelayafavila:
Ok @dongbohu, I made a first pass through the PR, and I had a couple of comments and suggestions, but I think it overall looks good!

interface/bower.json
dongbohu: That's right. You probably noticed that **bower,json** does not include `vega` explicitly. It is a dependency of `ng-vega` and `vega-lite`.

interface/src/app/node/high_range_exp.tpl.html
dongbohu: Good question. I have modified the code and now it is using the buttons. Take a look and let me know whether it looks better:
http://165.123.67.202/#/node/10998

interface/src/app/node/node.js
dongbohu: When defining a directive, we have two choices: a link function or a controller. I personally prefer using link function when possible. Since the directive here has its own template html file and isolated scope is used, any expression in the template html will always refer to the directive's own scope, **controllerAs**'s advantage of explicit scope is not so important (IMHO).

dongbohu:
@rzelayafavila: about the 2nd argument of $http.get(), I talked to @casey briefly. If we use this syntax, the line will be:
```
$http({method: 'GET', url: '/api/v0/sample/', params: {experiment: exp.accession}});
```
or:
```
var sampleURI = '/api/v0/sample/';
$http.get(sampleURI, {params: {experiment: exp.accession}});
```
Neither of them seems more readable than the current code. (The syntax in $resource is better). What do you think?

interface/src/app/node/high_range_exp.tpl.html
cgreene: I just took a look. I worry that there are too many "click here" type indications with the full button. For a user it can be confusing to have lots of branching paths with unclear priority.

cgreene: If you go the button route, a grey background might help over the blue.

cgreene: tagihqdqctlg
I made a quick sketchboard with my concerns about the buttonizing + thoughts.

cgreene: Don't think the help link styling needs to be here for this PR, but if the page exists we could link to the relevant anchor. I do suggest that we define a class for the links (maybe 'help'?) and we can visually style that later.

dongbohu: I changed the button's background to gray, and removed the number of experiments. I still like the two "+/-" icons more than the real buttons. The two buttons are taking larger space and kind of shift user's attention. I give "show all" button a color that is more obvious because I think it is important.

dongbohu: I can add an issue about the "help" link and class.

rzelayafavila: Hmmm, after seeing the buttons, I do agree that the icon does look better, and is less distracting. Also, is this page taking a little longer for everyone to load? When I load it in the browser, it gets stuck a little bit and asks me if I want to stop the script.

interface/src/app/node/node.js
rzelayafavila: Ok, that sounds good.

interface/src/app/node/high_range_exp.tpl.html
cgreene: @dongbohu sounds good on the help link. I agree with @rzelayafavila that the subtle nature of the icon is probably the way to go right now. User studies may lead us to change this in the future, but let's do this for now.

rzelayafavila:
@dongbohu: Regarding the syntax of $http.get(), I have a couple of questions there -
1) Should we write a $resource service for sample since that api endpoint might get requested by many parts in the code? and
2) I still think the second option you name looks more readable, because it explicitly names the parameters (which is helpful to somebody reading this code), instead of having them all together with a long URL. We don't need to define the sampleURI as a variable, right, we can just write ``$http.get('/api/v0/sample/', {params: {experiment: exp.accession}});`` ? What conclusion did you and @cgreene come to?

cgreene:
@rzelayafavila : 
Regarding 2 - I don't think we reached a specific conclusion. I think the syntax that you present in your comment here looks clean.

dongbohu:
Actually this sampleURI request part can be rewritten using $resource instead of $http. I used the latter because I wanted to limit the number of dependencies in this directive. And the reason why I stick with $http was because the `$http.get().then().then()` to get the activity and experiment data seems more natural than `$resource.get().$promise.then()`.

dongbohu:
The sampleURI data retrieval was not defined as a service because it is not used anywhere else (yet).

dongbohu:
I reproduced the Firefox performance issue on my desktop. It is bothering me.

dongbohu:
I defined a new variable **httpConfig** to hold the URI parameters, then add it as the second argument to all `$http.get()` function calls.

dongbohu:
A made a few more changes. Hopefully the code is more compact. Let me know what you think. Thanks.

cgreene:
All of my concerns have been addressed. The rest are outside my comfort zone. @rzelayafavila - is this :+1: ?

rzelayafavila:
Ok @dongbohu, I think the changes look good and I see you just filed an issue for the slow loading in Firefox. LGTM! :+1:

rzelayafavila:
Btw, I clicked on the 'approve' button for the changes just to make double sure, but the LGTM and/or thumbs up is also an indicator that the PR is approved.

Issue #47, opened 2016-12-12
Create a help page


cgreene:
It would be great to have a help page that explains specific elements used in the server. This would explain what `nodes`, `genes`, `experiments`, and `samples` are and how they all relate to each other. It could be linked to by a "help" box on specific elements.

Suggested via usability test with @georgiadoing .

rzelayafavila:
Help page has been created, and PR #75 is adding the content from the Google doc to it.

PR #50, opened 2016-12-20
Creating 'Help' page, which will contain documentation on what Genes, Samples, Experiments and Nodes are in adage, and how they relate to each other.

rzelayafavila:


cgreene:
Also, I think this could get merged first and then @tj8901nm and @georgiadoing could craft the help language to be most suitable through a second PR against the template file.

cgreene:
Also, we should cross-reference #47 here since that's the issue directly related to this.

tj8901nm:
Yes, I agree that this can be merged first and the help language can be added in another PR.

rzelayafavila:
@casey, @dongbohu, @tj8901nm : Ok, thanks! I only made one small change after it was approved, which was changing the id= attribute to the parent <div> instead of having it in the <span> tag (so we could get rid of those <span> tags). I will go ahead and merge.

Issue #1, opened 2016-09-01
"Download" WebUI is different after "bower install" Command


dongbohu:
After running **bower install** in interface/ to reinstall the packages specified in **bower.json**, I noticed that the dynamic rows of annotation type names retrieved from the back end look different from the three rows of static annotation types. It doesn't seem to affect the functionality, though.

Probably need to tweak **download.tpl.html** to make all rows on web UI look consistent.

mhuyck:
The static and dynamic parts *should* look a little different since the static ones cannot be changed, but I agree that the appearance could be refined a bit. Implementation suggestion: each row's HTML should be identical except for the use of a different CSS class to identify the static (or the non-static -- your choice!) rows.

dongbohu:
Thanks for your comments. I am going to rewrite this module when I get time. There are some other places that I want to improve too.

dongbohu:
Fixed.

PR #54, opened 2017-01-06
Updated "Download" Page with Unit Test

dongbohu:
I rewrote the **Download** page (to fix issue #1), and added some unit test cases written in Jasmine framework. It is pretty straightforward to write test cases in Jasmine. These test cases even exposed one minor issue in the download module. From now on, I will try to add test cases for all new JavaScript  modules that I created, and will add test cases to the other JavaScript modules that I wrote when I get time. 

In **download.tpl.html**, I am using tables instead of lists to show included and excluded types.
In **download.js**, I removed **$scope** from the controller and used **controllerAs** syntax instead.

The main references that I used when writing the unit tests are:
* **Angular Developer Guide: Unit Testing**
https://docs.angularjs.org/guide/unit-testing

* **AngularJS: Up & Running** (Shyam Seshadri & Brad Green, O'Reilly Media, 2014): 
Chapters 3, 7 and 12


dongbohu:
Some of the test cases in **download.spec.js** may seem trivial. I wrote them as a demo. If you have any questions, please feel free to let me know. I will try my best to explain.

By the way, the Two new issues reported by codeclimate seem trivial.

dongbohu:
This module doesn't have user-defined directive, I will write a test case for user-defined directive in another module and file it as a separate pull request.

dongbohu:
Here is the demo page:
http://165.123.67.202/#/download

mhuyck:
Codeclimate's "code similarity" algorithm has no appreciation for the beauty of symmetry and parallel structure. I like what you wrote; your code should stay as is and maybe one day someone will teach codeclimate about inverse operations. There would be no benefit from creating an abstract function to remove the "repetition" there... I'd argue it would actually be a harmful change that adds complexity.

dongbohu:
Thanks @mhuyck. I thought of merging the duplicate lines into one function after seeing this complaint. But  as you said, it would make the code less readable.

interface/src/app/download/download.tpl.html
```diff
@@ -29,52 +26,59 @@
...
-            <div as-sortable-item-handle>
-              <i class="fa fa-trash fa-lg" aria-hidden="true"
-                 ng-click="del_type($index)"></i>
-              {{$index + 4}}: <b>{{item}}</b>
-            </div>
-          </li>
-        </ul>
+      <div class="col-xs-4">
+        <!-- Three fixed types -->
+        <table class="table table-bordered" style="margin-bottom: 0">
```
mhuyck: I'm curious: why did you switch from a `<ul>` to a `<table>` here? The contents are not tabular, so shouldn't this be coded as a list?

interface/src/app/download/download.tpl.html
```diff
@@ -29,52 +26,59 @@
...
+              <td class="danger" as-sortable-item-handle>
+                <i class="fa fa-undo fa-lg" aria-hidden="true"
+                   ng-click="ctrl.addType($index)"></i>
+                {{$index + 1}}: {{item}}
+              </td>
+            </tr>
+          </tbody>
+        </table>
       </div>
+      <div class="col-xs-4"></div>
```
mhuyck: This is a nice touch for large window sizes, but be careful of what will happen on smaller screens. There are better ways to get the columns to reflow naturally... have a look at http://getbootstrap.com/css/#grid for the overview of the Bootstrap layout system. I'm thinking maybe the "mobile and desktop" example (http://getbootstrap.com/css/#grid-example-mixed ) might have what you're looking for here.

interface/src/app/download/download.js
```diff
@@ -1,80 +1,74 @@
...
-        $scope.annotations.query_message = '';
-      },
-      function(responseObject, responseHeaders) {
-        $log.error('Query errored with: ' + responseObject);
-        $scope.annotations.query_message = 'Connection to server failed';
-      }
-    );
+.controller('DownloadCtrl', ['$window', '$log', 'AnnotationTypes',
+  function DownloadController($window, $log, AnnotationTypes) {
+    var self = this;
```
dongbohu: Yes, I learned this approach from the book **AngularJS: Up and Running**. One advantage of this approach is that if the controller code is modified later, I never have to worry about whether **this** refers to the controller instance or not. Since in JavaScript, **self** actually refers to the same object as **this** does, the cost is minimal.

interface/src/app/download/download.js
```diff
@@ -1,80 +1,74 @@
...
+      function success(response) {
+        response.objects.forEach(function(element) {
+          self.includedTypes.push(element.typename);
+        });
+        self.queryStatus = '';
+      },
+      function error(errResponse) {
+        var errMessage = errResponse.statusCode + ' ' + errResponse.statusText;
+        $log.error('Failed to get annotation types: ' + errMessage);
+        self.queryStatus = 'Connection to server failed: ' + errMessage;
```
dongbohu: Yes, when this kind of error happens, the only thing an user can do is probably refresh the page. I added more details for two reasons:
1) Many users may be familiar with errors such as "404: page not found". 
2) In case our users need technical help from us, they can say that "I got an error 405 blah blah"; otherwise we will have to ask them to check the console and tell us the details.

I can add "Please refresh the page." at the end of `self.queryStatus` to make it a little more user-friendly.

interface/src/app/download/download.tpl.html
dongbohu: I tweaked **<li>** tags a few times with different attributes, but the styles of the three fixed annotation types and the included sortable types still seem slightly different. That is why I switched to <table>. Besides, I really like the **table-hover** class in Bootstrap CSS.

interface/src/app/download/download.tpl.html
dongbohu: Thanks for pointing it out. I will tune it.

dongbohu:
I committed the two trivial changes that I made on app/node/. One is renaming the HTML template to a name that is more uniform to the directive, the other is revision of the code that refers to this template.

interface/src/app/download/download.spec.js
```diff
@@ -2,21 +2,113 @@
...
+// Test the generic controller properties (without server calls):
+describe('DownloadCtrl w/o server calls', function() {
+  beforeEach(module('adage.download'));
+
+  var ctrl, mockService;
+  // Create a mocked "AnnotationTypes" service, which overwrites the service
+  // in download.js (because it is loaded after "adage.download" module).
+  beforeEach(module(function($provide) {
+    mockService = {
+      get: function() { }
```
mhuyck: <strike>Shouldn't the `mockService` be returning mocked-up data? Even if it's just a few lines, I think the purpose is to simulate the function of the service.</strike> Nevermind... I now see your comment below.

mhuyck: This is actually really easy to do, by the way: just make a manual call to the service in your browser and copy and paste a sample of the output here. You'll need the `format=json` parameter for TastyPie and you'll probably need to wrap the output to fit here, but that's about it.

interface/src/app/download/download.spec.js
```diff
@@ -2,21 +2,113 @@
...
+    };
+    $provide.value('AnnotationTypes', mockService);
+  }));
+
+  beforeEach(inject(function($controller) {
+    ctrl = $controller('DownloadCtrl');
+  }));
+
+  it('should be defined', function() {
+    expect(ctrl).toBeDefined();
```
mhuyck: Do we also want `expect(ctrl).toBeTruthy();` somewhere? That would replace the dummy test that was removed above. Although it doesn't do much, it does confirm that the code for that module compiles. It is nice to have a separate test that can quickly pinpoint a compile failure.

Unless maybe I'm missing something, I don't think the `.toBeDefined()` test here will cover that.

interface/src/app/download/download.spec.js
```diff
@@ -2,21 +2,113 @@
...
+  });
+
+  it('should have three fixed annotation types', function() {
+    expect(ctrl.fixedTypes).toEqual(
+      ['experiment', 'sample_name', 'ml_data_source']
+    );
+  });
+
+  // Because get() method in mocked AnnotationTypes service does nothing,
+  // all controller properties should have their initial values.
```
mhuyck: Aha... now I see why you did that above. And now I notice the other mock test below, so we don't actually want the mock to return data. Got it.

interface/src/app/download/download.spec.js
```diff
@@ -2,21 +2,113 @@
...
+    $httpBackend.expectGET(uri).respond(
+      {objects: [
+        {typename: 'mock_type1'},
+        {typename: 'mock_type2'},
+        {typename: 'mock_type3'}
+      ]}
+    );
+    // Instantiate the controller:
+    ctrl = $controller('DownloadCtrl');
+  }));
```
mhuyck: Nice! I am psyched that you figured out how to get `$httpBackend` working.

This might be good sample data as is, but see my comment in the other mock above for a tip on how to extract a sample of real data to include.

interface/src/app/download/download.spec.js
```diff
@@ -2,21 +2,113 @@
...
+    // Exclude 'mock_type1' too:
+    ctrl.delType(0);
+    expect(ctrl.includedTypes).toEqual(['mock_type3']);
+    expect(ctrl.excludedTypes).toEqual(['mock_type2', 'mock_type1']);
+
+    // Add 'mock_type1' back to includedTypes:
+    ctrl.addType(1);
+    expect(ctrl.includedTypes).toEqual(['mock_type3', 'mock_type1']);
+    expect(ctrl.excludedTypes).toEqual(['mock_type2']);
+  });
```
mhuyck: Awesome. Very nice tests!

interface/src/app/download/download.spec.js
```diff
@@ -2,21 +2,113 @@
...
+
+  // Ensure no outstanding expectation or request at the end.
+  afterEach(function() {
+    $httpBackend.verifyNoOutstandingExpectation();
+    $httpBackend.verifyNoOutstandingRequest();
```
mhuyck: So this checks that the mock server's data have been fully read and that there are no unfulfilled mock http requests, is that right?

interface/src/app/download/download.js
mhuyck: Part of my concern with this is that self is actually *not* a part of JavaScript. It can be found in the browser context because the global context is the `window` object, and there is a `window.self`, which points back to `window`. Creating a `var self =...` will mask that global. I think that masking is fine, actually, but it just doesn't feel like a best practice solution to me.

It works, and I know people do it that way (including the authors of that book you cite) but I think what bothers me about it is that it doesn't really clarify what is going on.

Anyway, this is really a style issue and not something we should resolve in a PR. Moving on...

interface/src/app/download/download.js
mhuyck: Fair enough. I think what I'm aiming for is a simple message to the user describing what they should try and possibly a link to click for the technical details. This type of *progressive disclosure* is helpful to make the UI friendly to a novice, but still useful to an expert.

I think your suggestion is a good solution for this example. Perhaps we should work up a standardized error reporting widget that we can reuse everywhere...

interface/src/app/download/download.tpl.html
```diff
@@ -52,7 +52,7 @@
               <td class="success" as-sortable-item-handle>
                 <i class="fa fa-trash fa-lg" aria-hidden="true"
                    ng-click="ctrl.delType($index)"></i>
-                {{$index + 4}}: {{item}}
+                {{$index + 1 + ctrl.fixedTypes.length}}: {{item}}
```
mhuyck: nice

mhuyck:
I see codeclimate has added another "duplicated code" issue, but it's all wet on that one too. Safe to ignore.

interface/src/app/download/download.js
dongbohu: I didn't know that **window.self** already exists.
From what I read, **self** seems to be a variable name used widely in JavaScript community. Google's AngularJS style guide recommends **controllerAs** syntax, but it doesn't specify whether it is a good idea to assign `this` to another variable:
https://google.github.io/styleguide/angularjs-google-style.html#dollarsign
(The example is too simple to assign `this` to another variable.)

Another style guide that seems to be popular actually recommend using the variable name **vm** (which stands for "view model") to refer to the controller instance:
https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#controlleras-with-vm 
As you said, this is totally a style issue. Once we reach agreement, whatever name is okay with me.

interface/src/app/download/download.js
dongbohu: I will modify `self.queryStatus` then. I agree that in the long run, it will be nice to have a standard error-reporting widget.

interface/src/app/download/download.spec.js
dongbohu: Good to know. Thanks.

interface/src/app/download/download.spec.js
dongbohu: `expect(ctrl).toBeDefined()` is a more specified test than `expect(ctrl).toBeTruthy()` is. If the former fails, the latter will fail for sure, right?

I probably should merge all `it()` blocks in the first `describe()` into one `it()` test case.

interface/src/app/download/download.spec.js
dongbohu: It took me a while to figure out how to test `$resource().get()` method. The "AngularJS: Up and Running" book only covered testing of `$http`. Eventually I realized that $resource is just a wrapper of $http, so testing of $resource is very similar.

interface/src/app/download/download.js
mhuyck: This is one of those things where there isn't yet any consensus, as far as I can tell. It works properly either way, so I'd say we should only change our code if it's confusing.

interface/src/app/download/download.spec.js
dongbohu: Because we are using `$httpbackend.expectGET(...).respond(...)` earlier, `$httpBackend.verifyNoOutstandingExpectation()` will fail if `$httpBackend.flush()` is not called later.  On the other hand, if we don't have `$httpbackend.expectGET(...)` earlier, but the tested code has GET request (through XMLHttpRequest, $http or $resource, etc), then `$httpBackend.verifyNoOutstandingRequest()` will fail.

$httpBackEnd also offers other methods (such as `when` or `whenGET`) that don't have such restriction. For more details, you can read Angular documentation on $httpBackend:
https://docs.angularjs.org/api/ngMock/service/$httpBackend

interface/src/app/download/download.spec.js
mhuyck: `var.toBeDefined()` is the same as `var` == `undefined`. For `var.toBeTruthy()` to pass, `var` would have to be not only defined, but evaluate == `True`.

If there is a syntax error in the module, I don't know if `expect(ctrl).toBeDefined()` would pass or not, but I have seen `expect(ctrl).toBeTruthy()` fail when the module had a syntax error. I think you would need to try it out and see.

interface/src/app/download/download.js
rzelayafavila: @mhuyck: I also like this approach of using ``self`` instead of ``this`` when using controllerAs syntax, since you can be sure that you are always calling the controller instance when calling ``self`` (as @dongbohu says). Otherwise, you can fall into the JavaScript trap (which I have!) of having ``this`` refer to different things, depending where it is being called from. I think that using ``var self = this;``, together with ``controllerAs`` syntax, and adding new variables/objects to the controller instance is the best practice.

interface/src/app/download/download.js
rzelayafavila: I totally agree with having more details so users can either recognize them or report those details to us when they happen, as well as having the message suggesting a page refresh.

interface/src/app/download/download.spec.js
dongbohu: According to **AngularJS: Up & Running**, page 43:
```
toBeTruthy:
Checks the value passed to the matcher to pass the JavaScript concept of true.
Nonnull objects, nonempty strings, nonzero numbers, and the Boolean true all
evaluate to be truthy in JavaScript.
```
So a non-Null object should be always truthy. Let me make some test too.

interface/src/app/download/download.spec.js
mhuyck: Excellent. Thank you for the documentation link, too!

interface/src/app/download/download.js
rzelayafavila: Hmmm, I see that GitHub had not refreshed the PR when I posted and thus I missed some of the later comments. I agree that whatever name is ok then and only change the code if it's confusing.

dongbohu: Yes, for some reason github seems slow today.

mhuyck: I have no ambivalence about the `controllerAs` syntax. I am 100% agreed that we should use that everywhere. Will try to remember to convert mine as I touch older code that manipulates `$scope` directly.

dongbohu:
@mhuyck: I don't know why github doesn't show your comment on the "drag to drop" glitch when the excluded types table is empty. (I did see it in the email.) Thank you very much for pointing it out. It is due to the fact that when there is no excluded type, the "draggable" area in excluded table is not generated yet. One way to fix it is to hide the excluded table when it is empty. Another way is to add a dummy row in excluded table. What do you think?

mhuyck:
I logged it as an issue rather than noting it in this PR since it isn't related to the code changes you've made here. It should get fixed, to be sure, but I think it doesn't need to be done with this set of changes.

Since you asked: I think it would be confusing to hide and show the excluded list. Probably the dummy row approach would work best from a usability perspective. You'd need logic to remove it when the first item is added and replace it if the last item is removed.

Another approach would be to simplify the page so that it's only one list. The items should still be sortable, but you could use a checkbox or other selection method to indicate what columns should appear in the download file.

interface/src/app/download/download.spec.js
dongbohu: After putting more thoughts on this, I believe that here `expect(ctrl).tpBeDefined()` and  `expect(ctrl).tpBeTruthy()` are almost equivalent. When some error prevents the controller instance from being instantiated, it wouldn't be defined, and it would not be truthy (but `truthy` is more general than `true`) either, so for an object this is equivalent. For a string like this:
```var foobar = '';```
then `foobar` is defined but it is not truthy.

**Comment by dongbohu** [comment]
@mhuyck: It should be easy to have the issue fixed. Let me fix it in this PR.

File: interface/src/app/download/download.spec.js (hunk 6 above)
Comment (mhuyck) [inline interface/src/app/download/download.spec.js]: Thinking it through is always good, but did you actually test it? It is possible that compilation failure results in `ctrl` being defined as a null object (`{}`) which would be "defined" but evaluate to "falsy."

Comment (mhuyck) [inline interface/src/app/download/download.spec.js]: Testing should be as simple as introducing a compilation error (remove a `;` somewhere) and seeing what happens.

File: interface/src/app/download/download.js [hunk 12]
```diff
@@ -1,80 +1,74 @@
...
 
-// "AnnotationTypes" is a service that retrieves a list of available
-// annotation columns, from which the user can select for download.
 .factory('AnnotationTypes', ['$resource', function($resource) {
-  return $resource(
-    '/api/v0/annotationtype/',
-    {q: '', limit: 0},
-    {'query': {method: 'GET', isArray: false}}
-  );
+  return $resource('/api/v0/annotationtype');
```
rzelayafavila: Nice! Much cleaner!

interface/src/app/download/download.spec.js
dongbohu: Since `;` is optional in JavaScript, I triggered this error by adding this line at the end of the controller definition:
```
return { };
```
Then the resulting controller is defined but not truthy.

Considering that this test case is more like a demo (this error will be detected by all the other test cases in the same describe block too), I am going to change it into:
```
  it('should be defined and truthy', function() {
    expect(ctrl).toBeDefined();
    expect(ctrl).toBeTruthy();
  });
```

interface/src/app/download/download.js
```diff
@@ -1,80 +1,74 @@
...
- * annotation typenames.
+ * "adage.download" module, which will download sample annotations based on
+ * selected annotation types.
  */
 
 angular.module('adage.download', [
   'ui.router',
-  'placeholders',
   'ui.bootstrap',
+  'ngResource',
```
rzelayafavila: This is just a side comment - I agree that we should include these dependencies on each module that requires them. I was thinking about it, and even though ``ngResource`` is included in the 'adage' module itself in ``app.js``, and could technically be omitted here, I agree that we should explicitly include it here since every one of these modules should be a stand-alone packet of code.

dongbohu: Unit tests will fail without it!

rzelayafavila: Makes sense, as they test it as a unit on it's own :+1:

dongbohu:
I fixed the issue that @mhuyck pointed out and addressed a few other problems. Please see the latest committed code.

rzelayafavila:
Ok @dongbohu - I have made it through all of the code and now that the dragging across the Annotation field boxes is fixed, almost all of my concerns have been addressed. There is this one error that is popping up in the JavaScript console when I load the page in Firefox, saying ``Error: f.scope.modelValue is null``. Do you get that at all when you load it in your browser?

dongbohu:
@rzelayafavila: I figured out where the error comes from. I am working on it and will update the PR when it is done.

dongbohu:
I fixed the bug that Rene pointed out. Please see the updated PR.

rzelayafavila:
Ok great! I am not getting the error anymore. This Pull Request LGTM :+1:

mhuyck:
Looks like you've addressed everything I asked about. Thanks again for leading the way on UI testing, @dongbohu!

dongbohu:
Thanks to both of you!

PR #55, opened 2017-01-11
Geneset enrichment table

rzelayafavila:
This is a Work In Progress Pull Request - it is not ready to be code reviewed yet, but I wanted to make the code I have been working on available to everyone.

- Adding functions for hypergeometric test in app/utils.js file and adding some corresponding unit tests.


rzelayafavila:
Ok, this Pull Request is finally ready to be reviewed! The geneset enrichment table in the Node page is up and running, and you can try it out here: http://165.123.67.154:8000/#/node/1

interface/src/app/node/node.js
```diff
@@ -299,4 +307,123 @@ angular.module('adage.node', [
...
-}]);
+}])
+
+.directive('enrichedGenesets', ['MathFuncts', 'PickledGenesetsService', '$log',
+  function(MathFuncts, PickledGenesetsService, $log) {
+    return {
+      templateUrl: 'node/enriched_genesets.tpl.html',
+      restrict: 'E',
+      scope: {
+        organism: '=',
```
dongbohu: Since you are not changing **organism** inside the directive, you can gain some efficiency by using **@**.

interface/src/app/node/node.js
```diff
@@ -299,4 +307,123 @@ angular.module('adage.node', [
...
+        hWGenes: '=' // hWGenes is an array of high-weight Genes
```
dongbohu: "@" is recommended for the same reason.

interface/src/app/node/node.js
```diff
@@ -299,4 +307,123 @@ angular.module('adage.node', [
...
+          }
+
+          relevantGenesetArray.sort(function(a, b) {
+            return a.pValue - b.pValue;
+          });
+
+          return relevantGenesetArray;
+        };
+
+        $scope.$watch('hWGenes', function() {
```
dongbohu: Since you are watching **hWGenes** by reference here, I have a feeling that it is not doing what you expected it to do. Will the interface still okay if you just write:
```
if ($scope.hWGenes.length > 0) { 
  // ... 
}
```
without the outside `$watch()` at all?

interface/src/app/node/node.tpl.html
```diff
@@ -7,13 +7,21 @@
   <div class="container">
     <div class="row">
       <div class="col-xs-6">
-        <high-weight-genes node-id="{{ctrl.id}}"></high-weight-genes>
+        <high-weight-genes node-id="{{ctrl.id}}" h-w-genes="ctrl.hWGenes"></high-weight-genes>
```
dongbohu: Since the directive is called "high-weight-genes* already, does it make sense to simply write:
```
<high-weight-genes node-id="{{ctrl.id}}" genes="ctrl.hWGenes">
```
Inside the controller, to make the variable name more descriptive, we can change it from "hWGenes" to "highWeightGenes", or something like this.

interface/src/app/node/node.tpl.html
```diff
@@ -7,13 +7,21 @@
...
       </div>
 
       <div class="col-xs-6">
         <high-range-exp node-id="{{ctrl.id}}" top-exp="20"></high-range-exp>
       </div>
     </div>
+    <div class="row">
+      <div class="col-xs-6">
+        <enriched-genesets organism="ctrl.organism" h-w-genes="ctrl.hWGenes"></enriched-genesets>
```
dongbohu: To make the tag more descriptive, maybe we should change the tag into:
```
<enriched-genesets organism="{{ctrl.organism}}" high-weight-genes="{{ctrl.highWeightGenes}}">
```
(suppose you will use "@" instead of "=" in the scope.

interface/src/app/tribe_client/resource.js
```diff
@@ -32,4 +32,14 @@ angular.module('adage.tribe_client.resource', ['ngResource'])
   );
 }])
 
+.factory('PickledGenesetsService', ['$resource', function($resource) {
+  return $resource(
+    '/tribe_client/return_unpickled_genesets',
+
+    {organism: '@organism'},
+
+    {}
```
dongbohu: I think this parameter is optional, right?

interface/src/app/utils.js
```diff
@@ -0,0 +1,41 @@
...
+
+.factory('MathFuncts', [function() {
+  return {
+    /*
+     * The binomLog and hyperGeometricTest functions are based heavily on the
+     * 'binomlog' and 'hyperg' functions in /integrator/static/js/imp.js
+     * See code here:
+     * https://bitbucket.org/cgreene/integrator/src/1194f7ae1b11057a4649114e1f55570542acbcde/integrator/static/js/imp.js?at=giant&fileviewer=file-view-default
+    */
+    binomLog: function(n, k) {
```
dongbohu: Can you write some comments to briefly explain what **n** and **k** are?

interface/src/app/utils.js
```diff
@@ -0,0 +1,41 @@
...
+      for (var i = n - k + 1; i <= n; i++) {
+        coeff += Math.log(i);
+      }
+      for (i = 1; i <= k; i++) {
+        coeff -= Math.log(i);
+      }
+      return coeff;
+    },
+
+    hyperGeometricTest: function(k, m, n, N) {
```
dongbohu: Not sure whether we should use some more descriptive variable names in this function and the one above.

interface/src/app/utils.spec.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * Unit tests.
+ */
+
+// Test the hyperGeometricTest function
+describe('Test for hypergeometric test function in utils.js', function() {
+  beforeEach(module('adage.utils'));
+
+  var mockService;
```
dongbohu: In this test, the name **mockService** is a little misleading, because it is assigned to the real "MathFuncts" service on line 13 (instead of being a MOCKED service). It may be better to simply call it **MathFuncts**.

interface/src/app/utils.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+  beforeEach(function() {
+    angular.mock.inject(function($injector) {
```
dongbohu: Just so you know, this line can also be written as:
```
  inject(function($injector) {
    // ...
  });
```
because angular.mock.inject() is introduced to the global scope (window) too;

Or the whole **beforeEach** can be more compact like this:
```
beforeEach(inject(function(MathFuncts) {
  mockService = MathFuncts;
}));
```
(Just my two cents on unit tests. After spending some serious time reading Angular doc, I found Angular's injector and injectable types tricky to understand.)

interface/src/app/utils.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  // hyperGeometricTest function matches the output produced by
+  // http://stattrek.com/m/online-calculator/hypergeometric.aspx and
+  // https://www.geneprof.org/GeneProf/tools/hypergeometric.jsp for
+  // some given k values.
+  it('should output a desired p-value for k=1', function() {
+    expect(+(1 - mockService.hyperGeometricTest(1, 5, 10, 50)).toFixed(10))
+    .toEqual(0.689437218);
+  });
+
+  it('should output a desired p-value for k=2', function() {
```
dongbohu: Do you think it makes sense to merge at least two **it()** calls into one to confirm that multiple calls of this service do not affect one another?

interface/src/app/node/node.js
```diff
@@ -73,12 +78,15 @@ angular.module('adage.node', [
...
+              entrezID = response.objects[i].gene.entrezid;
               if (desc.toLowerCase() === 'hypothetical protein') {
                 ++numHypo;
               }
               $scope.genes.push(
-                {sysName: sysName, stdName: stdName, desc: desc});
+                {sysName: sysName, stdName: stdName, desc: desc,
+                 entrezID: entrezID});
             }
+            $scope.hWGenes = $scope.genes;
```
dongbohu: Since "hWGenes" and "genes" are equal in the scope, do you think it makes sense to merge these variables into one?

interface/src/app/node/enriched_genesets.tpl.html
```diff
@@ -0,0 +1,21 @@
+<h4>
+  <b>Enriched gene sets: </b>
+</h4>
+
+<table id="enrichment-table" class="table table-striped table-bordered table-condensed">
+  <thead>
+    <tr>
+      <td>Pathways/Processes/Diseases</td>
```
rzelayafavila: Ah I see, those headers probably would look better in bold, huh? I will change those ``td``s to ``th``s.

interface/src/app/node/node.js
rzelayafavila: Ah, true - I will change this.

interface/src/app/node/node.js
rzelayafavila: Hmm - I hadn't thought of that, but you're right - since there is already a watcher on $scope hWGenes, I might not need to put it inside another $watch() statement. I will test this out.

interface/src/app/node/enriched_genesets.tpl.html
dongbohu: Yes, **th** is recommended inside **thead**. You will see the difference when you make the change.

interface/src/app/utils.spec.js
rzelayafavila: Ah gotcha - totally, I have found them pretty tricky to understand too. I will change this, thanks!

interface/src/app/utils.js
dongbohu: I wonder whether it is a good idea to put utils.js and utils.spec.js into a sub-directory such as **app/utils/**. I expect that we will write more and more utility code. If we dump everything into this util.js, it will become very big eventually, and we have to break it up eventually.

interface/src/app/node/node.js
```diff
@@ -4,7 +4,9 @@
 
 angular.module('adage.node', [
   'ui.router',
-  'ui.bootstrap'
+  'ui.bootstrap',
+
```
rzelayafavila: Yes, exactly - do you think that is a good way to go about it?

dongbohu: I like the way that angular's modules and ours are in different groups. But I know some coding guides (not Google's) don't like a blank line inside an array definition. I guess you can replace the blank line with a comment line if you want. It is totally up to you.

rzelayafavila: Mmm, interesting. I do like to keep them a bit separate, so I will put in a comment here (since I do see why it would be strange to have just a blank line inside an array definition).

interface/src/app/node/node.js
rzelayafavila: Hmm, Im pretty sure that ``'@'`` wouldn't work here because it would expect hWGenes to be a string, but we are expecting hWGenes to always be a JSON object.

rzelayafavila: It definitely would work for organism, though, since that is just the organism species name string.

interface/src/app/node/node.js
mhuyck: Nice catch. My gut says we should generally have a high bar for writing our own `$watch` functions manually. Whenever possible, let's have Angular features do the watching for us.

interface/src/app/node/node.js
```diff
@@ -300,4 +309,123 @@ angular.module('adage.node', [
...
+              for (var j = 0; j < genesetList.length; j++) {
+                var genesetID = genesetList[j];
+                if (!genesetGenes[genesetID]) {
+                  genesetGenes[genesetID] = [];
+                }
+                genesetGenes[genesetID].push($scope.hWGenes[i]);
+              }
+            } else {
+              $log.warn('Entrez ID: ' + geneEntrezID + ' not found in ' +
+                        'genesets.');
```
mhuyck: What are the consequences for our enrichment test of not finding a geneset here? I'm wondering if this is expected, so remaining silent about it to the user is okay or if it's an actual error condition and the user should be notified, or somewhere in between.

interface/src/app/node/node.js
```diff
@@ -300,4 +309,123 @@ angular.module('adage.node', [
...
+          // This will be the number of genes from the high weight gene list
+          // that are also present in any of the genesets that were returned
+          var m = 0;
+
+          // Fill out the genesetGenes object
+          for (var i = 0; i < $scope.hWGenes.length; i++) {
+            var genesetList = null;
+            var geneEntrezID = $scope.hWGenes[i].entrezID;
+
+            if (geneGenesets.hasOwnProperty(geneEntrezID)) {
```
mhuyck: Took me a minute to figure out that `geneGenesets` was not a typo of `genesetGenes`. I now see the symmetry... think I'm okay with it, but if I get confused again might recommend changing one to be a more distinct name.

mhuyck: Okay, that was not too confusing. I kind of like it, actually.

interface/src/app/node/node.js
```diff
@@ -5,7 +5,10 @@
 angular.module('adage.node', [
   'ui.router',
   'ui.bootstrap',
-  'ngResource'
+  'ngResource',
+  // Dependency modules from the 'adage' Angular app
+  'adage.utils'
```
cgreene: Not sure if `adage.utils` or a specific stats package would make more sense. Worried that we'll use these again and again. At least the hypergeometric probably will be [if that's in `adage.utils` now].

mhuyck: Huh. I thought there was other stuff in there, but all I see is the `MathFuncts` factory that @rzelayafavila just added. I must have been thinking of something else. I would advocate for changing the name from `adage.utils` to `adage.stats` or even `greenelab.stats` if we think this will be used outside of the `adage-server`.

interface/src/app/node/node.js
rzelayafavila: Mmm both of those points make sense - I will change all of these to simply "genes", which will get rid of some unnecessary code and make it clearer.

interface/src/app/node/node.tpl.html
rzelayafavila: I decided to just go with "genes", and share that in the scopes across all the directives that use them in this Node page.

interface/src/app/node/node.tpl.html
rzelayafavila: Right, I agree, changing all of these to simply "genes".

cgreene:
Random thing I noticed from the email titles - you can probably edit the PR to remove the work in progress tag.

interface/src/app/utils.spec.js
rzelayafavila: Ahh, I see, so it is not actually a real mocked service. Ok, I will change it to that, thanks!

interface/src/app/node/enriched_genesets.tpl.html
```diff
@@ -0,0 +1,21 @@
...
+      <td>P-Value</td>
+      <td>Genes</td>
+    </tr>
+  </thead>
+
+  <tr ng-repeat="geneset in enrichedGenesets">
+    <td><a href="{{geneset.url}}">{{geneset.name}}</a></td>
+    <td>{{geneset.dbase}}</td>
+    <td>{{geneset.pValue}}</td>
+    <td><span ng-repeat="gene in geneset.genes"> {{gene.stdName}} </span></td>
```
rzelayafavila: @mhuyck and @dongbohu: That makes sense - I am putting in Matt's suggested implementation, and that way, if we decide to use comma and space characters to separate the genes in the future, we can just add it to the function. Thanks!

adage/adage/config.py.template
```diff
@@ -131,7 +131,9 @@ AWS_CONFIG = {
         '<configure:http://example.com/tribe_client/get_token>',
     'tribe_scope': 'write',
     'tribe_login_redirect': '/#/home',
-    'tribe_logout_redirect': '/#/home'
+    'tribe_logout_redirect': '/#/home',
+    'public_geneset_folder': os.path.join(OS_CONFIG['home_dir'],
```
rzelayafavila: Ah yes - so this is actually to be able to access the genesets that are pickled from Tribe for this geneset enrichment analysis, and it is called by the tribe-client package. I totally agree this is unclear, so I will add some comments stating that purpose.

interface/src/app/utils.spec.js
rzelayafavila: Yep, good idea. In fact, I will just go ahead and put all of these 5 into one ``it()`` call to make the code clean and concise, while still making sure that multiple calls of the service do not affect one another.

interface/src/app/node/node.js
rzelayafavila: Ok, I actually really like the idea of having a 'greenelab.stats' module (inside a greenelab folder), so I switched it to that. That way, we can just make everything in the 'greenelab' folder be portable Angular modules that we can use in different servers.

interface/src/app/tribe_client/resource.js
rzelayafavila: Yep, I removed it.

interface/src/app/utils.js
rzelayafavila: Ok, I added some more comments for these!

rzelayafavila: True - I ended up putting them in a new module called greenelab.stats, in app/greenelab/stats.js

interface/src/app/utils.js
rzelayafavila: I just added some more comments/documentation about how these variable names are common in online descriptions of the hypergeometric test, and also they are the variable names used by Aaron in the corresponding functions in imp.js. Do you think those look good?

rzelayafavila:
Ok reviewers, I made a few changes, and I think I have addressed all the comments. Let me know how it looks! Thanks!

adage/adage/config.py.template
mhuyck: Aha. Cool, thanks!

So do we need to require a newer version of `tribe-client` now to avoid deployment problems? If so, please update `requirements.txt` to reflect the new minimum version.

interface/src/app/greenelab/stats.js
```diff
@@ -0,0 +1,50 @@
...
+  return {
+    /*
+     * The binomLog and hyperGeometricTest functions are based heavily on the
+     * 'binomlog' and 'hyperg' functions in /integrator/static/js/imp.js
+     * See code here:
+     * https://bitbucket.org/cgreene/integrator/src/1194f7ae1b11057a4649114e1f55570542acbcde/integrator/static/js/imp.js?at=giant&fileviewer=file-view-default
+    */
+    binomLog: function(n, k) {
+      // The 'n' and 'k' arguments are the same as the 'n' and 'k' arguments
+      // arguments for the hyperGeometricTest function below (the number of
```
mhuyck: typo in the comment here: "arguments arguments"

interface/src/app/greenelab/stats.spec.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+// Test the hyperGeometricTest function
+describe('Test the hypergeometric test function in stats.js', function() {
+  beforeEach(module('greenelab.stats'));
+
+  var MathFuncts;
+
+  beforeEach(function() {
+    inject(function($injector) {
```
dongbohu: Although this `beforeEach()` block works, it adds some unnecessary complexity. It can be simplified to either:
```
beforeEach(inject(function($injector) {
  MathFuncts = $injector.get('MathFuncts');
});
```
or:
```
beforeEach(inject(function(_MathFuncts_) {
  MathFuncts = _MathFuncts_;
});
```

dongbohu: By the way, the second version of beforeEach() is prefered here, because you only need to resolve `MathFuncts`. If you have more dependencies (such as >= 3), using $injector will be better, because you only need to put $injector as the function parameter (instead of putting three dependencies as parameters).

interface/src/app/node/node.js
```diff
@@ -300,4 +306,125 @@ angular.module('adage.node', [
...
+          }
+
+          relevantGenesetArray.sort(function(a, b) {
+            return a.pValue - b.pValue;
+          });
+
+          return relevantGenesetArray;
+        };
+
+        $scope.$watch('genes', function() {
```
dongbohu: So the code doesn't work if you remove this watcher? I thought it was unnecessary.

rzelayafavila: Nope, unfortunately it does not work without the watcher. It turns out, without the watcher, that code only gets executed once, even though Angular is still watching $scope.genes.

adage/adage/config.py.template
rzelayafavila: Ah yes, good call @mhuyck! :+1:

interface/src/app/greenelab/stats.js
```diff
@@ -0,0 +1,50 @@
...
+      for (var i = 0; i < k; i++) {
+        var logSum = this.binomLog(m, i) +
+            this.binomLog(N - m, n - i) - this.binomLog(N, n);
+        sum += Math.exp(logSum);
+      }
+      return sum;
+    }
+
+  };
+}])
```
mhuyck: I haven't dug into the math itself to verify, but this is very nice and clear coding. Thank you for the excellent comments and organization!

interface/src/app/greenelab/stats.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(+(1 - MathFuncts.hyperGeometricTest(3, 5, 10, 50)).toFixed(10))
+      .toEqual(0.0482603032);
+
+      expect(+(1 - MathFuncts.hyperGeometricTest(4, 5, 10, 50)).toFixed(10))
+      .toEqual(0.0040835205);
+
+      expect(+(1 - MathFuncts.hyperGeometricTest(5, 5, 10, 50)).toFixed(10))
+      .toEqual(0.0001189375);
+    }
```
mhuyck: Great tests! Guess I don't need to check the math now since this proves it works...

interface/src/app/node/node.js
```diff
@@ -300,4 +306,125 @@ angular.module('adage.node', [
...
+                1 - MathFuncts.hyperGeometricTest(k, m, n, N);
+            pValue = pValue.toPrecision(pValueSigDigits);
+
+            if (pValue < $scope.pValueCutoff) {
+              relevantGenesetArray.push({
+                'name': genesetInfoObj.name, 'dbase': genesetInfoObj.dbase,
+                'url': genesetInfoObj.url, 'pValue': pValue,
+                'genes': genesetGenes[gsID].map(function(gene) {
+                  return gene.stdName;
+                }).join(' ')
```
mhuyck: Aha. This isn't exactly what I was saying but if you don't need to have the individual genes accessible, then this works too. Good use of `.map()`!

interface/src/app/node/node.js
mhuyck: Hmm. A puzzle, but one for another day, I think.

rzelayafavila:
@mhuyck: Ok, thank you! Which was the niggly typo you were referring to? I don't think I found it in your latest comments.

mhuyck:
Looks like you got that typo... but I didn't see your correction. It was the one in `stats.js`. 👍

rzelayafavila:
Ok, thanks @mhuyck @dongbohu @cgreene for your help with this PR! I will go ahead and merge now.

PR #57, opened 2017-01-13
Unit test demo of Angular directive

dongbohu:
This PR includes a new spec file to test the directive `<high-weight-genes>`. Angular allows us to test not only the variables set by the scope function (or controller) in user-defined directives, but also the data in resulting HTML that is compiled from the directive. So strictly speaking, you may argue that the test is both unit test and integration test.

To avoid the duplicate lines, I didn't demonstrate all I learned recently into this file. (I may create a shared google doc to show them when I get time.)

### DIFF
--- a/interface/src/app/node/high_weight_genes.spec.js
+++ b/interface/src/app/node/high_weight_genes.spec.js
@@ -1,0 +1,86 @@
+/**
+ * Test <high-weight-genes> directive.
+ **/
+
+describe('<high-weight-genes> directive', function() {
+  beforeEach(module('adage.node'));
+
+  // Thanks to "grunt-html2js" module (called by Grunt), the template file
+  // "node/high_weight_genes.tpl.html" has been converted into an Angular
+  // module of the same name and put in $templateCache before karma runs.
+  beforeEach(module('node/high_weight_genes.tpl.html'));
+
+  var $compile, $httpBackend, $rootScope;
+  beforeEach(inject(function(_$compile_, _$httpBackend_, _$rootScope_) {
+    $compile = _$compile_;
+    $httpBackend = _$httpBackend_;
+    $rootScope = _$rootScope_;
+  }));
+
+  it('should render HTML correctly', function() {
+    var parentScope = $rootScope.$new();
+    parentScope.nodeID = 123;
+    var testHTML = '<high-weight-genes node-id="{{nodeID}}">' +
+        '</high-weight-genes>';
+    var uri = '/api/v0/participation?limit=0&node=' + parentScope.nodeID;
+    var element = $compile(testHTML)(parentScope);
+
+    // Mocked response data of genes:
+    var mockGenes = [
+      {gene: {'systematic_name': 'g1_sys',
+              'standard_name': 'g1_std',
+              'description': 'g1_desc'
+             }
+      },
+      {gene: {'systematic_name': 'g2_sys',
+              'standard_name': 'g2_std',
+              'description': 'hypothetical protein'
+             }
+      },
+      {gene: {'systematic_name': 'g3_sys',
+              'standard_name': 'g3_std',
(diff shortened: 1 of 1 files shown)


interface/src/app/node/high_weight_genes.spec.js
```diff
@@ -0,0 +1,86 @@
...
+ * Test <high-weight-genes> directive.
+ **/
+
+describe('<high-weight-genes> directive', function() {
+  beforeEach(module('adage.node'));
+
+  // Thanks to "grunt-html2js" module (called by Grunt), the template file
+  // "node/high_weight_genes.tpl.html" has been converted into an Angular
+  // module of the same name and put in $templateCache before karma runs.
+  beforeEach(module('node/high_weight_genes.tpl.html'));
```
dongbohu: Line 11 actually can be merged with line 6 into the following line:
```
beforeEach(module('adage.node', 'node/high_weight_genes.tpl.html'));
```
I am writing them separately to emphasize that the HTML template file **node/high_weight_genes.tpl.html** has been converted into a JavaScript module by **grunt-html2js**.

interface/src/app/node/high_weight_genes.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      mockGenes[0].gene.systematic_name);
+    expect(elementScope.genes[1].stdName).toEqual(
+      mockGenes[1].gene.standard_name);
+    expect(elementScope.genes[2].desc).toEqual(
+      mockGenes[2].gene.description);
+    expect(elementScope.hypoPercentage).toEqual(33);
+
+    // Confirm that the rendered HTML contains data in mockGenes too:
+    var tdElements = element.find('td');
+    expect(tdElements.text()).toContain(mockGenes[0].gene.standard_name);
```
dongbohu: It might be possible to get the individual rows in the table and test the text in each cell. I checked the documentation of **angular.element()** but couldn't figure out exactly how to do that:
https://docs.angularjs.org/api/ng/function/angular.element
So I am using **toContain()** function here to roughly confirm that these strings are found in the HTML table that is compiled from the directive.

cgreene:
Thanks for the ping! Won't be able to review this one so I'm going to unmark myself now :)

interface/src/app/node/high_weight_genes.spec.js
```diff
@@ -0,0 +1,86 @@
...
+        'description': 'hypothetical protein'
+      }},
+      {gene: {
+        'systematic_name': 'g3_sys',
+        'standard_name': 'g3_std',
+        'description': 'g3_desc'
+      }}
+    ];
+
+    $httpBackend.expectGET(uri).respond({objects: mockGenes});
```
rzelayafavila: Oh nice! It's good that we are also testing the backend. I'm curious, would testing this here with $httpBackend make this part more of an End to End test than a unit test?

dongbohu: It is more like an **integration test** than **E2E** test. As far as I understand, **E2E** will test the **final** HTML that is rendered on the browser. But the test here is still either a response from the server, or the HTML **template** that is generated by Angular's compilation process (the template that is being tested will eventually be inserted into the final HTML).

dongbohu: By the way, some Angular books or tutorials actually call Karma a framework of **unit and integration test**, because of the its ability to test both individual components (controller, service, etc) and the integrated parts (mocked server response, compiled HTML template in directive, etc).

dongbohu: Check the simple E2E test example in **app/home/home.e2e.js**, you will see that the test interacts with the browser directly.

rzelayafavila: Ok great, thanks for the info!

PR #59, opened 2017-01-22
Non modal heatmap

mhuyck:
This update reworks the sampleBin and nav bar a little and eliminates the modal dialog previously used for displaying the heatmap in favor of a new ui.router state.

No re-scaling has been applied to the heatmap itself yet, but this lays the groundwork for a more flexible display.


dongbohu:
@mhuyck: Do you have a website for demo?

mhuyck:
Great question, @dongbohu. My laptop, where I do development, is not accessible from outside my home network unfortunately. I could have updated the AWS site though. I'm at karate class with the kids and will be with them until bedtime but after that I can run a deployment to update AWS.

That's probably too late to help your review today but you could see it tomorrow morning.

dongbohu:
Thanks @mhuyck. I will review it tomorrow morning.

fabfile/adage_server.py
```diff
@@ -210,6 +210,7 @@ def init_instance():
 def build_interface():
     """ have grunt perform a deployment build for us """
     with cd(CONFIG['interface_dir']), \
+            shell_env(PHANTOMJS_BIN='/usr/local/bin/phantomjs'), \
```
dongbohu: Can you tell me what is the use of this line? What if `phantomjs` is not found in `/usr/local/bin/`

interface/src/app/analyze/analysis/sampleBin.tpl.html
```diff
@@ -1,10 +1,9 @@
...
-      xlink:href="static/assets/testtube-expt2.svg">
-  </image>
-  <g ng-if="sb.heatmapData.samples.length">
-    <text text-anchor="left" x="28" y="43">&times;</text>
-    <text text-anchor="left" x="44" y="43">{{
-      sb.heatmapData.samples.length }}</text>
-  </g>
-</svg>
+<li ui-sref-active="active" class="sample-bin">
+  <a href ui-sref="analysis-detail">
```
dongbohu: `href` is redundant here, right?

dongbohu:
I have finished the review. Most parts I can follow easily, except a few a few questions. Thank you @mhuyck.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -266,24 +266,11 @@ function($log, $cacheFactory, $q, Sample, Activity) {
...
-          return $scope.analysis;
-        }
-      }
-    });
-  };
 }])
 
 .directive('sampleBin', function() {
   return {
+    replace: true,
```
rzelayafavila: Out of curiosity - what happened when we didn't user ``replace: true`` here?

fabfile/adage_server.py
dongbohu: The unit tests that are run by Karma does need phantomjs, but **karma-phantomjs-launcher** is installed locally in `node_modules` sub-directory, not in `/usr/local/bin/`.

mhuyck: I was getting problems with this step during AWS deployment, so in an earlier step (under the `adage-deploy` repo actually... will be a separate PR) I am explicitly installing `phantomjs` using `apt-get` and this is where it lands.

Thanks for pointing out the local installation under `node_modules`, @dongbohu. If that one works okay I'll switch to that path.

interface/src/app/analyze/analysis/sampleBin.tpl.html
mhuyck: Hmm. Good question! This follows the pattern of all the other menu items found in `src/index.html`. Looking up the documentation...

Okay, take a look at the examples on Bootstrap for `navbar`s: https://getbootstrap.com/components/#navbar . All of the `<a>` tags have `href='#'` as an attribute. Strictly speaking, that attribute is optional (https://www.w3.org/MarkUp/1995-archive/Elements/A.html ), but without it the `<a>` tag is not a link but merely a text anchor and I suspect the browser will not behave the way we need it to for a menu.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: Without this, the `<sample-bin>` tags are left in the DOM and it totally messes up the CSS styling for the menu. Took me a while to figure that one out!

mhuyck:
Responding to @dongbohu's comments above:

> (1) Since the heatmap is a separate page now (instead of modal), does it make sense to include the sample IDs in the URL instead of using a generic #/analysis-detail? That way if a user wants to show the heatmap to another user, he/she can just send the URL. (I understand that if there are too many samples involved in the heatmap, the URL may reach its maximum length, so we may need a better solution if this functionality is really needed.)

I like the way you think! I am planning this, (it's on my own TODO list) but it was out of scope for this pull request. I will file an issue to be sure I don't forget to follow up.

> (2) Does it make sense to make the heatmap section as wide as the table that shows the sample information? Right now the heatmap section is narrower than the table.

Another excellent suggestion. As above, I intentionally left this out to keep the PR small. In addition, this is a bit non-trivial to implement with an embedded `vega` element. Already documented on our BitBucket issues list (as issue 21: https://bitbucket.org/greenelab/adage-server/issues/21/update-interface-code-to-preserve , see comment 1). The changes I made in this PR partially address the main portion of this issue, but the rest will come in an upcoming PR.

fabfile/adage_server.py
mhuyck: - [ ] test AWS deployment with the local install of `phantoms` under `node_modules` noted above

.gitignore
```diff
@@ -10,3 +10,4 @@ interface/vendor/
 adage/adage/config.py
 data/gene_history
 data/*.gene_info
+data/bootstrap_cache_*
```
mhuyck: Trying to keep `git status` nice and clean. 😄

interface/src/app/analyze/analysis/sampleBin.tpl.html
dongbohu: Thank you for the links. According to:
https://github.com/angular-ui/ui-router/wiki/Quick-Reference#ui-sref
`href` will be added automatically when the template html is compiled into the actual HTML. YOu can probably test it by removing `href` from the code, if it works, then it is safe to get rid of it.

dongbohu:
@mhuyck: about the two of my comments (adding parameters in heatmap page, and resizing the heatmap window), I understand that they are not trivial to implement. So it is totally fine with me to have them in the issues so that we can take care of them later. Thanks.

interface/src/app/analyze/analysis/sampleBin.tpl.html
mhuyck: Aha... right, I guess Angular is fiddling with the DOM there, so maybe we don't need it. I'll test.

- [ ] test to see if `navbar` items need the `href` property in them

mhuyck: Sure enough. Taking out `href` works fine. Angular will stick it back in when processing the `ui-sref` property as described in the `ui-router` docs referenced by @dongbohu above. Thank you!

interface/src/index.html
```diff
@@ -56,41 +56,41 @@
         <div class="collapse navbar-collapse" collapse="menuCollapsed">
           <ul class="nav navbar-nav">
             <li ui-sref-active="active">
-              <a href ui-sref="home">
+              <a ui-sref="home">
```
dongbohu: Thank you for removing all redundant `href` tags from this file too.

mhuyck: There are a couple more in the `experimentDetail.tpl.html` and `sampleDetail.tpl.html` pages, but I'll need to get those pages to handle state parameters properly before fixing that. Making an issue...

dongbohu: Thank you for creating issues for them. It is good enough for now.

fabfile/adage_server.py
mhuyck: Okay, here is the story on this:

I just got through a test deployment on AWS using the copy of `phantomjs` installed in `node_modules` as @dongbohu mentions above. It did not work.

It turns out that the binary that gets installed with the copy of `karma-phantomjs-launcher` (which is actually installed as `node_modules/phantomjs-prebuilt/bin/phantomjs`) doesn't work (for some unknown reason) on the AWS deployment server. That must have been the root cause of the trouble I was having with the AWS deployment and the manually-installed copy of `phantomjs` worked around it.

I completely agree with @dongbohu that we should be using what's packaged with `karma-phantomjs-launcher` and, in an ideal world, I would debug why it's not working and apply that fix. Given that we are on a deadline, however, I think it's a better use of our time to stick with this workaround for now and have another look at fixing this after our deadline.

We are also planning to transition to a Docker deployment, and that approach may make this whole problem moot anyway. In fact, the copy of `phantomjs` in `node_modules` on my VMware development instance works fine, unlike whatever is happening on AWS. This is probably some obscure node packaging issue. I'll file this as an issue to check out along with converting to Docker. Are you okay with that?

mhuyck:
Okay, I think I have responded to all review comments now. Let me know if you think this is okay to merge @rzelayafavila and @dongbohu!

PR #60, opened 2017-01-24
Add Links on Gene-Gene Network to Node Page

dongbohu:
This PR adds the hyper links on gene-gene networks **edge-tip** window to the **Node** page. 

In **network.less**, a new background color is used to distinguish the edge-tip window from gene-tip window. Some unused lines are also removed in this file..

### DIFF
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -176,18 +176,18 @@ angular.module('adage.gene.network', [
          */
         function getGeneInfo(data) {
           var result = '<div id="title">' + data.label + '</div>';
-          result += '<br />Entrez ID: ' + data.entrezid;
+          result += '<br>Entrez ID: ' + data.entrezid;
           if (data.systematic_name) {
-            result += '<br />Systematic name: ' + data.systematic_name;
+            result += '<br>Systematic name: ' + data.systematic_name;
           }
           if (data.standard_name) {
-            result += '<br />Standard name: ' + data.standard_name;
+            result += '<br>Standard name: ' + data.standard_name;
           }
           if (data.description) {
-            result += '<br />Description: ' + data.description;
+            result += '<br>Description: ' + data.description;
           }
           if (data.aliases) {
-            result += '<br />aliases: ' + data.aliases;
+            result += '<br>aliases: ' + data.aliases;
           }
           return result;
         }
@@ -210,7 +210,8 @@ angular.module('adage.gene.network', [
         function showEdgeTip(data) {
           geneTip.hide(); // Hide gene-tip window (if any) first.
           var rawWeight = data.weight + rawMinWeight;
-          var htmlText = 'Edge weight: ' + rawWeight.toFixed(3);
+          var weightPrecision = 3;
+          var htmlText = 'Edge weight: ' + rawWeight.toFixed(weightPrecision);
           var heavyGenes = [data.gene1.id, data.gene2.id].join(',');
           var target = d3.event.target;
           NodeService.get(
@@ -218,11 +219,11 @@ angular.module('adage.gene.network', [
             function success(response) {
               var i = 0, n = response.objects.length;
               var anchorTag;
-              htmlText += '<br />' + n + (n > 1 ? ' nodes are ' : ' node is ');
+              htmlText += '<br>' + n + (n > 1 ? ' nodes are ' : ' node is ');
               htmlText += 'related to both genes' + (n > 0 ? ':' : '.');
               for (; i < n; ++i) {
                 anchorTag = '<a href="#/node/' + response.objects[i].id + '">';
-                htmlText += '<br />* ' + anchorTag + response.objects[i].name;
+                htmlText += '<br>* ' + anchorTag + response.objects[i].name;
                 htmlText += '</a>';
               }
               edgeTip.html(htmlText);
@@ -232,7 +233,7 @@ angular.module('adage.gene.network', [
               var message = 'Failed to get node info for gene edge: ' +
                   response.statusCode + ' ' + response.statusText;
               $log.error(message);
-              htmlText += '<br />' + message + '. Please try again later.';
+              htmlText += '<br>' + message + '. Please try again later.';
               edgeTip.html(htmlText);
               edgeTip.show(data, target);
             }
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -5,6 +5,7 @@
 angular.module('adage.gene.network', [
   'ui.router',
   'ui.bootstrap',
+  'ngResource',
   'rzModule'
 ])
 
@@ -15,6 +16,19 @@ angular.module('adage.gene.network', [
       main: {
         templateUrl: 'gene/network/network.tpl.html',
         controller: 'GeneNetworkCtrl as ctrl'
+      }
+    },
+    // When "gene_network" state exits, remove "edge-tip" elements from the DOM.
+    // (Without this function, the edge tip window will be visible in other
+    // states too.) The "gene-tip" element will disappear when mouse is out,
+    // so we don't need to worry about it.
+    onExit: function() {
+      var elements = document.getElementsByClassName('edge-tip');
+      if (elements) { // This should be always true.
+        var n = elements.length;
+        for (var i = 0; i < n; ++i) {
+          elements[i].parentNode.removeChild(elements[i]);
+        }
       }
     },
     data: {pageTitle: 'Gene Network'}
@@ -187,20 +201,22 @@ angular.module('adage.gene.network', [
           NodeService.get(
             {'heavy_genes': heavyGenes, 'limit': 0},
             function success(response) {
-              var n = response.objects.length;
+              var i = 0, n = response.objects.length;
+              var uri;
               str += '<br />' + n + (n > 1 ? ' nodes are ' : ' node is ');
               str += 'related to both genes' + (n > 0 ? ':' : '.');
-              for (var i = 0; i < n; ++i) {
-                str += '<br />* ' + response.objects[i].name + '</li>';
+              for (; i < n; ++i) {
+                uri = '<a href="#/node/' + response.objects[i].id + '">';
+                str += '<br />* ' + uri + response.objects[i].name + '</a>';
               }
-
               edgeTip.html(str);
               edgeTip.show(data, target);
             },
-            function error(err) {
-              $log.error(
-                'Failed to get node info for gene edge: ' + err.statusText);
-              str += 'Can\'t get node information, please try again later.';
+            function error(response) {
+              var message = 'Failed to get node info for gene edge: ' +
+                  response.statusCode + ' ' + response.statusText;
+              $log.error(message);
+              str += message + '. Please try again later.';
               edgeTip.html(str);
               edgeTip.show(data, target);
             }
@@ -237,14 +253,22 @@ angular.module('adage.gene.network', [
         geneTip.html(getGeneInfo);
         network.onGene('mouseover.custom', geneTip.show);
         network.onGene('mouseout.custom', geneTip.hide);
-        network.onEdge('mouseover.custom', showEdgeTip);
-        network.onEdge('mouseout.custom', hideEdgeTip);
+        network.onEdge('click.custom', showEdgeTip);
 
         // Draw network svg with legend and filter.
         network.showLegend()
           .filter(0, self.maxNodeNum)
           .draw();
-      }
+
+        // Add event handler so that edge-tip will be hidden when clicked:
+        var edgeTips = document.getElementsByClassName('edge-tip');
+        if (edgeTips) {
+          var n = edgeTips.length;
+          for (var i = 0; i < n; ++i) {
+            edgeTips[i].onclick = hideEdgeTip;
+          }
+        }
+      } // End of drawNetwork()
 
       EdgeService.get(
         {genes: $stateParams.genes, mlmodel: $stateParams.mlmodel, limit: 0},
@@ -263,9 +287,11 @@ angular.module('adage.gene.network', [
           self.statusMessage = '';
           drawNetwork();
         },
-        function error(err) {
-          $log.error('Failed to get edges: ' + err.statusText);
-          self.statusMessage = 'Connection to server failed';
+        function error(response) {
+          var message = 'Failed to get edges: ' + response.statusCode +
+              ' ' + response.statusText;
+          $log.error(message);
+          self.statusMessage = message + '. Please try again later';
         }
       );
     }
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -18,18 +18,22 @@ angular.module('adage.gene.network', [
         controller: 'GeneNetworkCtrl as ctrl'
       }
     },
-    // When "gene_network" state exits, remove "edge-tip" elements from the DOM.
-    // (Without this function, the edge tip window will be visible in other
-    // states too.) The "gene-tip" element will disappear when mouse is out,
-    // so we don't need to worry about it.
+    // When "gene_network" state exits, remove "gene-tip" and "edge-tip"
+    // elements from the DOM. (Without this function, the tips window will
+    // be visible in other states too.)
     onExit: function() {
-      var elements = document.getElementsByClassName('edge-tip');
-      if (elements) { // This should be always true.
-        var n = elements.length;
-        for (var i = 0; i < n; ++i) {
-          elements[i].parentNode.removeChild(elements[i]);
+      var removeTips = function(className) {
+        var elements = document.getElementsByClassName(className);
+        var i, n;
+        if (elements) { // This should be always true.
+          n = elements.length;
+          for (i = 0; i < n; ++i) {
+            elements[i].parentNode.removeChild(elements[i]);
+          }
         }
-      }
+      };
+      removeTips('gene-tip');
+      removeTips('edge-tip');
     },
     data: {pageTitle: 'Gene Network'}
   });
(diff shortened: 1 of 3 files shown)


dongbohu:
Live demo:
http://165.123.67.202/#/gene_network/?genes=6546
Unlike gene-tip winow (which will show/hide when mouse is over/out), the edge-tip window will now show up only when a user clicks an edge. It will disappear only when the user clicks this edge-tip window. 

Some of the edge-tip window includes a hyper link that points to the node page, for example, in the network of the above URL, the tip window of the edge between the center gene and **sdhD** has a link to **Node72Pos**. If you click it, the page will be directed to that node's page.

dongbohu:
You can ignore the two new issues reported by codeclimate.

dongbohu:
I am not 100% satisfied with the current implementation, because the mouse events to show/hide gene-tip window and those to show/hide edge-tip window are inconsistent now. 

Another feature that I was thinking of (but not implemented in this PR) is to show the two related gene names in edge-tip window, and also allow user to hide the edge-tip window when the same edge is clicked again. Not sure whether that is a good idea.

dongbohu:
After talking to @cgreene, I changed the code so that both gene-tip and edge-tip will be hidden when clicked. So the event handlers are consistent now, and we will be able to add hyper links in the gene-tip window too (in the future).

mhuyck:
I was able to get the UI into this state:

screen shot 2017-01-26 at 3 36 24 pm

When drawing a new tip box of either type, we should probably first clear *any* existing gene- and edge-tip boxes first.

dongbohu:
Right now the edge-tip and gene-tip are independent. Is it possible that a user may want to keep both windows visible for whatever user case?

interface/src/app/gene/network/network.js
```diff
@@ -235,16 +255,29 @@ angular.module('adage.gene.network', [
           .call(edgeTip);
 
         geneTip.html(getGeneInfo);
-        network.onGene('mouseover.custom', geneTip.show);
-        network.onGene('mouseout.custom', geneTip.hide);
-        network.onEdge('mouseover.custom', showEdgeTip);
-        network.onEdge('mouseout.custom', hideEdgeTip);
+        network.onGene('click.custom', geneTip.show);
+        network.onEdge('click.custom', showEdgeTip);
```
mhuyck: It sounds as if you've discussed the design of this interaction with @cgreene directly, so I won't second-guess the rationale for this change, but I do want to add my 2¢ here. The nice thing about responding to a `mouseover` event in some way is that the feature is easy to discover. It's not necessarily obvious that clicking will do anything. In fact, I had to read the comments on this PR and in the code a bit to realize the link feature was being added to the tip that comes up when clicking on an *edge*. Not sure I would have thought to click while hovering over an edge otherwise.

If you've done user testing that indicates that this is not a problem, then all is good. I anticipate that this feature will be hard to discover for a new user, though.

interface/src/app/gene/network/network.js
```diff
@@ -17,6 +18,23 @@ angular.module('adage.gene.network', [
...
+        if (elements) { // This should be always true.
+          n = elements.length;
+          for (i = 0; i < n; ++i) {
+            elements[i].parentNode.removeChild(elements[i]);
+          }
+        }
+      };
+      removeTips('gene-tip');
+      removeTips('edge-tip');
+    },
```
mhuyck: I know we have to do it this way because the network is done in d3, but we are getting pretty deep into direct DOM manipulation here and that can be tricky to get right when working with Angular. Perhaps the best way to allay that concern would be to have a suite of UI tests that automatically verifies these features are working. I think we will be very thankful to have tests like that whenever we decide to update Angular to a new version.

We are on a tight timeline right now, so it probably doesn't make sense to add in tests to *this* pull request, but perhaps we should file an issue so we remember to do something about it before we contemplate any Angular updates. What do you think?

interface/src/app/gene/network/network.js
```diff
@@ -187,20 +205,22 @@ angular.module('adage.gene.network', [
           NodeService.get(
             {'heavy_genes': heavyGenes, 'limit': 0},
             function success(response) {
-              var n = response.objects.length;
+              var i = 0, n = response.objects.length;
+              var uri;
```
mhuyck: Now that this code is building a `uri` alongside the `str`, should we give `str` a more descriptive name? Maybe something as simple as `tipStr` would make this bit of code a little easier to follow.

interface/src/app/gene/network/network.js
```diff
@@ -187,20 +205,22 @@ angular.module('adage.gene.network', [
...
               str += '<br />' + n + (n > 1 ? ' nodes are ' : ' node is ');
               str += 'related to both genes' + (n > 0 ? ':' : '.');
-              for (var i = 0; i < n; ++i) {
-                str += '<br />* ' + response.objects[i].name + '</li>';
+              for (; i < n; ++i) {
+                uri = '<a href="#/node/' + response.objects[i].id + '">';
```
mhuyck: Hmm. Strictly speaking, this is more than just the `uri` here. This is the opening anchor tag. Let me know if you think this is too picky, but I find that precise variable names can make a big difference in debugging.

interface/src/app/gene/network/network.js
```diff
@@ -187,20 +205,22 @@ angular.module('adage.gene.network', [
...
+                str += '<br />* ' + uri + response.objects[i].name + '</a>';
```
mhuyck: One more comment here on the subject of tension between DOM manipulation and Angular and then I'll stop harping on it: if this code were written in a more Angular-compatible way, we could use Angular's excellent HTML template features to format this tip.

I suspect this level of integration of Angular with d3 to be somewhat tricky, so perhaps it's not practical to do it that way for now.

interface/src/app/gene/network/network.js
```diff
@@ -187,20 +205,22 @@ angular.module('adage.gene.network', [
...
             },
-            function error(err) {
-              $log.error(
-                'Failed to get node info for gene edge: ' + err.statusText);
-              str += 'Can\'t get node information, please try again later.';
+            function error(response) {
+              var message = 'Failed to get node info for gene edge: ' +
+                  response.statusCode + ' ' + response.statusText;
+              $log.error(message);
+              str += message + '. Please try again later.';
```
mhuyck: Did you test this error case? I suspect you will need a `<br>` (as you have at line 214 above) in order to have the whole `str` formatted properly.

interface/src/app/gene/network/network.js
```diff
@@ -5,6 +5,7 @@
 angular.module('adage.gene.network', [
   'ui.router',
   'ui.bootstrap',
+  'ngResource',
```
mhuyck: Aha! Never mind, I found it... just wasn't in the changes for this PR. Puzzled why it wasn't a problem to have it missing before, though!

interface/src/app/gene/network/network.js
```diff
@@ -235,16 +255,29 @@ angular.module('adage.gene.network', [
...
+          var i, n;
+          if (elements) {
+            n = elements.length;
+            for (i = 0; i < n; ++i) {
+              elements[i].onclick = handler;
+            }
+          }
+        };
+        addClickHandler('gene-tip', geneTip.hide);
+        addClickHandler('edge-tip', hideEdgeTip);
```
mhuyck: I'm getting picky again, but it bugs me that line 278 says `geneTip.hide` and line 279 says `hideEdgeTip`. Is there a reason these don't parallel each other more?

mhuyck: Aha. I see your `FIXME` comment at line 235 now. You can disregard my picky comment since you've got a task to follow up. Maybe you could check to see if that is still an issue now that these are handled as clicks rather than mouse movement events.

interface/src/app/gene/network/network.less
```diff
@@ -41,15 +41,11 @@ text.legend-title {
...
-  #title {
-    text-align: center;
-  }
-
   line-height: 1;
   font-weight: bold;
   padding: 12px;
-  background: rgba(100, 150, 200, 0.9);
-  color: #fff;
+  background: Gainsboro;
```
mhuyck: Awesome. I don't think I've ever seen anyone use color names other than `black`, `white`, and maybe `red`.

interface/src/app/gene/network/network.tpl.html
```diff
@@ -8,7 +8,6 @@
      rz-slider-model="ctrl.slider.position"
      rz-slider-options="ctrl.slider.options">
   </rzslider>
-  <br />
```
mhuyck: 👍

mhuyck:
@dongbohu: Just saw your reply to my earlier note about the overlapping tips. I can't imagine a use case in which having both boxes visible would not be considered buggy. It just seems confusing to me.

On the other hand, if it *is* a desirable feature, we should probably make it possible to show tips for multiple nodes at a time and multiple edges at a time, but that would require some more work on state management since we're not using Angular to handle the state for us with this feature.

interface/src/app/gene/network/network.js
dongbohu: Angular didn't complain before because `ngResource` has been included in other modules. I noticed this issue when writing unit tests for `download` module.

interface/src/app/gene/network/network.js
dongbohu: I plan to write more unit tests and E2E tests after this deadline. I have put it on the list of "Low priority tasks" on the shared google doc.

interface/src/app/gene/network/network.js
dongbohu: Sure. What about `htmlText`?

interface/src/app/gene/network/network.js
dongbohu: That is true. It is called `uri` mainly because in my original implementation it was acually an URI. After changing the code, I forgot to give it a more reasonable name. How about `anchorTag`?

interface/src/app/gene/network/network.js
dongbohu: I took a look at the source code of d3-tips library when I first started to use it. Due to its limited API, the integration with Angular is not easy (if possible). I may dig deeper on it some time later.

interface/src/app/gene/network/network.js
dongbohu: That's true. Thanks.

interface/src/app/gene/network/network.js
dongbohu: I added one line of text `Click the gene or edge to see more details` above the network graph, but that seems a little distracting. I would like to reply on the upcoming **Help** tab to describe the usage of each module.

interface/src/app/gene/network/network.js
dongbohu: That's true. Since the edge tip now can be hidden only by clicking the the tip window, it can be hidden synchronously. I will remove this "workaround" function.

interface/src/app/gene/network/network.js
cgreene: I agree that discovery may be tough. Can we create an issue to improve the discoverability of this feature? I agree that we should do it. I think that we'll have to get actual users touching it to find out if it's a problem. The issue will help us remember to do this.

dongbohu:
I tend to agree with what @mhuyck says about showing only one of **gene-tip** or **edge-tip** windows. What do you think @cgreene and @rzelayafavila?

interface/src/app/gene/network/network.js
dongbohu: Sure. I will create an issue.

cgreene:
@dongbohu I agree with one tool tip at a time.

dongbohu:
@cgreene: Thanks. I will change the code so that only one tip-window will be shown.

dongbohu:
I think I have addressed all comments made by @mhuyck. Please let me know if I haven't.

interface/src/app/gene/network/network.js
```diff
@@ -175,57 +193,52 @@ angular.module('adage.gene.network', [
...
+        /**
+         * Callback function to show edge tips when an edge is clicked.
          * @param {edge_data} data;
          * @return {void}.
          */
         function showEdgeTip(data) {
+          geneTip.hide(); // Hide gene-tip window (if any) first.
           var rawWeight = data.weight + rawMinWeight;
-          var str = 'Edge weight: ' + rawWeight.toFixed(3);
+          var htmlText = 'Edge weight: ' + rawWeight.toFixed(3);
```
rzelayafavila: Nice, I like this more descriptive variable name :+1: . Should the 3 be defined somewhere above this as a setting/parameter rather than a magic number of how many decimal digits we want?

interface/src/app/gene/network/network.js
```diff
@@ -175,57 +193,52 @@ angular.module('adage.gene.network', [
...
             {'heavy_genes': heavyGenes, 'limit': 0},
             function success(response) {
-              var n = response.objects.length;
-              str += '<br />' + n + (n > 1 ? ' nodes are ' : ' node is ');
-              str += 'related to both genes' + (n > 0 ? ':' : '.');
-              for (var i = 0; i < n; ++i) {
-                str += '<br />* ' + response.objects[i].name + '</li>';
+              var i = 0, n = response.objects.length;
+              var anchorTag;
+              htmlText += '<br />' + n + (n > 1 ? ' nodes are ' : ' node is ');
```
rzelayafavila: This is a very minor thing - according to the Google HTML style guide it is better to not close void elements (so writing <br> instead of <br /> here and in the ones below).

interface/src/app/gene/network/network.js
```diff
@@ -175,57 +193,52 @@ angular.module('adage.gene.network', [
...
             },
-            function error(err) {
-              $log.error(
-                'Failed to get node info for gene edge: ' + err.statusText);
-              str += 'Can\'t get node information, please try again later.';
-              edgeTip.html(str);
+            function error(response) {
+              var message = 'Failed to get node info for gene edge: ' +
+                  response.statusCode + ' ' + response.statusText;
+              $log.error(message);
```
rzelayafavila: :+1:  Nice logging of the error, and also handling the case when there is an error from the backend and how to present it to the user.

interface/src/app/gene/network/network.less
rzelayafavila: Hahah, same here!

interface/src/app/gene/network/network.js
```diff
@@ -175,57 +193,52 @@ angular.module('adage.gene.network', [
...
+         * @param {gene_data} data;
+         * @return {void}.
+         */
+        function showGeneTip(data) {
+          edgeTip.hide(); // Hide edge-tip window (if any) first.
+          geneTip.show(data);
+        }
+
+        /**
+         * Callback function to show edge tips when an edge is clicked.
```
dongbohu: Exactly.

interface/src/app/gene/network/network.js
dongbohu: I didn't know that. Thanks. I will correct them.

interface/src/app/gene/network/network.js
dongbohu: Good question. Let me create a variable for it instead of using the magic 3.

dongbohu:
I have addressed @rzelayafavila's comments. Thanks.

interface/src/app/gene/network/network.js
mhuyck: Unit tests, FTW!

interface/src/app/gene/network/network.js
mhuyck: Right, good call. This sort of test will be a little tricky to get working but it will be good to have.

interface/src/app/gene/network/network.js
mhuyck: works for me.

interface/src/app/gene/network/network.js
mhuyck: Definitely a step in the "more descriptive" direction, but `html` only describes the *format*, not the logical *function* of the variable. I guess a fully-verbose description would be the "HTML-formatted text of the tip to be displayed upon clicking a node." Maybe "nodeTipHTML" would be a succinct way of saying that. But this is a minor quibble so it's up to you whether you think it's worth fixing.

interface/src/app/gene/network/network.js
```diff
@@ -158,74 +176,70 @@ angular.module('adage.gene.network', [
...
-            function error(err) {
-              $log.error(
-                'Failed to get node info for gene edge: ' + err.statusText);
-              str += 'Can\'t get node information, please try again later.';
-              edgeTip.html(str);
+            function error(response) {
+              var message = 'Failed to get node info for gene edge: ' +
+                  response.statusCode + ' ' + response.statusText;
+              $log.error(message);
+              htmlText += '<br>' + message + '. Please try again later.';
```

interface/src/app/gene/network/network.js
mhuyck: Agreed on all counts. I may be wrong about how confusing this is so not worth adding the feature if it's not needed.

interface/src/app/gene/network/network.js
```diff
@@ -176,76 +176,70 @@ angular.module('adage.gene.network', [
...
+        /**
+         * Callback function to show edge tips when an edge is clicked.
          * @param {edge_data} data;
          * @return {void}.
          */
         function showEdgeTip(data) {
+          geneTip.hide(); // Hide gene-tip window (if any) first.
           var rawWeight = data.weight + rawMinWeight;
-          var str = 'Edge weight: ' + rawWeight.toFixed(3);
+          var weightPrecision = 3;
```
mhuyck: nice!

interface/src/app/gene/network/network.js
mhuyck: Nice catch.

rzelayafavila:
You got all of the things I commented on too, thanks @dongbohu!

PR #66, opened 2017-02-03
Fix glitch in test_get_annotations_get_param and elsewhere.

mhuyck:
This is a quick PR to resolve the testing glitch related to sporadic `fixtureless` errors I encountered yesterday. Details of the changes from the change log:

>This fixes the sporadic problem we encountered when fixtureless happened to
generate test data that didn't get URL encoded properly, causing an unwarranted
testing failure.
>
>TastyPie's api_client.get() uses a separate input parameter
(data=) to handle URL parameters. Although it works sometimes for us to
manually assemble a complete URI in the testing code, we are also then
responsible for properly URL-encoding the parameters. This set of changes now
calls api_client.get() with a data= object with those URL parameters to let
TastyPie handle this step for us.

### DIFF
--- a/adage/analyze/tests.py
+++ b/adage/analyze/tests.py
@@ -366,8 +366,24 @@ class APIResourceTestCase(ResourceTestCaseMixin, TestCase):
             **self.test_experiment))
 
         # Create a few annotation types to retrieve with the API
-        self.at_count = 10
-        factory.create(AnnotationType, self.at_count)
+        ann_type = AnnotationType(
+            typename="type name % one",
+            description="description with more words 1",
+        )
+        ann_type.save()
+        ann_type = AnnotationType(
+            typename="type name & two",
+            description="description with more words 2",
+        )
+        ann_type.save()
+        ann_type = AnnotationType(
+            typename="type name # three",
+            description="description with more words 3",
+        )
+        ann_type.save()
+        self.at_count = 3
+        factory.create(AnnotationType, 7)
+        self.at_count += 7
         self.annotationtype_listURI = self.baseURI + 'annotationtype/'
         self.annotationtypeURI = self.annotationtype_listURI + str(
             self.random_object(AnnotationType).id) + '/'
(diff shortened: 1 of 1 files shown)


adage/analyze/tests.py
```diff
@@ -432,25 +448,25 @@ def call_get_API(self, uri):
         '''
         Helper method: asserts that GET method is allowed via input URI.
         '''
-        resp = self.api_client.get(uri, data={'format': 'json'})
```
rzelayafavila: Just checking - removing ``data={'format': 'json'}`` here is not because of the glitch right? It just isn't necessary when making test requests with the test's 'api_client'?

adage/analyze/tests.py
```diff
@@ -625,19 +642,21 @@ def test_activity_mlmodel_filter(self):
         """
         Test "activity/?sample=<id>&mlmodel=<ml_id>" API.
         """
-        uri = (self.baseURI + "activity/?sample=" +
-               str(self.random_object(Sample).id) + "&mlmodel=" +
-               str(self.random_object(MLModel).id))
+        uri = self.baseURI + "activity/"
+        data = {
```
rzelayafavila: Nice encoding this in the request data! :+1:

adage/analyze/tests.py
```diff
@@ -366,8 +366,24 @@ def setUp(self):
             **self.test_experiment))
 
         # Create a few annotation types to retrieve with the API
-        self.at_count = 10
-        factory.create(AnnotationType, self.at_count)
+        ann_type = AnnotationType(
+            typename="type name % one",
```
rzelayafavila: So these first 3 test cases check that the ``%, &, #`` characters get encoded appropriately in the url?

mhuyck: Keep in mind that the Django test is not launching a web server. I think this means that URLs are not *actually* being generated and used, so this is not exactly a live API test, but `api_client.get()` is the right test harness for this.

These are examples I found that broke the test in the same place that the fixtureless-generated data were doing. More importantly, the fix I applied to the tests made these examples stop breaking the test.

Note the comment in the [docs](http://django-tastypie.readthedocs.io/en/latest/testing.html#get) that the method "Optionally accepts a `data` kwarg, which in the case of `GET`, lets you send along `GET` parameters. " This was the hint I needed that we should be doing it that way rather than building the URL manually.

adage/analyze/tests.py
mhuyck: Correct. I thought we had previously gotten rid of all the `format=json`s in our `api_client` calls, but these ones slipped by until I saw this calling method.

adage/analyze/tests.py
mhuyck: This is the actual bug fix. According to the (documentation)[http://django-tastypie.readthedocs.io/en/latest/testing.html#get], we should pass all parameters this way, even for our `GET` endpoints.

mhuyck:
Responded to all questions from @rzelayafavila. Thank you for having a look, @dongbohu and @rzelayafavila! Let me know if you have any other questions. I plan to merge this afternoon.

adage/analyze/tests.py
rzelayafavila: Right, that was what I was thinking, that makes sense.

adage/analyze/tests.py
rzelayafavila: Right - I agree, we should always do this. Also it makes sense that it was breaking the tests because of the strange characters that fixtureless produced were being manually concatenated onto the url string right?

mhuyck:
Exactly.

adage/analyze/tests.py
mhuyck: I replied via e-mail, but it never showed up...

This is exactly right. Those special characters I added to the above test (`#`, `&`, and `%`) are all specially treated in URLs and need to be properly escaped if they are in a parameter name or value.

PR #68, opened 2017-02-14
Creating Machine Learning Model selector, and adding it to gene search.

rzelayafavila:
By being able to select a Machine Learning Model, users can filter their gene searches by the organism tied to that Model.


rzelayafavila:
Oh by the way, forgot to say - you can access the interface and play around with the machine learning model selector here: http://165.123.67.154:8000/#/gene_search

mhuyck:
I have a general question about this PR, maybe more for @cgreene than @rzelayafavila. I know we ultimately need to allow the user to select the `mlmodel` and we have been careful to pull `organism` out into a variable with a default value wherever possible, but why do we need to make this parameter selectable right **now**? Don't we only have data for the one *Pseudomonas* model at the moment?

I thought the plan for the initial release was for a single model. If that has to change, then so be it, but it seems to be a fairly big and fundamental change to be throwing in just as we are pushing toward a release.

mhuyck:
Have you loaded data for two separate models or organisms to test how this works? That needs to be tested one way or another. I'd also like to just get a "feel" for how this will work with more than one option to select from in the drop-down.

cgreene:
I agree that this may be a somewhat large change. Two things lead me to leaning in favor of targeting this for our first release:
* If we have URLs that don't work with multiple models, those URLs might change after the initial release. If users have saved them, we will have to consider how to address any form of remapping. We may wish to find these now.
* I think we have 2 different ML models ( @tj8901nm please correct me if I'm wrong) - a 50 node ADAGE model and the 300 node eADAGE model.

I'm not totally set on pushing ahead with this before release. Perhaps you, @dongbohu and @rzelayafavila can discuss in your meeting this week?

mhuyck:
We just met and decided to forge ahead with this change for two reasons:

1. the point @cgreene makes about having stable URLs is a good one, we all agreed
2. the number of things affected by this is fewer than I was initially thinking, and we have already anticipated this change in several of the areas that will be affected so it shouldn't be too hard to roll with it.

@rzelayafavila and @dongbohu, please chime in if I've misrepresented what we discussed!

cgreene:
Thanks for the update @mhuyck! I would be ok with expanding the scope of this PR to include fixes to the various URLs so that everything lands at once. Up to what you guys prefer.

rzelayafavila:
@cgreene @mhuyck @dongbohu - Ok, I will include the fixes to the known URLs in this pull request so that everything is hooked up and that the landing on the URLs works.

rzelayafavila:
Ok @mhuyck @dongbohu @cgreene I have made the necessary changes to pass the selected MLModel as a parameter to the ``gene_network`` page. Thanks!

interface/src/app/mlmodel/mLModelComponents.js
```diff
@@ -0,0 +1,25 @@
+angular.module('adage.mlmodel.components', [
+  'adage.mlmodel.resource'
+])
+
+.component('mLModelSelector', {
```
dongbohu: Great! I like **component**!

interface/src/app/gene/gene-network.tpl.html
```diff
@@ -2,17 +2,28 @@
   <h1 class="page-header">
     Gene Network
   </h1>
-  <autocomplete-search-panel organism="searchCtrl.organism" autocomplete="searchCtrl.autocomplete"
-    ng-if="searchCtrl.autocomplete">
+
+  <m-l-model-selector selected-m-l-model="searchCtrl.selectedMLModel">
```
dongbohu: I found this tag a little weird. Do you think **ml-model-selector** and **selected-ml-model** (and **ml-model** later) will be better? (This is trivial, if you prefer "m-l-...", I am fine with it.

interface/src/app/mlmodel/mLModelComponents.js
```diff
@@ -0,0 +1,25 @@
...
+  templateUrl: 'mlmodel/selector.tpl.html',
+  bindings: {
+    selectedMLModel: '='
+  },
+  controller: ['MLModel', '$log', function(MLModel, $log) {
+    var self = this;
+
+    MLModel.get(
+      function success(response) {
```
dongbohu: Do you think `{limit: 0}` should be specified here in case the database has many ML models available? (I know we have at most two now, but we may have many more in the future.

interface/src/app/mlmodel/mLModelComponents.js
```diff
@@ -0,0 +1,25 @@
...
+
+.component('mLModelSelector', {
+  templateUrl: 'mlmodel/selector.tpl.html',
+  bindings: {
+    selectedMLModel: '='
+  },
+  controller: ['MLModel', '$log', function(MLModel, $log) {
+    var self = this;
+
+    MLModel.get(
```
dongbohu: Do you think `{limit: 0}` should be specified here in case the database has many ML models available? (I know we have at most two now, but we may have many more in the future.

interface/src/app/gene/searchFew/searchFew.js
```diff
@@ -14,7 +14,8 @@ angular.module('adage.gene.searchFew', [
     }],
     restrict: 'E',
     scope: {
-      organism: '=',
+      organism: '@',
+      mLModel: '@',
```
mhuyck: Similar to my comment elsewhere, we could make this an `mlModel` and I think you'd be able to refer to `ml-model`s in the `.tpl.html` files.

interface/src/app/mlmodel/mLModelComponents.js
rzelayafavila: Mmm good point - I guess right now we have a maximum limit of returning 50 objects at a time from Tastypie, but we should generalize it so that all Machine Learning Models come back. I'll put this in as a variable at the beginning of the controller that we can change.

interface/src/app/gene/gene-network.tpl.html
rzelayafavila: Hmmm good question - I guess the names for these tags are converted from the camelCase names of the component and variables by Angular. So instead of ``mLModelSelector`` and ``selectedMLModel`` we could name them ``mlModelSelector`` and ``selectedMlModel``. I was using the rules for initials in camelCase, but I don't have a strong opinion on this, what do you think?

interface/src/app/mlmodel/mLModelComponents.js
mhuyck: I agree that it's a good idea to use either `{limit: 0}` or implement pagination to avoid future bugs.

interface/src/app/mlmodel/mLModelComponents.js
```diff
@@ -0,0 +1,25 @@
...
+      function success(response) {
+        self.availableMLModels = response.objects;
+      },
+      function error(err) {
+        $log.error('Failed to get MLModels from REST API: ' + err.statusText);
+        self.errorMessage = 'Failed to get Machine Learning Models from server';
```
mhuyck: I like this: simple error message with details logged to the console for follow-up. Although this simple API interaction is not likely to produce too many problems, it's *always* helpful to leave ourselves tools for following up real-world errors.

interface/src/app/mlmodel/resource.js
```diff
@@ -0,0 +1,9 @@
+angular.module('adage.mlmodel.resource', ['ngResource'])
+
+.factory('MLModel', ['$resource', function($resource) {
+  return $resource(
+    '/api/v0/mlmodel/'
+  );
+}])
```
mhuyck: This is such a small component, I think it would be okay to have the resource as part of `mlModelComponents`. There is nothing technically wrong with having this factored out and I'd say it should be required if we needed to break up a larger module but there is also something to be said for having fewer files to flip through.

I guess I'd call this a "weak" recommendation to add this factory to `mlModelComponents`. Feel free to ignore if you don't agree.

interface/src/app/mlmodel/selector.tpl.html
```diff
@@ -0,0 +1,13 @@
+<div class="well">
+  <h3>
+    Machine Learning Model
+  </h3>
```
mhuyck: My preference would be to keep the actual component compact and easily "insertable," while leaving things like headers and `class="well"` up to the page that does the integrating. I haven't been great about doing this in my own pages yet, but I'd like us to start moving towards a cleaner separation of concerns between *modules* as building blocks and *layouts*, where those building blocks are assembled into a unified view or whole page.

interface/src/app/gene/gene-network.tpl.html
mhuyck: Exactly my thoughts. See review just submitted...

dongbohu: I just feel the `m-l-` part of the tag a little bizarre. If it is acceptable to the other reviewers, I am fine with it.

rzelayafavila: Ah I see, well especially in the light of @mhuyck's comments, I will change it to ``mlModelSelector`` and ``selectedMlModel`` to make it both easier to read and also improve the annoying tag names. Thanks!

interface/src/app/mlmodel/resource.js
```diff
@@ -0,0 +1,9 @@
+angular.module('adage.mlmodel.resource', ['ngResource'])
+
+.factory('MLModel', ['$resource', function($resource) {
```
rzelayafavila: @mhuyck and @dongbohu : Following the logic of the naming of the other variables and components in this PR, do you think the name of this resource should change to ``MlModel``?

dongbohu: Sounds good to me. As @mhuyck pointed out earlier, another option is not to create this service at all. Just create a resource inside your component code, since it is so simple and it is not being used anywhere else in the code.

interface/src/app/mlmodel/resource.js
rzelayafavila: @mhuyck and @dongbohu : I actually like having this resource be separate from the components file, since it is not actually a component (in the Angular sense of the term "component"), and it has its own specific function - to communicate with the REST API. From a pragmatic point of view, I think it makes sense to keep this modules loosely coupled like this, especially if we end up needing to use this resource in other places.

interface/src/app/mlmodel/selector.tpl.html
rzelayafavila: Aha - that makes sense. So maybe should I move these headers and styling to the ``gene-network.tpl.html`` template and just have the ``<select>`` element itself (with all its required logic) in this component template?

mhuyck: Exactly. `gene-network.tpl.html` is an example of the *layout* in the separation of concerns I'm trying to establish. Ideally, everything in that file should focus exclusively on assembling *modules* (or *components*, to use the specific Angular example) and where they fall in the flow of the page, how they resize with the window, etc.

interface/src/app/mlmodel/resource.js
dongbohu: Either way will be fine with me. When I first AngularJS last year, I liked to have separate API services when possible. (You can find them in my old code). The reason why I am getting more conservative is mainly because of two reasons:
(1) Sometimes they are so simple, just returns a `$resource(url)`;
(2) Many services created this way were actually only used in the current module.

Originally I thought that creating a service may improve the frontend efficiency, because a service is always a singleton. But if you read $resource doc, you will see that `$resource(url)` actually returns a light-weight "class" type (which is a constructor function). All the weight-lifting part will be done later by the methods such as `get`, `post`, etc.

Anyway, my two cents.

mhuyck: I'm fine with that. As I said, it's a weak recommendation and I can see it working both ways. The only thing I'll add here is that I think the resource is just as loosely-coupled whether it lives in this file or in the other one. The level of coupling has more to do with its abstraction as a separate component than what file it lives in.

rzelayafavila: Well, it would be loosely coupled if it just lived in the other file, but @dongbohu mentioned not creating the service at all, just calling the resource inside the component code, which is why I mentioned that. Also, I think it doesn't add too much overhead inconvenience to have it in this file, and having the other file just be for angular ``components`` in the same way you might have a file for ``directives``.

interface/src/app/mlmodel/selector.tpl.html
rzelayafavila: Ok, that separation of concerns makes sense. I just pushed those changes!

rzelayafavila:
Ok @mhuyck @dongbohu, I just pushed some changes, let me know how it all looks now!

interface/src/app/mlmodel/resource.js
mhuyck: Aha. I hadn't seen @dongbohu's comment when I wrote mine.

Architecturally, it is important to maintain the loose coupling @rzelayafavila describes. This is another example of separation of concerns: the UI code should not have direct knowledge of API endpoints but rather work their magic through a collection of thoughtfully managed services.

This is the same logic as abstracting database calls away from the application code as we do with Django's abstract model design. While there may be a small performance penalty involved, what we are optimizing for is maintainability and *programmer* efficiency. Slow running code is a concern too, of course, but that comes in at a distant second priority to readable and maintainable code.

PR #73, opened 2017-02-14
Add ttest to greenelab.stats

mhuyck:
This set of changes incorporates a ttest library from npm, repackages it for use in Angular via our new greenelab.stats module and includes some tests to prove it's actually performing a proper 2-sample t-test.


interface/Gruntfile.js
```diff
@@ -96,6 +98,37 @@ module.exports = function ( grunt ) {
     ],
 
     /**
+     * Browserify the ttest library since it's written just for node
+     * Note: this must be run before the copy step (see `build` below).
+     */
+    browserify: {
+      dist: {
+        src: 'node_modules/ttest/hypothesis.js',
+        dest: 'node_modules/ttest/ttest.js',
```
dongbohu: (Forgive me if I sound too sensitive.) Do you think we should choose a more specific name for the destination file? (Maybe something like `ttest_greenelab.js`). I am a little worried that the author of this package might create his own *ttest.js* file in the future for whatever reason.

interface/src/app/greenelab/stats.spec.js
```diff
@@ -40,3 +40,83 @@ describe('Test the hypergeometric test function in stats.js', function() {
...
+    var tt;
+
+    // Compare with R's output from:
+    // tt <- t.test(c(0.7, -1.6, -0.2, -1.2, -0.1, 3.4, 3.7, 0.8, 0.0, 2.0),
+    // c(1.9, 0.8, 1.1, 0.1, -0.1, 4.4, 5.5, 1.6, 4.6, 3.4))
+    tt = MathFuncts.tTest(
+      [0.7, -1.6, -0.2, -1.2, -0.1, 3.4, 3.7, 0.8, 0.0, 2.0],
+      [1.9, 0.8, 1.1, 0.1, -0.1, 4.4, 5.5, 1.6, 4.6, 3.4]
+    );
+    expect(+tt.testValue().toFixed(10)).toEqual(-1.8608134675);
```
dongbohu: This line can be also written like this:
```
expect(+tt.testValue()).toBeCloseTo(-1.8608134675, 10); 
```
in which *10* is the maximum number of decimal points to determine whether two numerical values are close to each other or not.

(This is my latest discovery in jasmine documentation. Just feel that it is something good to know.)

interface/Gruntfile.js
mhuyck: That is a really good thought. Thank you for bringing it up, because I didn't think of it!

How about `greenelab.stats.ttest.js`? That sort of mimics the Angular module hierarchy and helps make the connection between this source file and where it ends up.

interface/src/app/greenelab/stats.spec.js
mhuyck: Cool! I will do it that way. @rzelayafavila, should I update the `expect`s from your hypergeometric test's test while I'm in there?

interface/Gruntfile.js
mhuyck: - [ ] rename the build product from `ttest.js` to `greenelab.stats.ttest.js` to avoid any future name conflicts.

interface/src/app/greenelab/stats.spec.js
mhuyck: - [ ] change `.toFixed(10)...toEqual` to use `.toBeCloseTo` instead

rzelayafavila: Oh wow sure, thanks for doing that @mhuyck and Im glad you found that different way @dongbohu!

PR #74, opened 2017-02-15
Sample Annotation Values Page

dongbohu:
This PR includes a new module to show sample annotation values based on sample IDs and an optional node ID. I also created a new module in greenelab/utils.js to take care of the base API path and the error message. I have written the same error handling code so many times that eventually I decide to improve the process.

To make the retrieval of sample annotation values faster, I modified api.py a little bit to enable the *in* filter.


dongbohu:
This is a demo page:
http://165.123.67.202/#/sample_annotation?node=10945&samples=10,11,12,16 (with both node and sample IDs in URL)
http://165.123.67.202/#/sample_annotation?samples=10,11,12,16 (only sample IDs in URL)

mhuyck:
As usual, codeclimate gripes about Angular's dependency declarations... safe to ignore.

mhuyck:
Thank you for the demo page! I'm not sure it works well to have the description appear as a column like everything else. Is there a way to select what columns to view (exclude the description, for example) or treat cells with lots of text differently?

dongbohu:
@mhuyck: Customization of columns is not implemented yet in this module. @cgreene and I discussed this before we decided to do it after February after talking to more users.

interface/src/app/sample_annotation/annotation.js
```diff
@@ -0,0 +1,117 @@
...
+    var getSampleIDPromise = function() {
+      var deferred = $q.defer();
+      // If 'node' is specified in the URL, query the node first:
+      if ($stateParams.node) {
+        self.hasNode = true;
+        var nodeID = $stateParams.node;
+        var activityApiConfig = {
+          params: {'node': nodeID, 'sample__in': samplesInUrl}
+        };
+        $http.get(ApiBasePath + 'activity/', activityApiConfig)
```
mhuyck: Rather than making direct `$http` calls from the controller, please arrange your calls to the back end via a more abstract service layer.

interface/src/app/greenelab/utils.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * This module includes a few utility components that will be shared in other
+ * modules to avoid duplicate codes.
+ */
+
+angular.module('greenelab.utils', [])
+.constant('ApiBasePath', '/api/v0/')
```
mhuyck: It does not make sense to me that `greenelab/utils/js` should have any information about the API. Is this needed for the `errGen` service somehow?

interface/src/app/greenelab/utils.js
```diff
@@ -0,0 +1,16 @@
...
+
+// A service function that returns a string of error message based on input
+// context and response.
+.value('errGen', function(context, response) {
```
mhuyck: I have not seen this construct before... so this is basically an Angular service that takes some parameters and returns a snippet of text?

interface/src/app/sample_annotation/annotation.tpl.html
```diff
@@ -0,0 +1,27 @@
...
+    <th ng-if="ctrl.hasNode">Activity</th>
+    <th>Sample Name</th>
+    <th ng-repeat="typeName in ctrl.uniqueAnnotationTypes">
+      {{typeName}}
+    </th>
+  </thead>
+
+  <tbody>
+    <tr ng-repeat=
+        "sample in ctrl.samples | orderBy: ctrl.hasNode ? '-activity' : 'name'">
```
mhuyck: Nice. I like how this view "responds" to the presence or absence of Node data in a smart way.

interface/src/app/sample_annotation/annotation.tpl.html
```diff
@@ -0,0 +1,27 @@
...
+      <td ng-if="ctrl.hasNode">
+        <strong>{{sample.activity | number: 5}}</strong>
```
mhuyck: Clever use of Angular's formatting pipes! Can we make the "magic number" 5 into a constant that can be set somewhere?

interface/src/app/sample_annotation/annotation.js
```diff
@@ -0,0 +1,117 @@
...
+    self.uniqueAnnotationTypes = [];
+    // Note: Ideally self.uniqueAnnotationTypes should be a hashtable (such as
+    // an object of Set) instead of Array, but Set is only available since ES6.
+
+    self.hasNode = false;
+    var sampleDict = {};
+
+    // Function that sets valid sample IDs asynchronously:
+    var getSampleIDPromise = function() {
+      var deferred = $q.defer();
```
mhuyck: I love this use of promises. I will *defer* (pun intended) looking into this portion of the code a little, though, because I want to see it after the `$http` calls are swapped out for a more loosely-couple service.

interface/src/app/greenelab/utils.js
dongbohu: This is actually a constant of API path. Suppose we decide to change our backend path to `/api/v1/` in the future, without using this constant in every module, we will have to change many places in the codebase. But if we use this constant in every module that calls the API, then we only need to change this constant.

(I learned this trick from an online Angular course at coursera.)

interface/src/app/greenelab/utils.js
dongbohu: `value()` is a shortcut version of `factory()` method. If a service doesn't have any dependencies, then you can define the value to be any string, object or function. In this case, `errGen` is a function to compile error message based on input parameters.

interface/src/app/sample_annotation/annotation.tpl.html
dongbohu: Sure. I will do that.

interface/src/app/greenelab/utils.js
mhuyck: I get that it's a constant, but not why it belongs in this file. My point is that `greenelab/utils.js` should not have a constant that is specific to accessing adage-server API endpoints.

It may very well be handy to have such a constant, but it should go somewhere else.

interface/src/app/sample_annotation/annotation.js
dongbohu: As we discussed in Rene's last PR, I can definitely see the advantages of abstracting the $http into a separate service; But on the other hand, I am hesitant about using it everywhere possible. I personally feel that having a service as simple as:
```
myMod.service('foobar', ['$resource', function($resource) {
  return $resource(url);
}]);
```
is an overkill. Instead, in this program, I tried to convert the whole function `getSampleIDPromise` into a service, but because it involves the error handling, I eventually gave up. 

What do you think @rzelayafavila and @cgreene?

interface/src/app/greenelab/utils.js
cgreene: I think I understand @mhuyck's concern here. We have something called `greenelab/util.js` that may get abstracted and used in other servers in the future. Because this is specific to the ADAGE server, what we want is something like `adage/const.js`. Is this roughly it @mhuyck?

dongbohu: Where do you think this module (or just this constant) could be?

adage/analyze/api.py
```diff
@@ -143,6 +143,7 @@ class Meta:
         experiments_allowed_methods = allowed_methods
         annotations_allowed_methods = allowed_methods
         filtering = {
+            'id': ('in',),
```
rzelayafavila: :+1:

interface/src/app/greenelab/utils.js
rzelayafavila: Hmmm - I agree with this - ``greenelab.utils.js`` will probably get installed onto other servers of ours, so I agree that we should have this as a constant, but totally, maybe in a app/constants.js file?

interface/src/app/greenelab/utils.js
rzelayafavila: Nice! this is cool!

rzelayafavila: Also, just curious - what could be an example of something that ``context`` would be?

dongbohu: It is just a string that indicates the scenario. You can see an example in `annotation.js`.

rzelayafavila: Nice!

rzelayafavila:
Ok, I made a first pass through the PR and just made a couple of comments, but I'll look at it again after the refactoring that we were talking about. I like the way there are functions to handle common events, and also the logic to build the sample annotation table - great work @dongbohu!

interface/src/app/greenelab/utils.js
dongbohu: It is not proper to call it `consts.js`, because in this module, one service is const and the other is a value. What about `app/utils.js`?

dongbohu: If you guys like `constants.js`, I can change `errGen` service from a **value** to **constant**. It means that later we can not use decorator to customize `errGen` service.

cgreene: Sure - sorry - don't know the angular way in this regard :)

dongbohu:
I think I have addressed all issues you guys pointed out. Please review it again and let me know. Thanks.

interface/src/app/sample_annotation/annotation.js
```diff
@@ -0,0 +1,129 @@
...
+      // If 'node' is found in the URL, get the activity values based on node
+      // and sample(s):
+      if ($stateParams.node) {
+        self.hasNode = true;
+        self.precision = 5;
```
rzelayafavila: 👍

interface/src/app/sample_annotation/annotation.js
```diff
@@ -0,0 +1,129 @@
...
+            var errMessage = errGen('Failed to get activities', response);
+            $log.error(errMessage);
+            self.queryStatus = errMessage + '. Please try again later.';
+            deferred.reject(errMessage);
+          });
+      } else { // If 'node' is not in the URL, include all input sample IDs.
+        deferred.resolve(samplesInUrl);
+      }
+      return deferred.promise;
+    };
```
mhuyck: Note: I restructured the above code but wasn't able to test the changes, so there may be a minor bug or two.

interface/src/app/sample_annotation/annotation.js
```diff
@@ -0,0 +1,129 @@
...
+            self.samples.push(sampleDict[key]);
+          });
+          self.uniqueAnnotationTypes.sort(); // Sort the annotation types
+          self.queryStatus = '';
+        }, function error(response) {
+          var errMessage = errGen('Failed to get sample annotations', response);
+          $log.error(errMessage);
+          self.queryStatus = errMessage + '. Please try again later.';
+        });
+    }); // end of sidPromise.then()
```
mhuyck: Another way to clarify the flow of deferred promises is to do the opposite of what I did above and add *more* functions, making them truly reusable, and then just calling them in series. This would highlight the "do this, then that" aspects more clearly, but requires defining each step in a functional way with inputs and return values crisply defined.

interface/src/app/sample_annotation/annotation.tpl.html
mhuyck: see above note about changing from `ctrl.hasNode` to `ctrl.nodeInUrl`, if you make use of the block of code I rewrote.

interface/src/app/greenelab/utils.js
mhuyck: I think the reason for the confusion here is that we have two distinct entities: `ApiBasePath` and `errGen`. Both of them are useful to have at a globally reusable level, but I think they each belong in different files.

`ApiBasePath` is indeed a constant, and it could go into a new `app/constants.js` file as suggested above. For `errGen`, I like `app/utils.js`. Although each file will be small to start, we can use these new "homes" for similar entities that will get created down the road.

I'm also happy with just keeping them both in `app/utils.js` for now and splitting out the constants when that file grows.

interface/src/app/sample_annotation/annotation.js
dongbohu: The main reason this variable is defined here is because it is actually the precision of activity values. So it is only needed when `node` parameter is in the URL. Having it defined here will save a few bytes of memory when this parameter is not in the URL.

interface/src/app/sample_annotation/annotation.js
dongbohu: I am not really convinced that the version you wrote is more readable. I actually still prefer having a separate getSampleIDPromise() function to make the code more modular.

interface/src/app/sample_annotation/annotation.js
dongbohu: I am going to rename this property to `self.activityPrecision` to make it more clear.

interface/src/app/sample_annotation/annotation.js
```diff
@@ -0,0 +1,129 @@
...
+  function($resource, ApiBasePath) {
+    return $resource(ApiBasePath + 'activity');
+  }
+])
+
+.factory('Sample', ['$resource', 'ApiBasePath',
+  function($resource, ApiBasePath) {
+    return $resource(ApiBasePath + 'sample');
+  }
+])
```
dongbohu: In the long run, we may need to write a `resource.js` where all API resources will be defined. Then other app modules will select whatever resource(s) needed as dependencies.

interface/src/app/sample_annotation/annotation.js
dongbohu: True. I am still not used to so many parentheses/braces/brackets/semicolons at the end of a JavaScript statement.

Issue #76, opened 2017-02-17
Reorganize modules into a more rational source tree


mhuyck:
The code base is growing and some of the organization that made sense at the beginning of the project is now hiding components that would otherwise be useful elsewhere.

In particular, recent work to show `Sample`s with their annotations was not able to reuse the original set of `Sample` modules because they live in `src/analyze/sample` rather than a more reusable `src/sample`. Both `src/analyze/sample` and `src/analyze/experiment` should be generalized and factored out. This work should focus on a separation of concerns between `sample` and `experiment` as *components* and `analyze`, etc. as *view*s or *layouts*.

This issue should stay open to collect observations about other areas that would benefit from refactoring to reflect the current organization of the site.

mhuyck:
Just noticed: `src/app/greenelab` files actually belong in `src/common/greenelab` since they are reusable outside of `adage-server`.

mhuyck:
To clarify what is an `adage-server` reusable component *vs.* view with specific layout, it would probably be good to consolidate the components into a single directory. Something like `src/app/components`. Then, everything outside of `components` within `src/app` should be organized exactly as it is within the navbar. This would make navigating the hierarchy of source code parallel navigation on the site itself.

mhuyck:
The implementation of `src/app/sample_annotation` largely rehashes significant parts of `src/app/analyze/sample` and `src/app/analyze/analysis/sampleBin.js` with the addition of node activity information as a new column. These efforts should be consolidated around a modular view of samples, their annotations and activity within a node.

mhuyck:
Part of this reorganization should involve fixing controllers that make direct calls to the REST API rather than invoking a service. Noted in #77 (see, for example, `node/node.js`, line 282):

```javascript
$http.get(sampleURI, {params: {experiment: exp.accession}})
```

mhuyck:
When refactoring `getUri` as @dongbohu suggested in #78, like so:

```javascript
.factory('getPromise', ['$resource', 
  function($resource) {
    var fn =  function(uri) {
      return $resource(uri).get().$promise;
    };
    return fn;
  }
]);
```

I noticed that unit tests for `$resource` are once again stripping trailing slashes. This is because our fix for that is now included only in `app.js`. That's really the wrong place. Consider moving this block from `app.js` into `utils.js`:

```javascript
// This configuration is required for all REST calls to the back end.
.config(['$resourceProvider', function($resourceProvider) {
[...]

PR #77, opened 2017-02-20
Allow users to select samples and point to corresponding sample annotations page

dongbohu:
Here is a demo page:
http://165.123.67.202/#/node/10945

The two red rectangles in this screenshot are the new features: 
demo

### DIFF
--- a/interface/src/app/node/high_range_exp.tpl.html
+++ b/interface/src/app/node/high_range_exp.tpl.html
@@ -7,7 +7,7 @@
   <button class="btn btn-primary" ng-click="setMode()"
           ng-show="topMode">Show All</button>
 
-  <br />
+  <br>
   (Click
   <i class="fa fa-plus-square fa-lg" aria-hidden="true"></i> or
   <i class="fa fa-minus-square fa-lg" aria-hidden="true"></i>
@@ -30,15 +30,43 @@
           <!-- Location of vega-lite widget -->
           <div id="{{ 'exp-' + $index }}"></div>
 
-          <ol ng-if="exp.isExpanded">
-            <li ng-if="!exp.sampleStatus"
-                ng-repeat="sample in exp.samples | orderBy: '-activity'">
-              Activity Value: <b>{{sample.activity.toFixed(5)}}</b><br/>
-              Sample Name: <b>{{sample.name}}</b><br/>
-              Data Source: <b>{{sample.ml_data_source}}</b>
-            </li>
+          <div ng-if="exp.isExpanded">
+            <div ng-if="!exp.sampleStatus">
+              <table class="table table-bordered table-condensed table-hover">
+                <thead>
+                  <tr>
+                    <th>#</th>
+                    <th>Activity</th>
+                    <th>Sample Name</th>
+                    <th>Data Source</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  <tr ng-repeat="sample in exp.samples | orderBy: '-activity'"
+                      ng-class="{success: sample.selected}">
+                    <td>
+                      <label>
+                        <input type="checkbox" ng-model="sample.selected"
+                               ng-change="toggleSelection(exp, sample.selected)">
+                        {{$index + 1}}
+                      </label>
+                    </td>
+                    <td>{{sample.activity.toFixed(5)}}</td>
+                    <td>{{sample.name}}</td>
+                    <td>{{sample.ml_data_source}}</td>
+                  </tr>
+                </tbody>
+              </table>
+              <div style="text-align: center">
+                <button class="btn btn-md btn-success"
+                        ng-disabled="exp.numSelections === 0"
+                        ng-click="showAnnotations(exp)">
+                  Show Annotations of Selected Samples
+                </button>
+              </div>
+            </div>
             <div ng-bind="exp.sampleStatus"></div>
-          </ol>
+          </div>
         </td>
       </tr>
     </tbody>
--- a/interface/src/app/node/node.js
+++ b/interface/src/app/node/node.js
@@ -99,7 +99,8 @@ angular.module('adage.node', [
   }]
 )
 
-.directive('highRangeExp', ['$http', '$log', function($http, $log) {
+.directive('highRangeExp', ['$http', '$log', '$state',
+function($http, $log, $state) {
   return {
     templateUrl: 'node/high_range_exp.tpl.html',
     restrict: 'E',
@@ -281,11 +282,13 @@ angular.module('adage.node', [
           $http.get(sampleURI, {params: {experiment: exp.accession}})
             .then(function success(response) {
               exp.samples = [];
+              exp.numSelections = 0;
               // Add "activity" property to each sample that is related to the
               // current node. (It will be used to order samples on web UI.)
               response.data.objects.forEach(function(element) {
                 if (element.id in $scope.activities) {
                   element.activity = $scope.activities[element.id];
+                  element.selected = false; // default: sample not selected.
                   exp.samples.push(element);
                 }
               });
@@ -303,6 +306,32 @@ angular.module('adage.node', [
         $scope.topMode = !$scope.topMode;
         $scope.numExpShown =
           $scope.topMode ? $scope.topNum : $scope.experiments.length;
+      };
+
+      // Event handler when user clicks the checkbox in front of each sample.
+      // It keeps track of the number of selected samples in each experiment
+      // to indicate whether the "show annotations" button should be enabled.
+      $scope.toggleSelection = function(exp, selected) {
+        if (selected) {
+          ++exp.numSelections;
+        } else {
+          --exp.numSelections;
+        }
+      };
+
+      // Event handler when user clicks the "show annotations" button:
+      $scope.showAnnotations = function(exp) {
+        var selectedSampleIDs = [];
+        exp.samples.forEach(function(sample) {
+          if (sample.selected) {
+            selectedSampleIDs.push(sample.id);
+          }
+        });
+        var stateParams = {
+          node: $scope.nodeId,
+          samples: selectedSampleIDs.join()
+        };
+        $state.go('sampleAnnotation', stateParams);
       };
     }
   };


dongbohu:
The codeclimate issue is due to indentation on one line. I did it on purpose to avoid changing all lines in `highRangeExp` directive. I will fix the indentation and file another PR after this one is done.

interface/src/app/node/high_range_exp.tpl.html
```diff
@@ -7,7 +7,7 @@
   <button class="btn btn-primary" ng-click="setMode()"
           ng-show="topMode">Show All</button>
 
-  <br />
+  <br>
```
mhuyck: Nice catch. We are serving `text/html`, not `application/xhtml+xml`, so this change is good.

interface/src/app/node/node.js
```diff
@@ -281,11 +282,13 @@ angular.module('adage.node', [
           $http.get(sampleURI, {params: {experiment: exp.accession}})
```
mhuyck: We should discuss converting this to a service. Not for this PR, though, so I will flag this as a note in the issue I already created about restructuring the code.

interface/src/app/node/high_range_exp.tpl.html
```diff
@@ -30,15 +30,43 @@
...
+                  <tr>
+                    <th>#</th>
+                    <th>Activity</th>
+                    <th>Sample Name</th>
+                    <th>Data Source</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  <tr ng-repeat="sample in exp.samples | orderBy: '-activity'"
+                      ng-class="{success: sample.selected}">
```
mhuyck: Where is the class `success` defined? Is this buried somewhere I haven't seen in Bootstrap? Mostly wondering how you came to use this for naming the "selected" state because I thought Bootstrap used the name "selected" for this.

interface/src/app/node/high_range_exp.tpl.html
```diff
@@ -30,15 +30,43 @@
...
+                    <td>
+                      <label>
+                        <input type="checkbox" ng-model="sample.selected"
+                               ng-change="toggleSelection(exp, sample.selected)">
+                        {{$index + 1}}
+                      </label>
+                    </td>
+                    <td>{{sample.activity.toFixed(5)}}</td>
```
mhuyck: Thank you for using fixed precision! Please define a "constant" to replace the 5. I think we have one already somewhere, but maybe not in this view.

interface/src/app/node/high_range_exp.tpl.html
```diff
@@ -30,15 +30,43 @@
...
+                    <td>{{sample.name}}</td>
+                    <td>{{sample.ml_data_source}}</td>
+                  </tr>
+                </tbody>
+              </table>
+              <div style="text-align: center">
+                <button class="btn btn-md btn-success"
+                        ng-disabled="exp.numSelections === 0"
```
mhuyck: Nicely done.

interface/src/app/node/node.js
```diff
@@ -304,6 +307,32 @@ angular.module('adage.node', [
...
+      // Event handler when user clicks the checkbox in front of each sample.
+      // It keeps track of the number of selected samples in each experiment
+      // to indicate whether the "show annotations" button should be enabled.
+      $scope.toggleSelection = function(exp, selected) {
+        if (selected) {
+          ++exp.numSelections;
+        } else {
+          --exp.numSelections;
+        }
+      };
```
mhuyck: My intuition says there must be an easier way to do this than "reference counting," but I can't think of how right now. This is written pretty clearly, but let's keep an eye out for a way to do this that doesn't involve a separate state tracking variable.

interface/src/app/node/high_range_exp.tpl.html
rzelayafavila: Agree :+1:

interface/src/app/node/high_range_exp.tpl.html
rzelayafavila: I think the ``success`` class is defined in Bootstrap, no? I think it is what makes buttons, table rows, alerts, etc. *green* (whereas ``primary`` makes them *blue*, etc.) - is this what you were referring to?

interface/src/app/node/high_range_exp.tpl.html
rzelayafavila: Agree about the magic number 5. Also, how small is this decimal number? I know that for some really small decimal numbers (such as P values, which can have many zeroes between the decimal point and where the significant digits begin), it can be useful to use the ``toPrecision()`` method, since it will show the number in scientific notation if this is the case. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision

interface/src/app/node/high_range_exp.tpl.html
dongbohu: Yes, `success` is defined in bootstrap to specify color of rows (or cell) in a table:
https://www.w3schools.com/bootstrap/bootstrap_tables.asp
When a certain row is selected, its color will become "success", which is same as the button's color at the bottom.

interface/src/app/node/high_range_exp.tpl.html
dongbohu: I am going to create a new constant service called `activityDigits` (which will be set to 5) in `adage.utils` module, then refer to this constant in both the current directive and `sample_annotation/annotation.js`. I will create an issue for now and file a separate PR later.

We can use the same approach for the P values, if you agree. It wouldn't make sense to set a mandatory constant of precision for all floating point values in the whole app, because different variables may need different precision.

interface/src/app/node/node.js
dongbohu: I read angular doc but couldn't find a built-in flag for it. Let me know if you find it. Thanks.

interface/src/app/node/high_range_exp.tpl.html
dongbohu: By the way, the activity values are in the range of (-1.0, 1.0).

interface/src/app/node/node.js
rzelayafavila: On a slightly related topic - I think that this large ``highRangeExp`` directive can actually also be divided into smaller directives/components, where one is the directive/component for the Node page quadrant that holds all of the Highest Range Experiments, and it calls smaller directives/components for each experiment itself. That way, the experiment in each component only knows information about itself and the samples it contains. However, this would almost surely fall outside the scope of this Pull Request, but it is just something I was also thinking about.

interface/src/app/node/high_range_exp.tpl.html
rzelayafavila: Right, I agree that it should not be mandatory for all places where we have decimals.  What do you think about the toPrecision() method I mentioned though? The use case I was thinking of was when we have a decimal number, say +/- 0.000000392, where the toFixed() would just go up to the first 5 decimal places, and therefore convert it to 0.00000, right?

interface/src/app/node/node.js
dongbohu: True. I will address it sometime later. When I started to write this directive, I didn't expect that so many new features would be added to it.

dongbohu:
I think I answered all questions from the reviewers. I am going to address some of them in other PR(s), because they involve a lot more revisions.

interface/src/app/node/high_range_exp.tpl.html
dongbohu: For activity values, using `toFixed(5)` or `toPrecision(5)` doesn't make much difference, because these values behave reasonally well. For p values, I think `toPrecision()` should be preferred.

rzelayafavila: Ok, that sounds fine.

interface/src/app/node/node.js
rzelayafavila: Also, other than reference counting, we could possibly use the JavaScript array ``filter()`` function here, inside the ``$scope.toggleSelection()`` function. It could go something like:
```
var selectedSamples = exp.samples.filter(function(sample) {
  return (sample.selected === true);
}
if (selectedSamples.length > 0) {
  exp.selectedSamples = true;
} else {
  exp.selectedSamples = false;
}
```
or something along those lines. That is another possibility.

rzelayafavila:
I tested the demo page, and I do like the way the UI for this part looks and feels.

interface/src/app/node/node.js
dongbohu: I thought of this implementation. The disadvantage of it is that when an experiment has many samples, this filter may take much longer than simply incrementing or decrementing a single counter variable.

rzelayafavila: How many samples could an experiment have?

dongbohu: It depends, could be hundreds, probably.

dongbohu: By the way, is there any disadvantage of reference counting? Since it is in JavaScript, it is thread-safe.

dongbohu:
Glad you liked it.

interface/src/app/node/node.js
rzelayafavila: No, it would just make the code's intent easier to understand when someone else is reading it. I don't have strong feelings on it though, I am good with this implementation if you are good with it.

rzelayafavila:
Ok I think Im good with pretty much all the changes - @dongbohu, will you make an issue for the splitting up the highRangeExp issue we talked about so we don't forget about it? I think changes will be good on my end after that.

dongbohu:
@rzelayafavila: Sure. I will add a new issue about it.

interface/src/app/node/high_range_exp.tpl.html
mhuyck: Okay, I see that now. The intended semantics of the Bootstrap class is for things like alerts and such, where `Danger`, `Info`, and `Warning` (described at @dongbohu's link) would also make sense. Using it here is harmless since we already depend upon Bootstrap heavily, but we really ought to be defining our own semantic classes and using those instead. Not a big deal, but worth addressing along with the rest of Issue #76.

interface/src/app/node/node.js
mhuyck: @dongbohu I figured that performance was partly why you did it this way. I was completely surprised to find this kind of looping to be *very* fast when I was checking on the performance of the javascript in the heatmap. I fully expected a crazy loop I had written to be the source of some slow performance I observed and the code completely zipped right through a more complex loop over 600 nodes * 10 samples (so about 6000 entries) in a complex data structure.

dongbohu: Probably from a modern CPU's point of view, 6000 vs. 1 is not really a big deal. It will just crunch them before human being can sense the difference of its processing time.

mhuyck: Right. The thing to keep in mind is that memory is also a constraint (along with CPU) in the browser context. Although modern hardware is fairly capacious there, too, browsers need to juggle dozens of tabs and keep from getting bogged down by the JavaScript on *all* of those pages operating simultaneously. If the page needs to run on a mobile browser (not really a consideration for us yet), everything needs to be that much tighter.

PR #78, opened 2017-02-22
Patch sample conflict

mhuyck:
I still need to finalize the patch portion of this PR, but I'm filing it as a WIP so @dongbohu can see the tests I've added for `annotation.js`.

### DIFF
--- a/interface/karma/karma-unit.tpl.js
+++ b/interface/karma/karma-unit.tpl.js
@@ -35,6 +35,8 @@ module.exports = function ( karma ) {
      * How to report, by default.
      */
     reporters: 'dots',
+    // This option is handy for sorting out some kinds of testing errors
+    // logLevel: karma.LOG_DEBUG,
 
     /**
      * On which port should the browser connect, on which port is the test
--- a/interface/src/app/analyze/experiment/experiment.js
+++ b/interface/src/app/analyze/experiment/experiment.js
@@ -9,11 +9,7 @@ angular.module('adage.analyze.experiment', [
     '/api/v0/experiment/:accession/',
     // TODO need to add logic for handling pagination of results.
     // then, can change "limit" below to something sensible
-    {accession: '@accession', limit: 0},
-    // Angular expects a query service to give only a list but our Tastypie
-    // interface wraps the response list with pagination so we tell Angular to
-    // expect an object instead via isArray: false
-    {'get': {method: 'GET', isArray: false}}
+    {limit: 0}
   );
   Experiment.makeHref = {
     // These functions create formatted URLs for direct linking to source
--- a/interface/src/app/analyze/experiment/experiment.js
+++ b/interface/src/app/analyze/experiment/experiment.js
@@ -43,14 +43,14 @@ angular.module('adage.analyze.experiment', [
     $scope.show = function(id) {
       $scope.experiment = {
         status: 'retrieving...',
-        related_samples: []
+        relatedSamples: []
       };
       var getSampleDetails = function(uri) {
         Sample.getUri(uri).then(
           function(responseObject) {
             if (responseObject) {
               $scope.experiment.status = '';
-              $scope.experiment.related_samples.push(responseObject.data);
+              $scope.experiment.relatedSamples.push(responseObject.data);
             }
           },
           queryError
--- a/interface/src/app/sample/sampleMocks.spec.js
+++ b/interface/src/app/sample/sampleMocks.spec.js
@@ -1,97 +1,95 @@
 angular.module('adage.mocks.sample', [])
 
-.factory('SampleMocks', [function() {
-  return {
-    sample1: {
-      'annotations': {
-        'additional_notes': '9.5 h biofilms',
-        'biotic_int_lv_1': 'Human',
-        'biotic_int_lv_2': 'Lung epithelial cells (CFBE41o- cells)',
-        'description':
-          'Pseudomonas aeruginosa 9.5 hour static coculture with human blah…',
-        'genotype': 'WT',
-        'growth_setting_1': 'Biofilm',
-        'growth_setting_2': 'Static',
-        'medium': 'MEM, 0.4% arginine',
-        'nucleic_acid': 'RNA',
-        'od': '9.5 hours',
-        'strain': 'PA14',
-        'temperature': '37'
-      },
-      'id': 1,
-      'ml_data_source': 'GSM252496.CEL',
-      'name': 'GSE9989GSM252496',
-      'resource_uri': '/api/v0/sample/1/'
+.constant('SampleMocks', {
+  sample1: {
+    'annotations': {
+      'additional_notes': '9.5 h biofilms',
+      'biotic_int_lv_1': 'Human',
+      'biotic_int_lv_2': 'Lung epithelial cells (CFBE41o- cells)',
+      'description':
+        'Pseudomonas aeruginosa 9.5 hour static coculture with human blah…',
+      'genotype': 'WT',
+      'growth_setting_1': 'Biofilm',
+      'growth_setting_2': 'Static',
+      'medium': 'MEM, 0.4% arginine',
+      'nucleic_acid': 'RNA',
+      'od': '9.5 hours',
+      'strain': 'PA14',
+      'temperature': '37'
     },
-    sample1GetExperiments: [
-      {
-        'accession': 'E-GEOD-9989',
-        'description':
-          'We grew Pseudomonas aeruginosa biofilms on CFBE41o- bla…',
-        'name':
-          'Transcription profiling of P. aeruginosa biofilms treated blah…',
-        'resource_uri': '/api/v0/experiment/E-GEOD-9989/',
-        'sample_set': [
-          '/api/v0/sample/1/',
-          '/api/v0/sample/2/',
-          '/api/v0/sample/3/',
-          '/api/v0/sample/4/',
-          '/api/v0/sample/5/',
-          '/api/v0/sample/6/'
-        ]
-      },
-      {
-        'accession': 'E-GEOD-10030',
-        'description': 'This SuperSeries is composed of the following blah…',
-        'name':
-          'Transcription profiling of Pseudomonas aeruginosa treated blah…',
-        'resource_uri': '/api/v0/experiment/E-GEOD-10030/',
-        'sample_set': [
-          '/api/v0/sample/1/',
-          '/api/v0/sample/2/',
-          '/api/v0/sample/3/',
-          '/api/v0/sample/4/',
-          '/api/v0/sample/5/',
-          '/api/v0/sample/6/',
-          '/api/v0/sample/7/',
-          '/api/v0/sample/8/',
-          '/api/v0/sample/9/',
-          '/api/v0/sample/10/'
-        ]
-      }
-    ],
-    sample1Paginated: {
-      'meta': {
-        'limit': 1000,
-        'next': null,
-        'offset': 0,
-        'previous': null,
-        'total_count': 1
-      },
-      'objects': [
-        {
-          'annotations': {
-            'additional_notes': '9.5 h biofilms',
-            'biotic_int_lv_1': 'Human',
-            'biotic_int_lv_2': 'Lung epithelial cells (CFBE41o- cells)',
-            'description':
-              'Pseudomonas aeruginosa 9.5 hour static coculture with blah…',
-            'genotype': 'WT',
-            'growth_setting_1': 'Biofilm',
-            'growth_setting_2': 'Static',
-            'medium': 'MEM, 0.4% arginine',
-            'nucleic_acid': 'RNA',
-            'od': '9.5 hours',
-            'strain': 'PA14',
-            'temperature': '37'
-          },
-          'id': 1,
-          'ml_data_source': 'GSM252496.CEL',
-          'name': 'GSE9989GSM252496',
-          'resource_uri': '/api/v0/sample/1/'
-        }
+    'id': 1,
+    'ml_data_source': 'GSM252496.CEL',
+    'name': 'GSE9989GSM252496',
+    'resource_uri': '/api/v0/sample/1/'
+  },
+  sample1GetExperiments: [
+    {
+      'accession': 'E-GEOD-9989',
+      'description':
+        'We grew Pseudomonas aeruginosa biofilms on CFBE41o- bla…',
+      'name':
+        'Transcription profiling of P. aeruginosa biofilms treated blah…',
+      'resource_uri': '/api/v0/experiment/E-GEOD-9989/',
+      'sample_set': [
+        '/api/v0/sample/1/',
+        '/api/v0/sample/2/',
+        '/api/v0/sample/3/',
+        '/api/v0/sample/4/',
+        '/api/v0/sample/5/',
+        '/api/v0/sample/6/'
+      ]
+    },
+    {
+      'accession': 'E-GEOD-10030',
+      'description': 'This SuperSeries is composed of the following blah…',
+      'name':
+        'Transcription profiling of Pseudomonas aeruginosa treated blah…',
+      'resource_uri': '/api/v0/experiment/E-GEOD-10030/',
+      'sample_set': [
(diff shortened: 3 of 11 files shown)


mhuyck:
Note: some of the code duplication picked up by codeclimate will be resolved by the patch I'm writing. The "sample conflict" is, in fact, due to a pair of duplicate services that conflict with each other.

dongbohu:
Is it ready for review now? If it is, can you remove "WIP" from the title of this PR?

mhuyck:
I thought it was clear in the description above that "I still need to finalize the patch portion." Feedback is welcome any time so you are free to add review comments whenever you like, but I will remove the WIP when the PR is complete.

dongbohu:
I didn't realize you have written so many tests for `sample_annotation/annotation.js`. I am not going to work on mine, because many of mine are duplicate test cases.

mhuyck:
If you have other tasks that take precedence, that sounds good. As I indicated this morning at scrum, the tests I've written are focused solely on features that interact with the `Sample` service, since that is what I'm refactoring to resolve the conflict.

We will ultimately need tests for `annotation.js` that check the `Activity` service and data that are requested when the page has a `node=` parameter supplied. We did say that unit tests are mainly being deferred until after next Tuesday, though, so if you want to wait on that it makes sense to me.

mhuyck:
Well, that didn't quite make codeclimate as happy as I had hoped, but there's not much more I can do for now to remove redundancy. Testing just involves a bunch of boilerplate that's hard to factor out.

Look as if CI is happy so this is now ready for review. Removing the "WIP" label...

interface/src/app/sample/sampleMocks.spec.js
```diff
@@ -0,0 +1,97 @@
+angular.module('adage.mocks.sample', [])
```
dongbohu: So this file defines a module that is for test only? Since it doesn't include any test cases, it probably shouldn't be named "*.spec.js". 

I noticed that this module is needed by `sample_annotation/annotation.spec.js`. This makes me a little uneasy. I would prefer having test cases to be in separate units without this kind of extra dependencies. and have the tests in `sample_annotation/annotation.spec.js` provide fake data directly.

interface/src/app/sample/sampleMocks.spec.js
```diff
@@ -0,0 +1,97 @@
...
+
+.factory('SampleMocks', [function() {
```
dongbohu: By the way, since this service just returns an object literal, you can write like this:
```
.value('sampleMocks', { ... });
```
or:
```
.constant('sampleMocks', { ... });
```

interface/src/app/sample/services.js
```diff
@@ -0,0 +1,37 @@
...
+  'adage.utils'
+])
+
+.factory('Sample', ['$resource', '$http', 'ApiBasePath',
+  function($resource, $http, ApiBasePath) {
+    var Sample = $resource(
+      ApiBasePath + 'sample/:id/',
+      // TODO need to add logic for handling pagination of results.
+      // then, can change "limit" below to something sensible
+      {id: '@id', limit: 0},
```
dongbohu: `@id` means `id` will be from input data (which is an argument for non_GET method). I don't think we need it here.

interface/src/app/sample/services.js
```diff
@@ -0,0 +1,37 @@
...
+    );
+    Sample.getUri = function(uri) {
+      // We can't easily get Tastypie-supplied resource_uris via $resource, so
+      // this uses $http directly and returns the HttpPromise
+      return $http({url: uri, method: 'GET'});
+    };
+    return Sample;
+  }
+])
+
```
dongbohu: I wonder whether the definition of this service can be simplified into:
```
.factory('Sample', ['$resource', 'ApiBasePath',
  function($resource, ApiBasePath) {
    var Sample = $resource(
      ApiBasePath + 'sample/:id/',
      {limit: 0},
      {
        'getExperiments': {
          url: ApiBasePath + 'sample/:id/get_experiments/',
          method: 'GET',
          isArray: true
        }
      }
    );
    Sample.getUri = function(uri) {
      return $resource(uri).get().$promise;
    };
    return Sample;
  }
])
```
(I removed the comments to make the code easier to read. Of course you can keep them.)

In getUri() method, `$http` dependency is replaced by `$promise` property. I also feel that the method name `getUri` is misleading, since it actually gets a promise.

By the way, getUri() method seems to be unrelated to Sample at all. Why not create another service like this:
```
.factory('getPromise', ['$resource', 
  function($resource) {
    var fn =  function(uri) {
      return $resource(uri).get().$promise;
    };
    return fn;
  }
]);
```

interface/src/app/sample/services.spec.js
```diff
@@ -0,0 +1,63 @@
+describe('adage.sample.services', function() {
+  var $httpBackend, $log, Sample, mockSampleData;
+
+  beforeEach(module('adage.sample.services'));
+  beforeEach(module('adage.mocks.sample'));
```
dongbohu: If you want, lines 4-5 can be combined into one:
```beforeEach(module('adage.sample.services', 'adage.mocks.sample');```

File: interface/src/app/sample_annotation/annotation.js [hunk 6]
```diff
@@ -5,7 +5,9 @@
 angular.module('adage.sampleAnnotation', [
   'ui.router',
   'ui.bootstrap',
-  'adage.utils'
+  'ngResource',
```
dongbohu: Thanks. Yesterday I also found that 'ngResource' was missed.

interface/src/app/sample_annotation/annotation.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    $httpBackend.verifyNoOutstandingRequest();
+    // very useful $log capture information from:
+    // http://www.jvandemo.com/how-to-access-angular-log-debug-messages-from-within-karma/
+    if ($log.debug.logs.length > 0) {
+      console.log($log.debug.logs);
+    }
+  });
+
+  describe('SampleAnnotationCtrl', function() {
+    var SampleAnnotationCtrl, $location, $scope, mockStateParams, $log;
```
dongbohu: Actually here `$location`, `$scope`, `mockStateParams` and `$log` are not necessary, unless you want to do more in the  first test case ('should pass a dummy test'). This line in the dummy test:
```
SampleAnnotationCtrl = $controller('SampleAnnotationCtrl');
```
will automatically figure out the dependencies of `SampleAnnotationCtrl` and inject them in. (Besides, `$log` has  already been defined in the parent `describe` block.) So unless you need to use these dependencies directly (like the call of $log in afterEach block above), you don't need them.

And the beforeEach() on lines 26-39 can be simplified into:
```
beforeEach(inject(function($controller) {
  SampleAnnotationCtrl = $controller('SampleAnnotationCtrl');
}));
```

interface/src/app/sample_annotation/annotation.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    it('should pass a dummy test', inject(function() {
+      $httpBackend.expectGET(
+        '/api/v0/sample?id__in=1&limit=0'
+      ).respond(SampleMocks.sample1Paginated);
+      $httpBackend.flush();
+
+      expect(SampleAnnotationCtrl).toBeTruthy();
+    }));
+
+    it('should render the SampleAnnotationCtrl for a sample', inject(
```
dongbohu: Since the injected function doesn't have any dependencies, this test case can be simplified into:
```
it('should render the SampleAnnotationCtrl for a sample', 
  function() {
    ...
  }
);

File: interface/src/app/sample/sampleMocks.spec.js (hunk 1 above)
Comment (mhuyck) [inline interface/src/app/sample/sampleMocks.spec.js]: The files named `*.spec.js` are designated specifically for unit tests, yes. The mock data are designated specifically for the same tests, so I'm not sure why it would be a problem to name them this way.

From a practical perspective, I can't name them anything else without modifying the naming system used by the build system. We don't want mock data built and compiled into our deployed system, and that is what would happen if I named them `somethingelse.js` without the `spec.js` ending. Have a look at `build.config.js`, where the `app_files` are defined starting at line 22. Notice that the `*.spec.js` files are explicitly *excluded* from the `js` definition and *included* only in the `jaunt` definition. This is exactly what we want.

Thank you for pointing out the potential problem with dependencies. I've thought about that carefully as well. Now that we have refactored to remove the conflict with two `Sample` services, `sample_annotation/annotation.js` already depends upon having `sample/services.js` injected. I think it is reasonable for the unit tests to follow that same path of dependency, so it is logical that `annotation.spec.js` should depend upon some testing infrastructure that is also required by `sample/services.spec.js`. Remember that I'm doing this to uphold the DRY principle that `codeclimate` is helping to enforce. This structure helps keep the testing data for this service right along with the service itself, where it's bound to be needed again.

If this still doesn't feel right to you, please create an issue so we can discuss it further. We need this patch to get merged so it doesn't hold up the next round of pull requests.

File: interface/src/app/sample/sampleMocks.spec.js (hunk 2 above)
Comment (mhuyck) [inline interface/src/app/sample/sampleMocks.spec.js]: Good thinking. I will change it to a `.constant`.

- [ ] change `SampleMocks` from a `.factory` to a `.constant`

File: interface/src/app/sample/services.js (hunk 3 above)
Comment (mhuyck) [inline interface/src/app/sample/services.js]: Nice catch. This is one of the first services I wrote so I was probably following some sample code that I didn't fully understand. I will pull that part out and test.

- [ ] try simplifying `services.js` by removing `@id` from the `$resource` definition

File: interface/src/app/sample/services.js (hunk 4 above)
Comment (mhuyck) [inline interface/src/app/sample/services.js]: I will try the simplified code you wrote, thank you.

- [ ] try further simplification for `services.js`

That is a good point about `getUri` as well. I'd like to defer that change to go along with the broader refactoring I've documented in issue #76. Is it okay with you if I make a note about this there? (I think I just did, actually with the reference here.)

File: interface/src/app/sample/services.spec.js (hunk 5 above)
Comment (mhuyck) [inline interface/src/app/sample/services.spec.js]: Nice. It was handy to have these on separate lines as I refactored and tested the tests, but I like the combined call for the final product.

- [ ] combine `module` calls as @dongbohu suggests

File: interface/src/app/sample_annotation/annotation.spec.js (hunk 7 above)
Comment (mhuyck) [inline interface/src/app/sample_annotation/annotation.spec.js]: Good catch. I will clean this up a bit more. There's a bunch in there that's left over from the testing boilerplate that came with `ngBoilerplate` and is no longer used. I got a bunch out when ESLint found variables that were defined but not used, but these must have escaped its parser.

- [ ] removed unused testing artifacts

File: interface/src/app/sample_annotation/annotation.spec.js (hunk 8 above)
Comment (mhuyck) [inline interface/src/app/sample_annotation/annotation.spec.js]: Another good call. Thank you!

- [ ] strip empty injection call from this test

File: interface/src/app/sample/sampleMocks.spec.js (hunk 1 above)
Comment (mhuyck) [inline interface/src/app/sample/sampleMocks.spec.js]: Apparently autocorrect was thinking `jaunt` looked better than `jsunit`. Too bad that makes no sense.

File: interface/src/app/sample_annotation/annotation.spec.js (hunk 7 above)
Comment (mhuyck) [inline interface/src/app/sample_annotation/annotation.spec.js]: You are 100% correct about the `$location`, `$scope` and the extra redundant `$log`. I kept the `mockStateParams`, though, since that's how I'm specifying for the unit test what the `samples=` parameter would have been if this were invoked by `ui.router`. This is where we will need to also specify a `node` parameter when that test is ready to go.

**Comment by mhuyck** [comment]
Hmm. Not sure what happened with codeclimate. I was trying to make it happy!

**Comment by mhuyck** [comment]
Poor codeclimate. Perhaps there were too many diffs for it to handle?

At any rate, I think I've now either addressed everything mentioned by @dongbohu or deferred via an issue. Please let me know if you have any other questions!

File: interface/src/app/sample/sampleMocks.spec.js (hunk 1 above)
Comment (dongbohu) [inline interface/src/app/sample/sampleMocks.spec.js]: @mhuyck: Thank you for the clarification. The reason why I don't feel comfortable about the *.spec.js" is because in JavaScript test framework, `spec` is a synonym of "test cases", but this file includes a module instead of test case(s).

There are some interesting discussions online about DRY vs. DAMP when writing test code. For example:
http://stackoverflow.com/questions/6453235/what-does-damp-not-dry-mean-when-talking-about-unit-tests
I would rather have codeclimate exclude the spec files in order to keep the unit tests really UNIT. Considering the upcoming deadline, I am going to create an issue for further discussion in the future.

File: interface/src/app/sample/services.js (hunk 4 above)
Comment (dongbohu) [inline interface/src/app/sample/services.js]: I understand that this proposed revision may affect many other places in your code. If you want to close this PR as soon as possible, please feel free to create an issue for now.

File: interface/karma/karma-unit.tpl.js [hunk 9]
```diff
@@ -35,6 +35,8 @@ module.exports = function ( karma ) {
      * How to report, by default.
      */
     reporters: 'dots',
+    // This option is handy for sorting out some kinds of testing errors
+    // logLevel: karma.LOG_DEBUG,
```
rzelayafavila: Thanks for the helpful comments! :+1:

interface/src/app/analyze/experiment/experiment.js
```diff
@@ -9,11 +9,7 @@ angular.module('adage.analyze.experiment', [
     '/api/v0/experiment/:accession/',
     // TODO need to add logic for handling pagination of results.
     // then, can change "limit" below to something sensible
-    {accession: '@accession', limit: 0},
-    // Angular expects a query service to give only a list but our Tastypie
-    // interface wraps the response list with pagination so we tell Angular to
-    // expect an object instead via isArray: false
-    {'get': {method: 'GET', isArray: false}}
+    {limit: 0}
```
rzelayafavila: Nice simplification @mhuyck and @dongbohu! I guess this will still allow us to request an experiment by its accession number, right?

interface/src/app/sample/services.js
rzelayafavila: Ok - I'm totally on board with these changes and deferring them until a later point. I do like the idea of calling it something like ``getPromise()`` more than ``getUri()``, since ``getUri`` seems to me like a function that would return the ``resource_uri`` itself, rather than a promise. I think I would prefer even ``getViaUri()`` over ``getUri()``, but ``getPromise()`` sounds even better.

interface/src/app/analyze/experiment/experiment.js
dongbohu: Yes. For GET method, `@accession` doesn't do anything because it don't have input data argument.

interface/src/app/sample/services.js
mhuyck: Right. Just to be clear: I did actually try the change to `$resource` you suggested and got a couple of stack traces back from the tests so backed it out.

I completely agree with what you've both noticed about naming as well. If this function even survives reorganization (might not be necessary anymore with a good refactoring) it should definitely have a better name.

interface/karma/karma-unit.tpl.js
mhuyck: Karma like to be terse as it runs through the tests, most likely for efficiency. If the tests aren't working right, though, it can be maddening! This setting makes karma itself tell you more about what it's doing.

To get more debug info out of the tests themselves, I was able to use `console.log()` statements to good effect, along with explicitly reporting what the mock `$log` service captured as I've done in the `.spec.js` files.

interface/src/app/analyze/experiment/experiment.js
mhuyck: Right. I learned from @dongbohu here that the `@var` syntax does not apply for `GET` requests so it was doing nothing.

In addition, the `isArray: false` was something I needed for `query` calls (see the comment above) but does not apply to `get`, as we have here.

interface/src/app/sample/sampleMocks.spec.js
mhuyck: @dongbohu: it sounds as if you have a Vision for how we should be doing testing. I look forward to learning more as we follow up via issue #81.

PR #84, opened 2017-02-23
Indentation correction in node.js

dongbohu:
Although this seems to be a huge PR, it is actually trivial. The indentation of the second line of `highRangeExp` directive's definition has been corrected, which causes indentations of all the other lines below it to be adjusted too.

Another trivial change is the removal of trailing blank line in `about.tpl.html`.

### DIFF
--- a/interface/src/app/node/node.js
+++ b/interface/src/app/node/node.js
@@ -100,242 +100,248 @@ angular.module('adage.node', [
 )
 
 .directive('highRangeExp', ['$http', '$log', '$state',
-function($http, $log, $state) {
-  return {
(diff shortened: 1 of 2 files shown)


interface/src/app/node/node.js
```diff
@@ -100,242 +100,248 @@ angular.module('adage.node', [
...
+            samples: selectedSampleIDs.join()
+          };
+          $state.go('sampleAnnotation', stateParams);
         };
-        $state.go('sampleAnnotation', stateParams);
-      };
-    }
-  };
-}])
+      } // end of link function
```
rzelayafavila: Nice! This comment is helpful since there are so many nested curly braces!

PR #85, opened 2017-02-23
Adding step to README for pickling Tribe genesets.

rzelayafavila:


README.md
```diff
@@ -202,6 +202,16 @@ Documentation for using each command is available by typing:
 > python manage.py help <command>
 ```
 
+### Pickle Tribe gene sets
+
+This step downloads and pickles all the public Tribe gene sets for every
+organism that has been loaded into the database. These pickled files are
+necessary for doing gene set enrichment analyses.
```
mhuyck: I know you're adding it to the `tribe_client` documentation, but do you want to make a note here that this command depends upon some settings? Maybe a link directly to the `tribe_client` docs would be the right thing. I'll go review those changes now.

rzelayafavila: Well, I thought about that, but notice the instructions in line 83 that talk about having the settings for tribe-client, along with the link to tribe-client. I didn't know whether we should mention that again here.

rzelayafavila: Oh also @mhuyck one more question - since I bumped the tribe-client version number, I should probably update the requirements.txt for adage and include it here, right? Even though the only thing that will have changed in tribe-client is the README, but we should keep them sync'ed anyway?

mhuyck: Oh, gotcha. I forgot that other parts of this page have instructions on installing `tribe_client` already. Hmm... a short link wouldn't hurt though. Sometimes I search documentation and jump into the middle somewhere. In a case like that it's handy to have a cross reference right where I need it. Something short like "for more information, see the tribe_client documentation," with an embedded link, maybe? Up to you.

Bumping requirements.txt would be nice. Good idea.

dongbohu: I agree that we should keep the version number sync'ed n `requirements.txt`.

rzelayafavila: @mhuyck : Ok, yeah, I agree, that sounds good and will be a quick change. I will insert a small hyperlink.

rzelayafavila:
Ok @mhuyck and @dongbohu I think I made a couple more changes, let me know how these look!

dongbohu:
I approved it again.

rzelayafavila:
Ok thanks @dongbohu @mhuyck!

PR #86, opened 2017-02-24
New service of constant: ActivityDigits

dongbohu:
This PR includes:

* Definition of a new constant service `ActivityDigits`, which is the number of digits after decimal point in activity values;

* Usage of this service in `high-range-exp` directive on **Node** page;

* Usage of this service on sample annotation values module.

### DIFF
--- a/interface/src/app/node/high_range_exp.tpl.html
+++ b/interface/src/app/node/high_range_exp.tpl.html
@@ -51,7 +51,7 @@
                         {{$index + 1}}
                       </label>
                     </td>
-                    <td>{{sample.activity.toFixed(5)}}</td>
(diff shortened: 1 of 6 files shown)


dongbohu:
By the way, this PR is for issue #86.

mhuyck:
Nothing we can do about codeclimate's complaints here. (Aside: I have a gut feeling that the refactoring discussed in #76 would make for fewer redundant dependency lists and improve our score on that front.)

dongbohu:
Thanks @mhuyck!

dongbohu:
Just realized that the unit test doesn't need to provide the constant service explicitly. It will be found automatically by dependency injection.

PR #87, opened 2017-02-25
Volcano plot groups

mhuyck:
This set of changes allows users to put each sample into one of three groups, labeled in the code as "group a," "group b" and "other." A convenience method `getSampleGroups()` is also supplied, making it easier to retrieve the list of samples within each group.

No demo server running yet. I will post here when it's up.

### DIFF
--- a/interface/src/app/analyze/analysis/analysis.less
+++ b/interface/src/app/analyze/analysis/analysis.less
@@ -1,3 +1,9 @@
+.sample-group-a {
+  background-color: #b7baff;
+}
+.sample-group-b {
+  background-color: #fee7a9;
+}
 .as-sortable-item, .as-sortable-placeholder {
   display: table-row;
 }
--- a/interface/src/app/analyze/analysis/analysis.less
+++ b/interface/src/app/analyze/analysis/analysis.less
@@ -25,9 +25,7 @@ th .row-control, thead .row-control {
   visibility: hidden;
 }
 
-// sample group highlighting
-@sample-group-a: #b7baff;
-@sample-group-b: #fee7a9;
+// use special colors for highlighting sample groups (see less/variables.less)
 .table-row-variant(sample-group-a; @sample-group-a);
 .table-row-variant(sample-group-b; @sample-group-b);
 
--- a/interface/src/app/analyze/analysis/analysis.less
+++ b/interface/src/app/analyze/analysis/analysis.less
@@ -1,9 +1,37 @@
-.sample-group-a {
-  background-color: #b7baff;
+/*
+ * styles to make sampleBin editing controls more user-friendly
+ */
+.row-control {
+  float: none;
+  .opacity(.2);
+
+  &:hover,
+  &:focus {
+    text-decoration: none;
+    cursor: pointer;
+    .opacity(.5);
+  }
+
+  button& {
+    padding: 0;
+    cursor: pointer;
+    background: transparent;
+    border: 0;
+    -webkit-appearance: none;
+  }
 }
-.sample-group-b {
-  background-color: #fee7a9;
+// allow us to use controls to help align headers, but don't show them
+th .row-control, thead .row-control {
+  visibility: hidden;
 }
+
+// sample group highlighting
+@sample-group-a: #b7baff;
+@sample-group-b: #fee7a9;
+.table-row-variant(sample-group-a; @sample-group-a);
+.table-row-variant(sample-group-b; @sample-group-b);
+
+// helper styles for as-sortable control
 .as-sortable-item, .as-sortable-placeholder {
   display: table-row;
 }
--- a/interface/src/app/analyze/analysis/analysisDetail.tpl.html
+++ b/interface/src/app/analyze/analysis/analysisDetail.tpl.html
@@ -1,27 +1,35 @@
 <!--
   TODO each sample should be built using a reusable Angular directive
 -->
-<h1>Samples <small>(drag to reorder)</small></h1>
+<h1>Samples
+  <small>(drag <i class="fa fa-bars" aria-hidden="true"></i> to reorder,
+    click <i class="fa fa-times" aria-hidden="true"></i> to delete,
+    choose <input type="radio" disabled> to set group)</small>
+</h1>
 <status-bar msg-text="analysis.status"></status-bar>
 <!-- TODO need a tool to select and order annotation columns -->
 <table class="table table-hover">
   <thead>
     <tr>
-      <td>
-        group
-      </td>
-      <td>id</td>
-      <td>ML Data Source</td>
-      <td>Sample Name</td>
-      <td ng-repeat="at in analysis.exampleCols">{{
+      <th> </th>
+      <th>
+        <span class="row-control">
+          <i class="fa fa-bars" aria-hidden="true"></i>
+        </span>
+        ML Data Source
+      </th>
+      <th>id</th>
+      <th>Sample Name</th>
+      <th ng-repeat="at in analysis.exampleCols">{{
         at.typename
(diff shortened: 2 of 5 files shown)


mhuyck:
Demo now running at http://greenelab.fgtech.com (had to wait for DNS to propagate). You may need to refresh your browser cache for this to fully load the new UI.

interface/src/app/analyze/analysis/analysis.less
```diff
@@ -1,3 +1,37 @@
...
+    border: 0;
+    -webkit-appearance: none;
+  }
+}
+// allow us to use controls to help align headers, but don't show them
+th .row-control, thead .row-control {
+  visibility: hidden;
+}
+
+// sample group highlighting
```
dongbohu: I like the highlight colors you set on the web interface. Do you think it will be helpful to describe the color roughly using comments?

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -1,33 +1,57 @@
...
   <tbody ng-model="sb.heatmapData.samples" as-sortable="sortableOptions">
-    <tr ng-repeat="id in sb.heatmapData.samples" as-sortable-item>
-      <td>
-        <button type="button" class="btn close"
-        ng-click="sb.removeSample(id)" aria-label="Delete">
+    <tr ng-repeat="id in sb.heatmapData.samples"
+        ng-mouseover="showControls = true" ng-mouseleave="showControls = false"
+        ng-class="'sample-' + sb.sampleGroups[id]" as-sortable-item>
+      <td class="row-control">
+        <input type="radio" ng-model="sb.sampleGroups[id]"
```
dongbohu: On my web browser (Google chrome), these three radio buttons are not very easy to find because they are gray. How easy is it to make them more obvious? (If it needs a lot of tweaking, you can do it sometime later.)

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -77,6 +80,26 @@ function($log, $cacheFactory, $q, Sample, Activity) {
...
+    getSampleGroups: function() {
+      var keys = Object.keys(this.sampleGroups);
+      var samplesByGroup = {};
+      var i;
+
+      // each distinct value in sampleGroups becomes a key in samplesByGroup,
+      // and the keys of sampleGroups are collected in a list within each
+      // corresponding value of samplesByGroup
+      for (i = 0; i < keys.length; i++) {
+        if (samplesByGroup[this.sampleGroups[+keys[i]]]) {
```
dongbohu: I guess lines 92-97 can be also written a little more compact like this:
```
if (!samplesByGroup[this.sampleGroups[+keys[i]]]) {
  samplesByGroup[this.sampleGroups[+keys[i]]] = [];
}
samplesByGroup[this.sampleGroups[+keys[i]]].push(+keys[i]);
```
or even:
```
samplesByGroup[this.sampleGroups[+keys[i]]] = samplesByGroup[this.sampleGroups[+keys[i]]] || [];
samplesByGroup[this.sampleGroups[+keys[i]]].push(+keys[i]);
```
Just my random thoughts, you don't have to change if you don't like them.

rzelayafavila: I think this code is definitely good as it is, but I am just curious of what you all think of using a variable inside this for-loop to hold the value of what ``+keys[i]`` is, just to make it more readable. Maybe something like ``var groupID = +keys[i]`` or ``var groupNumber = +keys[i]``? What do you think?

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -77,6 +80,26 @@ function($log, $cacheFactory, $q, Sample, Activity) {
...
+          samplesByGroup[this.sampleGroups[+keys[i]]].push(+keys[i]);
+        } else {
+          samplesByGroup[this.sampleGroups[+keys[i]]] = [];
```
rzelayafavila: Also, in here, what do you think about instead using this form:
```
if (!samplesByGroup[this.sampleGroups[+keys[i]]]) {
  samplesByGroup[this.sampleGroups[+keys[i]]] = [];
}
samplesByGroup[this.sampleGroups[+keys[i]]].push(+keys[i]);
```
? Also, ``+keys[i]`` can be renamed, as I mentioned above.

interface/src/app/analyze/analysis/sampleBin.js
dongbohu: I agree that it will be more readable if we can use a variable to hold either `+keys[i]` or `sampleGroups[+keys[i]]`.

interface/src/app/analyze/analysis/analysis.less
mhuyck: That's a good question. I know in some editors you can see the color by selecting the hex RGB code `#b7baff` and using a menu command so it may not be necessary. Now that you mention it, however, it is not entirely clear from this context that these two variables are colors. I will add a comment that helps with that.

- [ ] clarify that "sample group highlighting" is the *colors* used for labeling samples groups

The reason I didn't try to come up with a name that describes the colors (e.g. "purple-highlight") is that I wasn't sure we would stick with these exact colors. I wanted the variable name to describe what it is *for* more than what it *is* so we can experiment with different colors if we choose. Does that rationale make sense?

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: That is an unfortunate side effect of the highlighting. I want to tweak it some more, but it will have to come later. One problem is that web browsers don't allow customization of the standard radio button colors. There are ways around that but none of them looked simple, and I didn't find anything in Bootstrap that helped either.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: Nice catch. I think your first suggestion is a little more clear. I will change it to that.

- [ ] streamline the logic here as @dongbohu suggests

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: That's okay. Let's deal with it some other time.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: I'll try that and see if it looks better. I'm usually more bothered by too many temp vars than extra brackets, but I can see how that might help in this case. Might even get a minuscule performance boost, too, but I'm no expert in JavaScript parsing optimizations, so I could be wrong there.

- [ ] try @rzelayafavila's temp var suggestion

interface/src/app/analyze/analysis/analysis.less
mhuyck: Turns out that constants like this belong in `src/less/variables.less`, so I moved them there. I also put a better comment here. There's more cleanup in the `.less/css` department to tackle along with issue #76.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: I went with a slightly different approach that I think helps make this murky code block even more clear: the ambiguous `sampleGroups` is now called `sampleToGroup` to emphasize its function, and the method `getSampleGroups` is now called `getSamplesByGroup`. I also used a temp var in the loop to clarify what that was. I think it's much easier to follow now.

mhuyck:
Okay, I think I got everything. Let me know if you think so, too @dongbohu & @rzelayafavila!

PR #88, opened 2017-02-27
Fetching local database gene information when selecting genesets and sending them to the gene-gene network page.

rzelayafavila:
(Since the gene-gene network page expects gene database IDs). Also cleaning up these files, and converting directives in tribe_client/genesets.js to Angular components.

### DIFF
--- a/interface/src/app/tribe_client/geneset-result-table.tpl.html
+++ b/interface/src/app/tribe_client/geneset-result-table.tpl.html
@@ -1,15 +1,18 @@
-<div>
-  <table class="table table-striped" name="searchResults">
-    <tr>
-      <th>Title</th>
-      <th>Species</th>
-      <th>Creator</th>
-    </tr>
+<table class="table table-striped" name="searchResults">
+  <tr>
+    <th></th>
+    <th>Title</th>
+    <th>Species</th>
+    <th>Creator</th>
+    <th>Genes</th>
+  </tr>
 
-    <tr ng-repeat="geneset in genesets">
-      <td>{{ geneset.title }}</td>
-      <td>{{ geneset.organism.scientific_name }}</td>
-      <td>{{ geneset.creator.username }}</td>
-    </tr>
-  </table>
-</div>
+  <tr ng-repeat="geneset in genesets">
+    <td><input type="radio" ng-model="selectedGeneset.geneset"
+      ng-value="{{geneset}}"></td>
+    <td>{{ geneset.title }}</td>
+    <td><i>{{ geneset.organism.scientific_name }}</i></td>
+    <td>{{ geneset.creator.username }}</td>
+    <td>{{ geneset.tip.genes.join(' ') }}</td>
+  </tr>
+</table>
--- a/interface/src/app/tribe_client/geneset-result-table.tpl.html
+++ b/interface/src/app/tribe_client/geneset-result-table.tpl.html
@@ -1,4 +1,4 @@
-<table class="table table-striped" name="searchResults">
+<table class="table table-striped">
   <tr>
     <th></th>
     <th>Title</th>
@@ -7,12 +7,12 @@
     <th>Genes</th>
   </tr>
 
-  <tr ng-repeat="geneset in genesets">
-    <td><input type="radio" ng-model="selectedGeneset.geneset"
+  <tr ng-repeat="geneset in $ctrl.genesets">
+    <td><input type="radio" ng-model="$ctrl.selectedGeneset"
       ng-value="{{geneset}}"></td>
     <td>{{ geneset.title }}</td>
     <td><i>{{ geneset.organism.scientific_name }}</i></td>
     <td>{{ geneset.creator.username }}</td>
-    <td>{{ geneset.tip.genes.join(' ') }}</td>
+    <td>{{ $ctrl.getGeneLabelList(geneset.geneObjs) }}</td>
   </tr>
 </table>
--- a/interface/src/app/tribe_client/geneset-search-bar.tpl.html
+++ b/interface/src/app/tribe_client/geneset-search-bar.tpl.html
@@ -1,9 +1,7 @@
 <div>
   <form class="form-inline">
     <label>Organism:</label>
-    <select ng-disabled="true">
-      <option ng-selected="true" value="organism">{{organism}}</option>  
-    </select>
+    <i>{{organism}}</i>
 
     <input ng-model="search.query" placeholder="Search text">
 
--- a/interface/src/app/tribe_client/geneset-search-bar.tpl.html
+++ b/interface/src/app/tribe_client/geneset-search-bar.tpl.html
@@ -1,10 +1,11 @@
 <div>
   <form class="form-inline">
     <label>Organism:</label>
-    <i>{{organism}}</i>
+    <i>{{$ctrl.organism}}</i>&nbsp;
+    <input ng-model="$ctrl.search.query" placeholder="Search text">
 
-    <input ng-model="search.query" placeholder="Search text">
-
-    <button class="btn btn-default" ng-click="searchGenesets()">Search</button>
+    <button class="btn btn-default" ng-click="$ctrl.searchGenesets()">
+      Search
+    </button>
   </form>
 </div>
--- a/interface/src/app/tribe_client/geneset-search-bar.tpl.html
+++ b/interface/src/app/tribe_client/geneset-search-bar.tpl.html
@@ -1,14 +1,19 @@
 <div>
-  <form class="form-inline">
-    <label>Organism:</label>
-    <i>{{$ctrl.organism}}</i>&nbsp;
-    <input ng-model="$ctrl.search.query" placeholder="Search text">
+  <label>Organism:</label>
+  <i>{{$ctrl.organism}}</i>
+  <span ng-if="!$ctrl.organism" class="text-muted">
+    No organism yet. Please select a <b>Machine Learning Model</b> above
+  </span>
+  <br>
 
-    <button class="btn btn-default" ng-click="$ctrl.searchGenesets()"
-      ng-disabled="!$ctrl.organism">
-      Search
-    </button>
-  </form>
+  <input ng-model="$ctrl.search.query" placeholder="Search gene set titles">
+  <button class="btn btn-default" ng-click="$ctrl.searchGenesets()"
+    ng-disabled="!$ctrl.organism">
+    Search
+  </button>
+  &nbsp;
+
+  {{$ctrl.searchStatus}}
   <div class="alert alert-danger" ng-if="$ctrl.genesetQueryStatus" role="alert">
     {{$ctrl.genesetQueryStatus}}
   </div>
--- a/interface/src/app/tribe_client/geneset-search-bar.tpl.html
+++ b/interface/src/app/tribe_client/geneset-search-bar.tpl.html
@@ -4,8 +4,15 @@
     <i>{{$ctrl.organism}}</i>&nbsp;
     <input ng-model="$ctrl.search.query" placeholder="Search text">
 
-    <button class="btn btn-default" ng-click="$ctrl.searchGenesets()">
+    <button class="btn btn-default" ng-click="$ctrl.searchGenesets()"
+      ng-disabled="!$ctrl.organism">
       Search
     </button>
   </form>
+  <div class="alert alert-danger" ng-if="$ctrl.genesetQueryStatus" role="alert">
+    {{$ctrl.genesetQueryStatus}}
+  </div>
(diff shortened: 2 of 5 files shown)


rzelayafavila:
I have a live instance running on my machine - you can test it out at http://165.123.67.154:8000/#/gene_search . However, you will need to be logged in to Tribe through this instance (you will only need to have a Tribe account) to be able to see this geneset selection panel.

interface/src/app/tribe_client/genesets.js
```diff
@@ -1,67 +1,98 @@
...
 
-      $scope.searchGenesets = function() {
-        Genesets.query($scope.search, function(data) {
-          $scope.genesetResultCount = data.meta.total_count;
-          $scope.genesets = data.objects;
+    self.searchGenesets = function() {
+      var qparams = {};
+      qparams['query'] = self.search.query;
+      qparams['organism__scientific_name'] = self.organism;
+      qparams['limit'] = self.limit;
```
mhuyck: A nicer way to write lines 50-53 (a little less typing) would be:

```javascript
var qparams = {
  'query': self.search.query,
  'organism__scientific_name': self.organism,
  'limit': self.limit
};
```

interface/src/app/tribe_client/geneset-result-table.tpl.html
```diff
@@ -1,15 +1,18 @@
-<div>
```
mhuyck: Thank you for getting rid of the extraneous `<div>`. I ❤️ clean HTML!

interface/src/app/tribe_client/geneset-result-table.tpl.html
```diff
@@ -1,15 +1,18 @@
...
-      <td>{{ geneset.organism.scientific_name }}</td>
-      <td>{{ geneset.creator.username }}</td>
-    </tr>
-  </table>
-</div>
+  <tr ng-repeat="geneset in $ctrl.genesets">
+    <td><input type="radio" ng-model="$ctrl.selectedGeneset"
+      ng-value="{{geneset}}"></td>
+    <td>{{ geneset.title }}</td>
+    <td><i>{{ geneset.organism.scientific_name }}</i></td>
```
mhuyck: Another ❤️ for italicized Latin! 😄

interface/src/app/tribe_client/geneset-result-table.tpl.html
```diff
@@ -1,15 +1,18 @@
...
+  </tr>
 
-    <tr ng-repeat="geneset in genesets">
-      <td>{{ geneset.title }}</td>
-      <td>{{ geneset.organism.scientific_name }}</td>
-      <td>{{ geneset.creator.username }}</td>
-    </tr>
-  </table>
-</div>
+  <tr ng-repeat="geneset in $ctrl.genesets">
```
mhuyck: I'm still playing catch-up with you on the switch to Angular `.component`s, so forgive the naïve question: I don't see `$ctrl` declared anywhere. Does that come in for "free" via `.component`?

mhuyck:
@rzelayafavila: I'm getting a debug page (**TemplateDoesNotExist at /tribe_client/**) when I sign in to my Tribe account and choose "My Gene Sets" from the Tribe account drop-down menu. Maybe this is because I don't have any gene sets defined on Tribe, but is there also a configuration error on your demo site?

mhuyck:
I also can't get anything to come back from the "Have a Tribe gene set you want to use?" section by searching for any *Organism*. Is this searching all of Tribe or possibly just looking for my non-existent gene sets again?

I guess this could also be an issue with missing indexes...?

rzelayafavila:
@mhuyck Hmmm let me check it out - I can sign in through tribe-client with no problem, let me see what is going on.

mhuyck:
Just to clarify: I can sign in *fine*, but after that I'm getting nothing to come back from searches.

mhuyck:
Ah... wait. I think I just didn't have the MLModel selected...

rzelayafavila:
@mhuyck - Ah, I just saw what was going on. I had forgotten to add a base.html file (which is actually just a symlink to the index.html file). This is the template that the tribe-client templates extend.

rzelayafavila:
@mhuyck : Yes, so I just realized - I should include some way that tells the user that they need to select a MLModel before they can search for Tribe genesets. This is something we do want to enforce, because we do not want to send the users to the gene-gene network page without having selected a MLModel (and through it an organism), right?

interface/src/app/tribe_client/geneset-search-bar.tpl.html
```diff
@@ -1,12 +1,11 @@
 <div>
   <form class="form-inline">
     <label>Organism:</label>
-    <select ng-disabled="true">
-      <option ng-selected="true" value="organism">{{organism}}</option>  
-    </select>
+    <i>{{$ctrl.organism}}</i>&nbsp;
+    <input ng-model="$ctrl.search.query" placeholder="Search text">
```
mhuyck: When no MLModel is selected (*way* at the top of this page, so out of sight), this looks like a form to search for an **Organism** because `{{$ctrl.organism}}` (line 4 above) is blank. In addition, all of my searches were failing silently, so I couldn't figure out how to make this work.

I'd suggest adding a message at line 4 saying something like "please select a **Machine Learning Model** above" and disabling this input until an MLModel is selected. (Assuming that is required for this to work.)

mhuyck: Continuing my sleep-deprived simulation of a dumb user: what index are we actually searching here? I tried some gene names but using genes that came up as hits under other more general searches (e.g. "dna") do not return results. Some things I tried: "mexR", "PA2196".

interface/src/app/tribe_client/genesets.js
```diff
@@ -1,67 +1,98 @@
...
+        var returnedGenesets = data.objects;
+        angular.forEach(returnedGenesets, function(geneset) {
+          var entrezIds = geneset.tip.genes.join(',');
+          Gene.get({'entrezid__in': entrezIds},
+            function success(response) {
+              geneset.geneObjs = response.objects;
+              self.genesets.push(geneset);
+            },
+            function error(err) {
+              $log.error('Failed to get gene information: ' + err);
```
dongbohu: Have you ever tested the error message generated on line 66? According to:
https://docs.angularjs.org/api/ngResource/service/$resource
```The error callback is called with (httpResponse) argument.```

And according to:
https://docs.angularjs.org/api/ng/service/$http
```
The response object has these properties:
* data – {string|Object} – The response body transformed with the transform functions.
* status – {number} – HTTP status code of the response.
* headers – {function([headerName])} – Header getter function.
* config – {Object} – The configuration object that was used to generate the request.
* statusText – {string} – HTTP status text of the response.
```
So the error message on line 66 will be something like:
```Failed to get gene information: object [Object]```
This is actually one reason why I wrote `errGen` service in **app/utils.js**.

File: interface/src/app/tribe_client/genesets.js [hunk 7]
```diff
@@ -1,67 +1,98 @@
...
+  'adage.tribe_client.resource',
+  'adage.gene.resource'
 ])
 
-.directive('genesetSearchForm', [function() {
-  return {
-    controller: ['$scope', function($scope) {
+// Wrapping component that houses the genesetSearchBar and the
+// genesetResultTable.
+.component('genesetSearchForm', {
```
dongbohu: **component** rocks!

interface/src/app/tribe_client/genesets.js
```diff
@@ -1,67 +1,98 @@
...
-        Genesets.query($scope.search, function(data) {
-          $scope.genesetResultCount = data.meta.total_count;
-          $scope.genesets = data.objects;
+    self.searchGenesets = function() {
+      var qparams = {};
+      qparams['query'] = self.search.query;
+      qparams['organism__scientific_name'] = self.organism;
+      qparams['limit'] = self.limit;
+
+      Genesets.query(qparams, function(data) {
```
mhuyck: There is a little lag before search results come back. It's short, but just enough to make me wonder if the site took my click. A little status note similar to what we have here would help: http://greenelab.fgtech.com/#/analyze?q=pao1

interface/src/app/tribe_client/genesets.js
```diff
@@ -1,67 +1,98 @@
...
 // Directive for table containing search results
-.directive('genesetResultTable', [function() {
-  return {
-    replace: true,
-    restrict: 'E',
-    scope: {
-      // TODO: Implement pagination for geneset search results in
-      // case the number of results ($scope.genesetResultCount above)
-      // is greater than this 'limit' parameter (the maximum number of
-      // geneset results that the Tribe API should return at once).
```
mhuyck: Did you implement pagination as noted here? Maybe I'm missing it. If not, no worries, but let's keep the comment so we don't forget!

interface/src/app/tribe_client/genesets.js
mhuyck: Nice catch!

interface/src/app/tribe_client/genesets.js
```diff
@@ -1,67 +1,98 @@
...
-    restrict: 'E',
-    scope: {
-      limit: '=',
-      organism: '=',
-      genesets: '='
-    },
-    templateUrl: 'tribe_client/geneset-search-bar.tpl.html'
-  };
-}])
+      });
```
mhuyck: I noticed that there's no feedback during the search (see my comment on line 55) as well. Also having a message to the effect that no results were found (or another error occurred with the search service) would be very helpful.

interface/src/app/tribe_client/geneset-result-table.tpl.html
rzelayafavila: Yep, exactly - it automatically does the 'controller as $ctrl' syntax for you.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Ok, I will change this.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Ah gotcha @dongbohu - thanks for catching this! I will change this to use the ``errGen`` service.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Yes, that makes sense, I will put in an error function for that as well.

interface/src/app/tribe_client/genesets.js
mhuyck: By the way: a lazy way around this `object [Object]` problem would be to say `JSON.stringify(err)`, but since @dongbohu already provided `errGen`, we should use that.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Mmm good point... ok, I will add a note to make sure the users know that the site did receive the click and is searching.

interface/src/app/tribe_client/genesets.js
rzelayafavila: Aha - good point. I actually have even considered not implementing pagination here at all, but I think that is a topic for another day, and in the meantime I will put this comment back in here so I don't forget.

rzelayafavila:
Ok @mhuyck and @dongbohu : I think I got all the comments - a new version of the page is running on that same url. Let me know how that looks! Thanks!

interface/src/app/tribe_client/genesets.js
mhuyck: Ah, I see. You mean that probably users wouldn't care about search hits beyond `limit`? That's a good question, but I agree it's a topic for another day.

interface/src/app/tribe_client/geneset-search-bar.tpl.html
mhuyck: Search feedback looks good here now. I'm still not getting a reply when I search on "mexR" or "PA2196," but I can see now that it's trying. Perhaps this is an Elasticsearch configuration or performance issue? At any rate, this doesn't appear to be an issue with the code changed in this PR.

rzelayafavila: Yeah - so this is definitely more related to how Tribe handles gene set searches. You can only really search by the gene set title or any word contained in the abstract right now, not by any genes present in the actual gene sets. I created an issue for it in the Tribe repo. in bitbucket.

rzelayafavila:
Ok, I replied to that comment @mhuyck. Thank you and @dongbohu for your help with this PR!

interface/src/app/tribe_client/geneset-search-bar.tpl.html
mhuyck: Good to know. Thanks!

PR #89, opened 2017-02-28
Volcano plot vega

mhuyck:
Live volcano plots!

A couple of things missing:

1. hovering over a node should display its *name* rather than its ID number
2. we should be able to pull a list of selected nodes back out of vega
3. selecting nodes should be simpler

The first two items are important for integrating this feature with the rest of the `adage-server` so I
have been working on getting them fixed. Both of these require debugging vega, which I am not having
any luck with at the moment. It's probably just that I'm tired, but also my procedure for dropping vega
into a debugger is just not working and I can't sort out why.

I will probably work around vega for (1) and just push node names into the source data. That workaround will add a method to sampleBin and alter some lines in its `getVolcanoPlotData()` method but the bulk of it will remain the same so it could be reviewed now.


mhuyck:
Regarding codeclimate: I tried to get it to ignore files named `*-vgspec.js`. Can anyone figure out why it's analyzing that? I can probably do something about that other block with similar code, actually.

mhuyck:
Demo site now updated: http://greenelab.fgtech.com/.

To see a volcano plot, add samples, select at least two samples for each group and then click on the "Volcano Plot" button below the heatmap.

interface/src/app/volcano-plot/volcano-plot.js
```diff
@@ -0,0 +1,67 @@
...
+  'greenelab.stats',
+  'adage.analyze.sampleBin',
+  'adage.volcano-plot-vgspec'
+])
+
+.controller('VolcanoPlotCtrl', ['$scope', 'VolcanoPlotSpec', 'SampleBin',
+  function VolcanoPlotCtrl($scope, VolcanoPlotSpec, SampleBin) {
+    // retrieve the nodeInfo map (for node names) via the SampleBin service
+    // $scope.nodeInfo = SampleBin.getNodeInfo(); // TBD
+    $scope.nodeInfo = [
```
dongbohu: Is this `nodeInfo` assignment a test? I didn't see it being used anywhere else.

If it is not a test, since you already used `controllerAs` syntax, you may want to assign it as `this.nodeInfo`, then $scope dependency can be removed.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -277,15 +280,65 @@ function($log, $cacheFactory, $q, Sample, Activity) {
...
+          nodeObject.activityA, nodeObject.activityB
+        ).pValue());
+
+        return nodeObject;
+      }, this);
+      return retval;
     }
   };
 
   return SampleBin;
```
dongbohu: Never mind, using `service` will save a few lines, but if you are not familiar with it, forget it.

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -29,9 +29,9 @@ angular.module('adage.analyze.analysis', [
   });
 }])
 
-.controller('AnalysisCtrl', ['$scope', '$log', '$location', '$q', 'Sample',
-'Activity', 'AnnotationType', 'SampleBin',
-function AnalysisCtrl($scope, $log, $location, $q, Sample, Activity,
+.controller('AnalysisCtrl', ['$scope', '$log', '$location', '$q', '$state',
+'Sample', 'Activity', 'AnnotationType', 'SampleBin',
+function AnalysisCtrl($scope, $log, $location, $q, $state, Sample, Activity,
```
rzelayafavila: Hmmm - Are ``$location``, ``Sample``, and ``Activity`` being used in this controller? I don't think I see them below.

interface/src/app/volcano-plot/view.js
```diff
@@ -0,0 +1,34 @@
+angular.module('adage.volcano-plot.view', [
```
rzelayafavila: I'm just curious - why the hyphen instead of the camelCase for ``volcano-plot``?

interface/src/app/volcano-plot/view.js
```diff
@@ -0,0 +1,34 @@
...
+}])
+
+.controller('VolcanoPlotViewCtrl', ['SampleBin', '$stateParams',
+  // TODO make use of $stateParams (pulling from SampleBin for initial tests,
+  //      but the right way to do this is to refactor so view.js pulls params
+  //      for sample-group-a and sample-group-b and does what's necessary to
+  //      make a plot from those lists)
+  function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
+    // give our templates a way to access the SampleBin service
+    this.sb = SampleBin;
```
rzelayafavila: Is ``this.sb`` being used anywhere right now?

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: I had forgotten all about `.service`! I will take a look at using that where appropriate in the context of issue #76.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -277,15 +280,65 @@ function($log, $cacheFactory, $q, Sample, Activity) {
...
 }])
 
-.controller('SampleBinCtrl', ['$scope', '$log', '$uibModal', 'Sample',
-'SampleBin',
-function SampleBinCtrl($scope, $log, $uibModal, Sample, SampleBin) {
+.controller('SampleBinCtrl', ['$scope', '$log', 'SampleBin',
```
mhuyck: Good catch. I fixed `$log`. There is, indeed, lots that no longer makes sense about the organization of `sampleBin.js`. I already have some thoughts along those lines, but they will have to wait for #76.

interface/src/app/volcano-plot/view.js
```diff
@@ -0,0 +1,34 @@
...
+    data: {pageTitle: 'Volcano Plot'}
+  });
+}])
+
+.controller('VolcanoPlotViewCtrl', ['SampleBin', '$stateParams',
+  // TODO make use of $stateParams (pulling from SampleBin for initial tests,
+  //      but the right way to do this is to refactor so view.js pulls params
+  //      for sample-group-a and sample-group-b and does what's necessary to
+  //      make a plot from those lists)
+  function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
```
mhuyck: It should be used for this view, but it's not in there yet because of the trouble I ran into with the volcano plot selection. I'll put a comment in here to indicate how it will be used.

- [ ] add a note about why $stateParams is injected into `volcano-plot/view.js`.

interface/src/app/volcano-plot/volcano-plot.js
mhuyck: Yes, this is test data. I started writing the fetch for the actual `nodeInfo` data in line 17, above, but decided to get the vega part working first. I haven't circled back to remove it because I'm still getting that part to work.

.codeclimate.yml
```diff
@@ -51,3 +51,4 @@ ratings:
   - "**.py"
 exclude_paths:
 - "adage/**/migrations/"
+- "interface/**/*-vgspec.js"
```
mhuyck: Right, and it seems to have worked. Codeclimate had nothing more to say about `volcano-plot-vgspec.js` after I moved the exclusion here.

interface/src/app/analyze/analysis/analysis.js
mhuyck: Good catch. I will fix that!

- [ ] check dependencies injected into `AnalysisCtrl`

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -57,6 +57,10 @@ AnnotationType, SampleBin) {
     });
   };
 
+  $scope.showVolcanoPlot = function() {
+    $state.go('volcano');
```
mhuyck: Yes, we need parameters but it's not implemented yet. Two reasons: for (1), see the comment I made above in response to Dongbo in `view.js`, line 25, about the selection. There are other parameters that would be useful, too. We should be able to specify the sample groups that way, for example. Reason (2) that these parameters aren't implemented yet is that I'll need to do some refactoring to make them work.

I guess that reason (2) sounds a bit cryptic without explaining my thinking a little more. With the current implementation, the sample groups are defined in `SampleBin` and pulled out of there (behind the scenes) by `view.js`. If this page were to respond automatically to what's in the URL parameters, I'd need to re-implement a bunch of the code that currently retrieves activity data, *etc.*, inside `SampleBin`. The right way to make this happen is to refactor `SampleBin` (which I plan to do, as I mentioned above) and then use those refactored components to implement the parameters here.

The picture for what components we need and how they fit together has become more clear to me as I see how the volcano plot and heatmap play off of each other. We can evolve the code to get there, but not before our release deadline! This gets noted under issue #76.

interface/src/app/volcano-plot/view.js
mhuyck: Another insightful comment. Something didn't feel right about naming the whole directory `volcanoPlot` or `volcano_plot`, so I went with `volcano-plot`. In part, I chose this because that's how Angular modules themselves are organized. Inside `interface/vendor/`, you can find `angular-animate`, `angular-bootstrap`, `angular-mocks`, etc. I followed that pattern.

I'm flexible, however. If you don't like it, I can switch it to `volcanoPlot`. But not `volcano_plot`. Underscores bother me. Hmm, maybe that's the real reason I prefer `volcano-plot` over the others: both `volcanoPlot` and `volcano_plot` require me to make use of the shift key to type them. Darned shift key.

interface/src/app/volcano-plot/view.js
mhuyck: Nope. Good catch! Removed it.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -277,15 +280,65 @@ function($log, $cacheFactory, $q, Sample, Activity) {
       //  note: progress can be reported by returning a $promise to the caller
       // respObj.queryStatus = 'Retrieving sample activity...';
       this.rebuildHeatmapActivity(this.heatmapData.samples);
+    },
+
+    // volcano plot methods
+    getVolcanoPlotData: function() {
```
mhuyck: As you may remember, I have a divided mind about re-assigning `this` to self. I don't think the reason to do so would be *readability*, exactly. The main issue, as I see it, is *correctness*. My feeling is that as long as `this` is used correctly, there is no trouble with using it and I think there are benefits to using it in an undisguised fashion.

There are six occurrences of the keyword in the `getVolcanoPlotData` method here. We use `this` three times to invoke a sibling method on the `SampleBin` object, which seems straightforward and clearly correct to me. In each of the remaining three cases, `this` is handed off as a parameter of `Array.prototype.map`, where it must be used because the callback function in the `.map` call needs it in order to properly access the `activityCache` property of the `SampleBin` instance. I suppose that is the one case where it *might* seem confusing, but now that I've learned how JavaScript's `this` actually works, I am no longer confused by it.

Clear as mud, I guess. What do you think, would the code be easier to follow if `this` were re-assigned to, say, `SampleBin`, which is what it is? If we do that, then we're masking the outer `SampleBin` and maybe *that's* even more confusing. See what I mean?

interface/src/app/volcano-plot/view.js
mhuyck: Oh, I already had a note! See the TODO starting at line 21.

interface/src/app/analyze/analysis/analysis.js
mhuyck: Gone. Also got rid of `AnnotationType` and its resource. We get those for "free" via `SampleBin`. I think these were missed from a previous refactoring.

rzelayafavila:
Woah - what happened to the circle-ci tests/checks?

.codeclimate.yml
rzelayafavila: Nice!

interface/src/app/analyze/analysis/analysis.js
rzelayafavila: Ok, as I said this is good getting filed under an issue for now. I also saw your TODO comments in volcano-plot/view.js. Ok, cool!

interface/src/app/analyze/analysis/sampleBin.js
rzelayafavila: Me too.

rzelayafavila:
Ok, I think you have gotten all the stuff that I commented on @mhuyck! Good stuff!

rzelayafavila:
It appears that github is making a few changes to the UI today. It does look a bit more stylish, but maybe that is what is going on with the circle-ci tests bug.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: I'm not wild about that aspect of JavaScript, either, but the only thing to be done for that is to write in a different language. 😄 

We don't really have a choice, since it is a part of the language. Assigning to a temp variable doesn't really make the problem go away, either. It's just something you need to do in some instances to work around this aspect of the language. All of the cases where it's used in this part of the code are done correctly.

mhuyck:
Yeah, I noticed that CircleCI disappeared here but was still there on your PR #88. I'm thinking it's a glitch and maybe it will go away if I push another update?

mhuyck:
At any rate, I got a green light from Codeship, so I think we're covered.

mhuyck:
Okay, I'll merge this as it is and file a separate PR once I've got the node labels worked in. Thank you both!

PR #98, opened 2017-03-08
Adding some content to Home page from ADAGEpath README.

rzelayafavila:


interface/src/app/home/home.tpl.html
```diff
@@ -1,6 +1,40 @@
 <div>
   <h1>ADAGE: Home</h1>
-  <p class="lead">
-    This shell is the framework into which the ADAGE server will grow.
-  </p>
+  <div class="lead">
+    <h2>ADAGE introduction</h2>
+    <p>
+      ADAGE (or eADAGE) models are neural networks trained in an unsupervise
```
rzelayafavila: @tj8901nm: by the way, it looks like there is a typo in "...unsupervise" with no "d" at the end in the README of ADAGEpath itself. I will correct it here.

tj8901nm:
@cgreene @georgiadoing This is basically copied from the README of the ADAGEpath repo. It briefly described what ADAGE is and what ADAGE signature analysis is. I think the home page should also show the users how to start an analysis in the server.

rzelayafavila:
By the way reviewers, you can browse a live version of this Home page here: http://165.123.67.154:8000/#/home

PR #99, opened 2017-03-08
Change the regular slider into a range slider in Gene Network

dongbohu:
Demo page:
http://165.123.67.202/#/gene_network/?genes=6547

@tj8901nm: You don't have to check the code (unless you want). Please check the demo page and let me know whether this is the exact functionality that you need. Thanks.

### DIFF
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -93,7 +93,7 @@ angular.module('adage.gene.network', [
 
       var geneTip, edgeTip;
 
-      self.updateSvg = function() {
+      self.renderNetwork = function() {
         geneTip.hide();
         edgeTip.hide();
         var correlationSign;
@@ -121,7 +121,7 @@ angular.module('adage.gene.network', [
           showTicksValues: true,
           noSwitching: true,
           onEnd: function(id, low, high) {
-            self.updateSvg();
+            self.renderNetwork();
           }
         }
       };
@@ -282,7 +282,7 @@ angular.module('adage.gene.network', [
 
         // Draw network svg with legend.
         network.showLegend();
-        self.updateSvg();
+        self.renderNetwork();
 
         // Add event handlers to gene-tip and edge-tip so that they will be
         // hidden when clicked:
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -57,9 +57,15 @@ angular.module('adage.gene.network', [
    ['$stateParams', 'EdgeService', 'NodeService', '$log',
     function GeneNetworkController($stateParams, EdgeService, NodeService,
                                    $log) {
+      var self = this;
+      // Do nothing if no genes are specified in URL.
+      if (!$stateParams.genes || !$stateParams.genes.split(',').length) {
+        self.statusMessage = 'No genes are specified.';
+        return;
+      }
+
       var rawMinWeight = -1.0;
       var rawMaxWeight = 1.0;
-      var minCorrelation = rawMinWeight;
       var scaledMaxWeight = rawMaxWeight - rawMinWeight;
       var setEdgeColor = d3.scale.linear()
           .domain([0, scaledMaxWeight / 2.0, scaledMaxWeight])
@@ -76,33 +82,26 @@ angular.module('adage.gene.network', [
           .legendEnd(rawMaxWeight)
           .legendText('Correlation');
 
-      var self = this;
       // The following properties of "self" will be available to HTML.
       self.maxNodeNum = Number.MAX_SAFE_INTEGER;
       self.statusMessage = 'Connecting to the server ...';
       self.slider = {  // Slider configuration
-        position: minCorrelation,
+        minValue: rawMinWeight,
+        maxValue: rawMaxWeight,
         options: {
           floor: rawMinWeight,
           ceil: rawMaxWeight,
           step: 0.01,
           precision: 2,
-          showSelectionBarEnd: true,
           showTicks: 0.5,
           showTicksValues: true,
-          onEnd: function(id, value) {
-            minCorrelation = value;
-            network.filter(minCorrelation - rawMinWeight, self.maxNodeNum)
+          onEnd: function(id, low, high) {
+            network.filter(low - rawMinWeight, high - rawMinWeight,
+                           self.maxNodeNum)
               .draw();
           }
         }
       };
-
-      // Do nothing if no genes are specified in URL.
-      if (!$stateParams.genes || !$stateParams.genes.split(',').length) {
-        self.statusMessage = 'No genes are specified.';
-        return;
-      }
 
       var urlGenes = $stateParams.genes.split(',');
       var genes = [];
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -6,7 +6,8 @@ angular.module('adage.gene.network', [
   'ui.router',
   'ui.bootstrap',
   'ngResource',
-  'rzModule'
+  'rzModule',
+  'adage.utils'
 ])
 
 .config(['$stateProvider', function($stateProvider) {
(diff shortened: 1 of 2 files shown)


tj8901nm:
@dongbohu It seems correct. I cannot confirm because there is no negative correlation edge in the example. Could you use the gene PA0485 to build the network?

dongbohu:
@tj8901nm: Please try this URL:

dongbohu:
I changed the code so that the gene and edge tips will be hidden every time a user changes the slider bar.

rzelayafavila:
I just played around a bit with the UI and it looks great!

tj8901nm:
@dongbohu thanks! After playing with it, I realized that what we really want is to filter by absolute correlation values. Users only want to focus on edges with high correlations no matter it's positive or negative. Perhaps we should use the slider to represent absolute correlation and then add two click dots to select whether to show positive or negative correlated edges. Sorry for not being clear enough at the beginning!

dongbohu:
@tj8901nm: Thank you for the clarification. So the min and max values of the range slider will be always the same sign? Is it possible that a user may want to set the range to something like (-0.6, 0.3)?

dongbohu:
@rzelayafavila and @mhuyck: Please do not review this PR now. I am working with Jie to figure out the exact user case of this feature.

mhuyck:
Ah.. I see you've changed this to a WIP, so I'm happy to look again after you've made further changes.

rzelayafavila:
@dongbohu Ok, gotcha

mhuyck:
Another thought about the slider that might not be trivial to implement, so probably for a separate PR:

I find myself guessing where the next set of edges will get filtered out when adjusting the slider. Often I get it wrong and go too far, resulting in a network with no edges left. Would there be a way to plot the distribution of edge correlations right on the slider somehow? Perhaps you could put tick marks showing the actual distribution along the axis of the slider. Not sure if this would be possible with the slider we are using.

tj8901nm:
@dongbohu The slider reflects absolute correlation value. For example, if it is set to 0.5 to 1, both edges between -0.5 to -1 and edges between 0.5 to 1 will be shown. Then there are two check boxes on the side to choose whether to only show positive correlation and negative correlation. The default would be having both boxes checked. I feel it's quite unlikely for users to be interested in the range (-0.6, 0.3). 

Related to this topic, I think we can set a default range to the slider so that not all edges are shown when opening a new network page. I think (0.5, 1) would be a good default.

dongbohu:
After reading the documentation, I don't think it is easy to show the exact distribution of all edge weights, but we can set min and maximum values, beyond which a user can not move the slider bar. Let me implement it and you guys tell me what you think.

@tj8901nm: I just saw your reply. Thank you the explanation. I will implement the two check boxes and the default 0.5-1 range. Will let you know when it is ready for review.

dongbohu:
@tj8901nm: Please take a look at the demo page again:
http://165.123.67.202/#/gene_network/?genes=6323
and tell me whether this is what you need. Thanks.

tj8901nm:
@dongbohu It looks great! Exactly what I asked for. I just have two comments. 1. Could you please make the edge color still reflect correlation in the (-1,1) range? Right now, it's difficult to tell which edge is a positively correlated edge and which is negative. 2. The text beside the check box. Could you change `Include positive weights` to `Include positive correlation`? Thanks!

dongbohu:
@tj8901nm: Thank you very much for your feedback. 

For your comment #2, it is pretty easy to do. I have changed the code already.

For your comment #1, it is not as straightforward as I thought, because the library that I am using to show the network (d3.network) only accepts positive edge weights, based on which the edge color is calculated. Let me check the code again and figure out how to change that part.

tj8901nm:
@dongbohu I think in the old version, the correlation color bar has a range from -1 to 1, right?

dongbohu:
Yes, the older version does have correlation from -1 to 1, but internally it was converted into (0, 2) so that d3.network will accept them. But now with the implementation of range of absolute value, I can not use that trick any more. Anyway, let me think about the best way to handle it. Thanks.

tj8901nm:
@dongbohu Thanks for explaining it!

cgreene:
I am a bit confused about the two different sliders. Is the goal to convert that back to just one slider?

Also, will negative correlations "pull" on the force graph if it's via an absolute value? Is this what we want from a user point of view?

dongbohu:
@cgreene: The original version actually has only one slider, which allows a user to set the minimum correlation value on edges. For example, if the slider's position is 0.34, then all correlation values in the range of (0.34, 1.0) will be shown. The new version has two sliders, which allow the user to set both minimum and maximum range of correlations (instead of having the maximum set to 1.0 implicitly). For example, if the left sider is at 0.34 and the right one at 0.56, then only the edges whose correlations are in (0.34, 0.56) will be shown.

The negative correlations are always converted into their absolute values. That is why the range bar is from 0 to 1.

dongbohu:
@tj8901nm: I modified the d3.network library. Can you try the same URL again and tell me whether your comment (1) has been addressed? Thanks.

tj8901nm:
@dongbohu Yes, the color now looks great! It's just that the checkboxes seem to do nothing in the example link. Are you going to fix this also? Thanks!

dongbohu:
@tj8901nm: Thanks for your feedback. The checkbox issue is fixed now.

tj8901nm:
Looks great to me!

dongbohu:
Thanks again to @tj8901nm for testing this new feature! I will file a separate PR in d3.network and ask the author of d3.network to review it. After that I will  update the current PR.

dongbohu:
The code has been updated based on Jie's feedback. Please review this PR again. The demo page is still valid. Thanks.

dongbohu:
Don't know why eslint on codeclimate doesn't like the indentation of those two lines. eslint on my desktop is okay with them.

cgreene:
@dongbohu : probably one of the config settings but not necessary to pass to merge

dongbohu:
Although the code may seem to be revised a lot, most of them are related to the configuration of the the new range slider and adoptions of `errGen` and `ApiBasePath` services.

interface/src/app/gene/network/network.js
```diff
@@ -39,71 +40,92 @@ angular.module('adage.gene.network', [
...
 
-      var self = this;
-      // The following properties of "self" will be available to HTML.
-      self.maxNodeNum = Number.MAX_SAFE_INTEGER;
-      self.statusMessage = 'Connecting to the server ...';
-      self.slider = {  // Slider configuration
-        position: minCorrelation,
+      var geneTip, edgeTip;
+
+      self.updateSvg = function() {
```
mhuyck: Minor quibble: can we name this function something closer to its purpose within the scope of `network.js`? I know the network diagram is drawn using SVG, but that is an implementation detail and none of the actual code inside this function actually does anything to manipulate SVG. I'm thinking the right abstraction level would be something like `renderNetwork` or `redrawNetwork`.

dongbohu:
The eslint issues reported by codeclimate seem to be because my desktop was using an older version of eslint than codeclimate does, and the rule of indentation in the older version is not as strict as the one in the newer version. After I update my local eslint, it showed the two issues and then I fixed them.

mhuyck:
I'm getting a "Template does not exist" error on the demo site...

dongbohu:
Probably because I changed the location of my github repository a moment ago. Let me take a look.

dongbohu:
Fixed.

interface/src/app/gene/network/network.js
dongbohu: Sure. I thought of a name like `updateNetowrk`, but the word `network` seems to be too generic (because it includes both the network data and the graphical presentation). I like `renderNetwork`.

mhuyck: Good thinking. Using "render" makes it clear it's for the visual representation.

interface/src/app/gene/network/network.js
```diff
@@ -39,71 +40,92 @@ angular.module('adage.gene.network', [
...
+  function($resource, ApiBasePath) {
   // Possible parameters for this endpoint when making a query are:
   // {
   //   genes: Database IDs of genes associated with the edge,
   //   mlmodel: Database ID of MLModel associated with the edge,
   //   limit: Maximum number of results to return
   //  }
-  return $resource('/api/v0/edge/');
-}])
+    return $resource(ApiBasePath + 'edge');
```

interface/src/app/gene/network/network.js
```diff
@@ -39,71 +40,92 @@ angular.module('adage.gene.network', [
...
+        self.statusMessage = 'No genes are specified.';
+        return;
+      }
+
+      // The following properties of "self" will be available to HTML.
+      self.includePositive = true;
+      self.includeNegative = true;
+      self.statusMessage = 'Connecting to the server ...';
+
+      var minCorrelation = -1.0, maxCorrelation = 1.0;
```
rzelayafavila: It's good that these parameters are defined up here!

PR #103, opened 2017-03-24
Adding fdr multiple test correction library to installed front-end packages and browserifying to make it available to our UI code, and adding a unit test.

rzelayafavila:


rzelayafavila:
Hmmm, I will check out this indentation issue that codeclimate is complaining about.

rzelayafavila:
Ok, I fixed the indentation issue. Also thanks @dongbohu for the tips on updating eslint!

interface/build.config.js
```diff
@@ -79,8 +79,13 @@ module.exports = {
       'vendor/angularjs-slider/dist/rzslider.min.js',
       'vendor/d3-tip/index.js',
       'node_modules/hclusterjs/hcluster.js',
+
+      // note: greenelab.stats.ttest.js and greenelab.stats.multtest.js are
+      // generated via browserify. greenelab.stats.ttest.js also needs to use
+      // babel to be able to be browserified.
```
mhuyck: To be precise, browserifying works fine without the babel step, but we need to use babel for compatibility. The `ttest` code was written in ES6 and all of our other JavaScript is in ES5. Beyond simply standardizing on ES5, we use babel so `ttest`s will run without issues on a broad spectrum of browsers.

Side note: in addition, our `uglify` minification step cannot handle certain ES6 constructs yet. Work is being done on that, but the future of minification might actually turn out to be using babel...

interface/src/app/greenelab/stats.spec.js
```diff
@@ -120,3 +120,102 @@ describe('tTest in stats.js', function() {
...
+      0.8478539, 0.7310656, 0.8563821, 0.7310656, 0.7310656, 0.9933630,
+      0.5110014, 0.8478539, 0.8060911, 0.9851209, 0.8728572, 0.9851209,
+      0.8478539, 0.8478539, 0.6831366, 0.8563821, 0.7310656, 0.5110014,
+      0.8478539, 0.7310656, 0.8478539, 0.8478539, 0.8478539, 0.7310656,
+      0.4836888, 0.8478539, 0.8478539, 0.8478539, 0.9851209, 0.8060911,
+      0.8478539, 0.8478539, 0.7310656, 0.8478539, 0.9851209, 0.8478539,
+      0.8478539, 0.8728572, 0.8478539, 0.5039586
+    ]);
+  });
+});
```
mhuyck: Nice tests! 👍

rzelayafavila:
Ok great, I changed that comment in build.config.js based on your suggestion @mhuyck, so I think this PR should be all set. Thanks!

PR #105, opened 2017-03-31
New model of gene-sample expression value and management command to load data sheet into database

dongbohu:
Although this PR includes 4 files, you can skip these two:
* migrations/0006_expressionvalue.py (generated by database migration)
* data/all-pseudomonas-gene-normalized.pcl (datasheet provided by Jie)

`models.py` includes a new class to model the gene-sample relationship.
`import_gene_sample_expr.py` is the management command to read the data sheet into backend database.

### DIFF
--- a/adage/analyze/management/commands/import_gene_sample_expr.py
+++ b/adage/analyze/management/commands/import_gene_sample_expr.py
@@ -70,7 +70,7 @@ def import_expr(file_handle, organism_tax_id):
     except Organism.DoesNotExist:
         raise Exception("Input organism_tax_id is not found in the database. "
                         "Please use the management command "
-                        "'organism_create_or_update.py' in djago-organism "
+                        "'organism_create_or_update.py' in django-organisms "
                         "package to create this organism.")
 
     # Enclose reading/importing process in a transaction context manager.
@@ -114,7 +114,7 @@ def read_header(tokens, samples):
                 samples.append(sample)
             except Sample.DoesNotExist:
                 samples.append(None)
-                logger.warn(
+                logger.warning(
                     "Input file line #1: data_source in column #%d not found "
                     "in the database: %s", index + 2, data_source)
 
@@ -148,7 +148,7 @@ def import_data_line(line_num, tokens, samples, organism):
     except Gene.DoesNotExist:
         # If a gene is not found in database, generate a warning message
         # and skip this line.
-        logger.warn(
+        logger.warning(
             "Input file line #%d: gene name %s (column #1) not found in "
             "database", line_num, gene_name)
         return
--- a/adage/analyze/management/commands/import_gene_sample_expr.py
+++ b/adage/analyze/management/commands/import_gene_sample_expr.py
@@ -2,13 +2,13 @@
 
 """
 This management command reads an input file of gene-sample expression values
-and loads the valid data into the database.  It should be invoked like this:
+and loads the valid data into database.  It should be invoked like this:
 
   python manage.py import_gene_sample_expr <expression_filename>
 
 If a data source (on the first row) or gene name (on the first column) is
 not found in the database, a warning message will be generated and the
-corresponding rows or columns will be skipped.
+corresponding column or row will be skipped.
 """
 
 
--- a/adage/analyze/management/commands/import_gene_sample_expr.py
+++ b/adage/analyze/management/commands/import_gene_sample_expr.py
@@ -2,11 +2,11 @@
 
 """
 This management command reads an input file of gene-sample expression values
-and load the valild data into the database.  It should be invoked like this:
+and loads the valid data into the database.  It should be invoked like this:
 
   python manage.py import_gene_sample_expr <expression_filename>
 
-If a data source (on the first row) or gene name (on the c=first column) is
+If a data source (on the first row) or gene name (on the first column) is
 not found in the database, a warning message will be generated and the
 corresponding rows or columns will be skipped.
 """
--- a/adage/analyze/management/commands/import_gene_sample_expr.py
+++ b/adage/analyze/management/commands/import_gene_sample_expr.py
@@ -2,19 +2,35 @@
 
 """
 This management command reads an input file of gene-sample expression values
-and loads the valid data into database.  It should be invoked like this:
+and loads the valid data into the database.  It should be invoked like this:
 
-  python manage.py import_gene_sample_expr <expression_filename>
+  python manage.py import_gene_sample_expr <expression_filename> \
+<organism_tax_id>
 
-If a data source (on the first row) or gene name (on the first column) is
-not found in the database, a warning message will be generated and the
-corresponding column or row will be skipped.
+The two required arguments are:
+  (1) expression_filename: input file of gene-sample expression values;
+  (2) organism_tax_id: taxonomy ID of the organism.
+
+For example, to load the sample-gene expression values for the organism
+"Pseudomonas aeruginosa" (whose taxonomy ID is 208964), the command will be:
+  python manage.py input_filename 208964
+
+IMPORTANT:
+(1) Before running this command, please make sure that "django-organisms"
+package has been installed and organism_tax_id already exists in the database.
+If organism_tax_id is not in the database yet, please use the management
+command "organisms_create_or_update.py" (in "django-organisms" package)
+to add it.
+(2) If a data source (on the first row of input file) or gene name (on the
+first column of input file) is not found in the database, a warning message
+will be generated and the corresponding column or row will be skipped.
 """
 
 
 from __future__ import print_function
 from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
+from organisms.models import Organism
 from genes.models import Gene
 from analyze.models import Sample, ExpressionValue
 
@@ -28,10 +44,12 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument('expression_filename', type=file)
+        parser.add_argument('organism_tax_id', type=int)
 
     def handle(self, **options):
         try:
-            import_expr(options['expression_filename'])
+            import_expr(options['expression_filename'],
+                        options['organism_tax_id'])
             self.stdout.write(self.style.NOTICE(
                 "Imported gene-sample expression values successfully"))
         except Exception as e:
@@ -40,11 +58,20 @@ class Command(BaseCommand):
                 "values: %s" % e)
 
 
-def import_expr(file_handle):
+def import_expr(file_handle, organism_tax_id):
     """
     Function that reads input file and load gene-sample expression values
     into the database.
     """
+
+    # Make sure input organism_tax_id already exists in database.
+    try:
+        organism = Organism.objects.get(taxonomy_id=organism_tax_id)
+    except Organism.DoesNotExist:
+        raise Exception("Input organism_tax_id is not found in the database. "
+                        "Please use the management command "
+                        "'organism_create_or_update.py' in djago-organism "
+                        "package to create this organism.")
 
     # Enclose reading/importing process in a transaction context manager.
     # Any exception raised inside the manager will terminate the transaction
@@ -57,7 +84,7 @@ def import_expr(file_handle):
                 tokens = tokens[1:]
                 read_header(tokens, samples)
             else:
-                import_data_line(line_index + 1, tokens, samples)
+                import_data_line(line_index + 1, tokens, samples, organism)
 
 
 def read_header(tokens, samples):
@@ -67,7 +94,7 @@ def read_header(tokens, samples):
     using "ml_data_source" field. If a token does not match any sample's
     ml_data_source, put None into "samples".)
 
-    An exception will be raised if any of the following errors is detected:
+    An exception will be raised if any of the following errors are detected:
       * Sample token is blank (null or consists of space characters only);
       * Sample token is duplicate;
     """
@@ -92,10 +119,10 @@ def read_header(tokens, samples):
                     "in the database: %s", index + 2, data_source)
 
 
-def import_data_line(line_num, tokens, samples):
+def import_data_line(line_num, tokens, samples, organism):
     """
     Function that imports numerical values in input tokens into the database.
-    An exception will be raised if any of the following errors is detected:
+    An exception will be raised if any of the following errors are detected:
       * The number of columns on this line is not equal to the number of
         samples plus 1.
       * The gene's "systematic_name" field (column #1) is blank;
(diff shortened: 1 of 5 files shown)


dongbohu:
I also updated database schema at: 
https://sketchboard.me/rzMAA9ecMFEh#
The box whose border is red is the class that I added in `models.py`.

dongbohu:
The new issues in Codeclimate are probably due to the mixture of if/else and try/except.

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def read_header(tokens, samples):
+    """
+    Read input tokens on header line and save the corresponding sample
+    object into "samples". (Each token will be searched in the database
+    using "ml_data_source" field. If a token does not match any sample's
+    ml_data_source, put None into "samples".)
+
+    An exception will be raised if any of the following errors is detected:
```
dongbohu: Believe it or not, I originally used "are" but changed to "is" later. There are some interesting discussions on this topic:
https://www.englishforums.com/English/AnyIsAre/hxhhn/post.htm
Most people seem to think that both `are` and `is` are okay.

kurtwheeler: English is weird.

adage/analyze/models.py
```diff
@@ -218,3 +218,17 @@ def __unicode__(self):
...
+
+
+class ExpressionValue(models.Model):
+    """
+    This class models the many-to-many relationship between Sample and Gene.
+    Each sample/gene pair has an expression value of floating type.
+    """
+    sample = models.ForeignKey(Sample, on_delete=models.PROTECT)
+    gene = models.ForeignKey(Gene, on_delete=models.PROTECT)
+    value = models.FloatField()
```
rzelayafavila: Nice, neat model, and good, descriptive docstring too!

adage/analyze/models.py
```diff
@@ -218,3 +218,17 @@ def __unicode__(self):
...
+
+    def __unicode__(self):
+        return "Sample %s and Gene %s with expression value %f" % (
```
rzelayafavila: Hmm, I think I would phrase this more like "Expression value %f for Sample %s and Gene %s", so that the emphasis is on the fact that this is an Expression value. What do you think?

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -0,0 +1,140 @@
...
+        try:
+            float_val = float(value)
+        except ValueError:
+            raise Exception("Input file line #%d column #%d: %s is not numeric"
+                            % (line_num, col_num, value))
+        if sample is not None:
+            records.append(
+                ExpressionValue(sample=sample, gene=gene, value=float_val))
+        col_num += 1
+    ExpressionValue.objects.bulk_create(records)  # Create records in bulk.
```
rzelayafavila: :+1: for ``bulk_create``

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python
+
+"""
+This management command reads an input file of gene-sample expression values
+and loads the valid data into database.  It should be invoked like this:
```
rzelayafavila: Quick minor thing: I think this sentence would flow better if it was either "... and loads the valid data into **the** database" or "... and loads valid data into database." I kind of prefer this first one, though.

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -0,0 +1,140 @@
...
+        raise Exception("Input file line #%d: Number of columns is not %d" %
+                        (line_num, len(samples) + 1))
+
+    gene_name = tokens[0]
+    if not gene_name or gene_name.isspace():
+        raise Exception("Input file line #%d: gene systematic_name (column #1)"
+                        " is blank" % line_num)
+
+    try:
+        gene = Gene.objects.get(systematic_name=gene_name)
```
rzelayafavila: Hmmm... is this command agnostic to the organism/species? What if there are two genes with the same systematic_name from different species?

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -0,0 +1,140 @@
...
+    into the database.
+    """
+
+    # Enclose reading/importing process in a transaction context manager.
+    # Any exception raised inside the manager will terminate the transaction
+    # and roll back the database.
+    with transaction.atomic():
+        samples = []
+        for line_index, line in enumerate(file_handle):
+            tokens = line.rstrip('\r\n').split('\t')
```
rzelayafavila: Why are we specifying ``'\r\n'`` in ``rstrip()`` here? Is it so that rstrip will not remove single spaces and tabs?

adage/analyze/models.py
dongbohu: Sure. I like that.

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: Sure. I will change it to the first one.

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: I think it is also okay to simply use `rstrip()`, but I want to keep the trailing spaces and tabs (which will be detected later) so that the error will be "blank gene names" instead of "number of columns is not xxx". Does this make sense?

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: Good question. I never thought of this. Is it possible that two different organisms may have different genes whose systematic names are identical?

adage/analyze/management/commands/import_gene_sample_expr.py
mhuyck: The plural form is correct.

For such grammatical "edge cases," I usually turn to *The Elements of Style*, by William Strunk, Jr., and E.B. White, more commonly known as "Strunk & White." The relevant rule is described in Strunk & White in Section I, part 9: "The number of the subject determines the number of the verb." In this case, the part that's tricky is that the subject of the sentence is "any of."

The example given in the book is:

> A common blunder is the use of a singular verb form in a relative clause following "one of ..." or a similar expression when the relative is the subject.

and then two examples are given showing the wrong wording in a column on the left and the corrected wording in a column on the right. Rather than type them all out, the corrected first example says

> One of the ablest men who have attacked this problem

But we are not writing a thesis here, so I'm good with it either way. Just thought you'd like to know about Strunk & White. It is considered an authoritative source and a worthwhile companion for any writer of English.

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -0,0 +1,140 @@
...
+      * The number of columns on this line is not equal to the number of
+        samples plus 1.
+      * The gene's "systematic_name" field (column #1) is blank;
+      * Data field (from column #2 to the end) can not be converted into a
+        float type.
+    """
+
+    if len(tokens) != len(samples) + 1:
+        raise Exception("Input file line #%d: Number of columns is not %d" %
+                        (line_num, len(samples) + 1))
```
mhuyck: Thank you for reporting line numbers! I know you always do, but want to make sure you know that the effort is greatly appreciated. This type of error always seems to crop up when it's most helpful to have a quick solution.

adage/analyze/management/commands/import_gene_sample_expr.py
mhuyck: Yes, unfortunately it is. I think we may need to specify an organism on the command line. Biology is even more complicated than English. 😁

adage/analyze/models.py
tj8901nm: I don't think gene expression values have units. They are relative measurements. To make it a little more precise, you can say "RMA processed and zero-one normalized gene expression values".

mhuyck: Right, I guess if it's normalized then the units are taken out. Your description of the pre-processing is exactly what I was looking for.

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: Thanks. I will add an argument as input organism.

adage/analyze/models.py
dongbohu: Thank you both for the comments. I will change it.

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: Thank you for the recommendation of `The Elements of Style` book, it is definitely something good to know. But I am not 100% convinced that the rule in Section I, part 9 applies to the current scenario, because in "any of the following errors is/are detected", the subject is "any", which could refer to either singular or plural, while in:
```One of the ablest men who have attacked this problem```
"who" clearly refers to "men", which is plural. 
As you said, this is just a comment line and not that important, so I will change it into plural.

File: adage/analyze/management/commands/import_gene_sample_expr.py (hunk 7 above)
Comment (rzelayafavila) [inline adage/analyze/management/commands/import_gene_sample_expr.py]: Yep, that's what I imagined, sounds good.

**Comment by dongbohu** [comment]
I think I have addressed all concerns from the reviewers. I also added a `unique_together` restriction in the new model (and another migration file, which you can ignore). Please  take a look again and let me know what you think. Thanks.

File: adage/analyze/management/commands/import_gene_sample_expr.py [hunk 9]
```diff
@@ -2,19 +2,35 @@
...
+
+IMPORTANT:
+(1) Before running this command, please make sure that "django-organisms"
+package has been installed and organism_tax_id already exists in the database.
+If organism_tax_id is not in the database yet, please use the management
+command "organisms_create_or_update.py" (in "django-organisms" package)
+to add it.
+(2) If a data source (on the first row of input file) or gene name (on the
+first column of input file) is not found in the database, a warning message
+will be generated and the corresponding column or row will be skipped.
```
mhuyck: Thank you for the excellent documentation!

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -40,12 +58,21 @@ def handle(self, **options):
...
     """
 
+    # Make sure input organism_tax_id already exists in database.
+    try:
+        organism = Organism.objects.get(taxonomy_id=organism_tax_id)
+    except Organism.DoesNotExist:
+        raise Exception("Input organism_tax_id is not found in the database. "
+                        "Please use the management command "
+                        "'organism_create_or_update.py' in djago-organism "
+                        "package to create this organism.")
```
mhuyck: A user-friendly error message with concrete steps to solve the problem. Awesome! :100:

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -67,7 +94,7 @@ def read_header(tokens, samples):
     using "ml_data_source" field. If a token does not match any sample's
     ml_data_source, put None into "samples".)
 
-    An exception will be raised if any of the following errors is detected:
+    An exception will be raised if any of the following errors are detected:
```
mhuyck: 👍

adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -109,17 +136,21 @@ def import_data_line(line_num, tokens, samples):
...
+                        " multiple genes in the database" %
+                        (line_num, gene_name))
     except Gene.DoesNotExist:
         # If a gene is not found in database, generate a warning message
         # and skip this line.
         logger.warn(
-            "Input file line #%d: gene systematic_name (column #1) is not "
-            "found in the database: %s", line_num, gene_name)
+            "Input file line #%d: gene name %s (column #1) not found in "
+            "database", line_num, gene_name)
```
mhuyck: Maybe I'm missing something, but don't we need this to say:

```python
...database" % (line_num, gene_name))
```

(as in lines 145-147 above) and I think line 151 also needs to be `logger.warning(`, right?

mhuyck:
By the way, the `unique_together` constraint is a very good idea. Thanks for catching that, too!

adage/analyze/management/commands/import_gene_sample_expr.py
rzelayafavila: :100:

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: I checked python doc and seems that both formats of the arguments will work:
https://docs.python.org/2/library/logging.html#logging.Logger.debug
The arguments of `warning()` method is same as those in `debug()` method.
(I think I read an online article some time ago which argues that the format that I am using has better performance, but I couldn't find the link now).

You are definitely right about the method name `warning`. `warn()` is a method for `warnings` library.
https://docs.python.org/2/library/warnings.html#warnings.warn

Two other management commands `import_data.py` and `import_activity.py` are also using `warn()` instead of `warning()`, I will file another PR after this one to fix them.

dongbohu: Wait a minute: I just remembered that when I used `logger.warn()`, the warning messages are correctly printed out on the screen during importing process. Let me read the python doc more carefully.

adage/analyze/management/commands/import_gene_sample_expr.py
rzelayafavila: :+1:

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: `warn()` seems to be a "shortcut" of  `warning()` in `logger` class, but I don't understand why python online doc doesn't mention the former.

dongbohu: I am going to change `warn` to `warning` in this file to avoid confusion in the future.

adage/analyze/management/commands/import_gene_sample_expr.py
dongbohu: Just noticed the typo: `djago-organism` should be `django-organisms`. I will fix it.

dongbohu: Found a similar typo in `add_ml_model.py`, fixed it too and included it in this PR.

adage/analyze/management/commands/import_gene_sample_expr.py
mhuyck: Sounds good. If it's not documented it might go away in the future without `warn`ing. (Heh.)

mhuyck:
Looks great. Thank you @dongbohu!

PR #106, opened 2017-04-04
Changing name of file with Node page unit tests to node.spec.js, and adding a couple of unit tests for the high range experiments directive.

rzelayafavila:


interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+  });
+
+  afterEach(function() {
+    $httpBackend.verifyNoOutstandingExpectation();
+    $httpBackend.verifyNoOutstandingRequest();
+  });
+});
+
+
+describe('<high-range-exp> directive', function() {
```
rzelayafavila: This is where the test code for the high-range-exp directive starts. The code above this should have no changes from what @dongbohu wrote.

rzelayafavila:
Oops, I'm going to fix that indentation issue that codeclimate is complaining about.

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+  it('should render HTML correctly', function() {
+    var parentScope = $rootScope.$new();
+    parentScope.nodeID = 123;
+    parentScope.topExp = 20;
+    var testHTML = '<high-range-exp node-id="{{nodeID}}" ' +
+        ' top-exp="{{topExp}}"></high-range-exp>';
+
+
+    var uri1 = '/api/v0/activity/?limit=0&node=' + parentScope.nodeID;
+    var uri2 = '/api/v0/experiment/?limit=0&node=' + parentScope.nodeID;
```
mhuyck: Quibble: could `uri1` and `uri2` have more descriptive names? Looks like `uri1` could be called `uriActivity` and `uri2` could be `uriExperiment`.

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+      'name': 'Sample experiment name',
+      'resource_uri': '/api/v0/experiment/E-GEOD-00001/',
+      'sample_set': ['/api/v0/sample/1/', '/api/v0/sample/2/']
+    }];
+
+    $httpBackend.expectGET(uri1).respond({objects: mockActivity});
+    $httpBackend.expectGET(uri2).respond({objects: mockExperiment});
+    parentScope.$digest(); // Start Angular's digest cycle manually.
+
+    // Get isolated scope of <high-range-exp> directive:
```
mhuyck: typo: should say "isolate scope" (as it does in the code below)

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+        'systematic_name': 'g3_sys',
+        'standard_name': 'g3_std',
+        'description': 'g3_desc'
+      }}
+    ];
+
+    $httpBackend.expectGET(uri).respond({objects: mockGenes});
+    parentScope.$digest(); // Start Angular's digest cycle manually.
+
+    // Get isolated scope of <high-weight-genes> directive:
```
mhuyck: typo: should say "isolate scope"

interface/src/app/node/node.spec.js
mhuyck: Thank you for this  note, by the way.

interface/src/app/node/node.spec.js
dongbohu: Good catch. I didn't realize that `isolate` is an adjective too.

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
+/**
+ * 'adage.node' module unit tests
+ *
+ * Test <high-weight-genes> directive.
```
dongbohu: Since this file tests other directives too, you may want to put this commend line in a separate section right before line 6 (instead of the comments for the whole file).

interface/src/app/node/node.spec.js
dongbohu: Maybe add a comment line such as `Test <high-range-exp> directive`?

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+
+    // Make backend data available to the directive.
+    $httpBackend.flush();
+
+    // Confirm that properties in directive are what we expect them to be
+    // The queryStatus should now be an empty string.
+    expect(elementScope.queryStatus).toBe('');
+  });
+
+  it('should get the appropriate activity values', function() {
```
dongbohu: I noticed that a few variables (`uri1`, `uri2`, `testHTML`, `element`, `elementScope`, etc) in this test case are set to same values as those in the first case (line 102). You can save a few lines by setting them once in `beforeEach()`. If you plan to keep the two cases completely independent from each other, then you can leave the code as is.

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+      'name': 'Sample experiment name',
+      'resource_uri': '/api/v0/experiment/E-GEOD-00001/',
+      'sample_set': ['/api/v0/sample/1/', '/api/v0/sample/2/']
+    }];
+
+    $httpBackend.expectGET(uri1).respond({objects: mockActivity});
+    $httpBackend.expectGET(uri2).respond({objects: mockExperiment});
+    parentScope.$digest(); // Start Angular's digest cycle manually.
+
+    // Get isolated scope of <high-range-exp> directive:
```
dongbohu: As @mhuyck pointed out, it should be `isolate`. My fault too.

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,246 @@
...
+    var elementScope = element.isolateScope();
+
+
```
dongbohu: Is this extra blank line added on purpose?

interface/src/app/node/node.spec.js
rzelayafavila: Ah, good catch.

interface/src/app/node/node.spec.js
rzelayafavila: Good call, they should be more descriptive. I will change this.

rzelayafavila:
Ok @mhuyck, @dongbohu I think I got everything - let me know how it looks and if I missed anything! Thanks!

mhuyck:
Looks great. Thank you!

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,234 @@
...
+      {id: 16, node: 123, sample: 1016, value: 0.0743736434104904},
+      {id: 17, node: 123, sample: 1017, value: 0.07871175918013},
+      {id: 18, node: 123, sample: 1018, value: 0.0614160606502279},
+      {id: 19, node: 123, sample: 1019, value: 0.0661371872868119},
+      {id: 20, node: 123, sample: 1020, value: 0.0783672660119669},
+      {id: 21, node: 123, sample: 1021, value: 0.0769023791604038}
+    ];
+
+    // Mocked experiment response data:
+    var mockExperiment = [{
```
dongbohu: `mockExperiment` is identical in two test cases. Should we set it once in `beforeEach()` too?

interface/src/app/node/node.spec.js
```diff
@@ -0,0 +1,234 @@
...
+      'resource_uri': '/api/v0/experiment/E-GEOD-00001/',
+      'sample_set': ['/api/v0/sample/1/', '/api/v0/sample/2/']
+    }];
+
+    $httpBackend.expectGET(activityUri).respond({objects: mockActivity});
+    $httpBackend.expectGET(experimentUri).respond({objects: mockExperiment});
+    parentScope.$digest(); // Start Angular's digest cycle manually.
+
+    // Get isolate scope of <high-range-exp> directive:
+    var elementScope = element.isolateScope();
```
dongbohu: Can we set `elementScope` in `beforeEach()`?

interface/src/app/node/node.spec.js
rzelayafavila: Yeah sure, good catch!

interface/src/app/node/node.spec.js
rzelayafavila: So it looks like something like mockExperiment can, but not elementScope, since it needs to have ``parentScope.$digest()`` be called before it. However, this ``parentScope.$digest()`` should not be called before the ``$httpBackend.expect``s, or it will run the digest and get the http calls without expecting them. At least this is how I understand it is working, so I think we are good leaving it in each `it()`` unit test. Does that make sense?

dongbohu: Yeah, I think you are right. I forgot that the scope variable is special. Thanks for the explanation.

interface/src/app/node/node.spec.js
dongbohu: Please go ahead closing this PR after moving the experiment setup into `beforeEach()`.

rzelayafavila:
Ok thank you @dongbohu and @mhuyck!! Merging this.

PR #107, opened 2017-04-06
Volcano plot selection extraction

mhuyck:
Adds implementation required to extract node selection from vega and a button to invoke the gene network with genes associated with the selected nodes.

### DIFF
--- a/interface/src/app/volcano-plot/view.js
+++ b/interface/src/app/volcano-plot/view.js
@@ -28,10 +28,13 @@ angular.module('adage.volcano-plot.view', [
   //      for sample-group-a and sample-group-b and does what's necessary to
   //      make a plot from those lists)
   function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
+    var ctrl = this;
     SampleBin.getVolcanoPlotData();
     this.data = SampleBin.volcanoData;
-    // FIXME need to dig into vega internal data format to make selection work
     this.selection = [];
+    this.updateSelection = function updateSelection(selectedNodes) {
+      ctrl.selection = selectedNodes;
+    };
   }
 ])
 ;
--- a/interface/src/app/volcano-plot/view.js
+++ b/interface/src/app/volcano-plot/view.js
@@ -30,7 +30,7 @@ angular.module('adage.volcano-plot.view', [
   function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
     var ctrl = this;
     console.log('VolcanoPlotView $stateParams:', $stateParams);
-    this.mlmodel = $stateParams.mlmodel;
+    this.mlModel = $stateParams.mlmodel;
     SampleBin.getVolcanoPlotData();
     this.sampleGroups = SampleBin.getSamplesByGroup();
     this.data = SampleBin.volcanoData;
--- a/interface/src/app/volcano-plot/view.js
+++ b/interface/src/app/volcano-plot/view.js
@@ -11,7 +11,7 @@ angular.module('adage.volcano-plot.view', [
 
 .config(['$stateProvider', function($stateProvider) {
   $stateProvider.state('volcano', {
-    url: '/volcano',
+    url: '/volcano?mlmodel',
     views: {
       main: {
         templateUrl: 'volcano-plot/view.tpl.html',
@@ -29,6 +29,8 @@ angular.module('adage.volcano-plot.view', [
   //      to make a plot from those lists)
   function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
     var ctrl = this;
+    console.log('VolcanoPlotView $stateParams:', $stateParams);
+    this.mlmodel = $stateParams.mlmodel;
     SampleBin.getVolcanoPlotData();
     this.sampleGroups = SampleBin.getSamplesByGroup();
     this.data = SampleBin.volcanoData;
--- a/interface/src/app/volcano-plot/view.js
+++ b/interface/src/app/volcano-plot/view.js
@@ -30,6 +30,7 @@ angular.module('adage.volcano-plot.view', [
   function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
     var ctrl = this;
     SampleBin.getVolcanoPlotData();
+    this.sampleGroups = SampleBin.getSamplesByGroup();
     this.data = SampleBin.volcanoData;
     this.selection = [];
     this.updateSelection = function updateSelection(selectedNodes) {
--- a/interface/src/app/volcano-plot/view.js
+++ b/interface/src/app/volcano-plot/view.js
@@ -25,8 +25,8 @@ angular.module('adage.volcano-plot.view', [
 .controller('VolcanoPlotViewCtrl', ['SampleBin', '$stateParams',
   // TODO make use of $stateParams (pulling from SampleBin for initial tests,
   //      but the right way to do this is to refactor so view.js pulls params
-  //      for sample-group-a and sample-group-b and does what's necessary to
-  //      make a plot from those lists)
+  //      for sample-base-group and sample-comp-group and does what's necessary
+  //      to make a plot from those lists)
   function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
     var ctrl = this;
     SampleBin.getVolcanoPlotData();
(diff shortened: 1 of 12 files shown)


mhuyck:
FYI: working on a demo site... sorry I forgot to get that going earlier.

mhuyck:
Test server now up and running at http://greenelab.fgtech.com/

I have not adjusted the timeouts for `gunicorn` or `nginx` yet (my temporary workaround), so for now just limit the number of nodes you select in the volcano plot and you will be able to (eventually) see a gene network.

interface/src/app/volcano-plot/volcano-plot.js
```diff
@@ -4,50 +4,134 @@
...
+// TODO when refactoring for issue #76, this service belongs in the top-level
+//      src/app/ module called `node` in its `services.js` file
+.factory('NodeService', ['$resource', 'ApiBasePath',
+  function($resource, ApiBasePath) {
+    // private methods for NodeService
+    var Participation = $resource(ApiBasePath + 'participation/');
+
+    // public methods for NodeService
+    return {
+      getGenesForNodesPromise: function getGenesForNodesPromise(nodes) {
```
dongbohu: As I said earlier, the second `getGenesForNodesPromise` can be removed to make the line shorter. But if you prefer giving it a name, it is totally fine with me.

interface/src/app/volcano-plot/volcano-plot.js
```diff
@@ -4,50 +4,134 @@
...
-        //   {
-        //     'node': "<node-name-2>"
-        //   },
-        //   {...repeat for each node...}
-        // ]
-    },
-    templateUrl: 'volcano-plot/volcano-plot.tpl.html',
-    controller: 'VolcanoPlotCtrl as vp'
-  };
+.component('volcanoPlot', {
```
dongbohu: Nice conversion to `component`!

interface/src/app/volcano-plot/volcano-plot.js
```diff
@@ -4,50 +4,134 @@
...
+      var $ctrl = this;
+      this.showNetwork = function showNetwork() {
+        // genes, mlmodel, base_group, comp_group
+        var nodeIds = this.selectedNodes.map(function nodesToIds(nodeObj) {
+          return nodeObj.id;
+        });
+        NodeService.getGenesForNodesPromise(nodeIds).then(
+          function success(genes) {
+            $location.path('/gene_network/').search({
+              'mlmodel': 1,
```
dongbohu: Should we set the value of `mlmodel` to be a parameter instead of hard-coded `1`?

interface/src/app/volcano-plot/volcano-plot.js
rzelayafavila: Good question. Same as with the other named function in ``volcano-plot/view.js ``, I have a hard time deciding if I should name these functions or not.

interface/src/app/volcano-plot/volcano-plot.js
rzelayafavila: Agree! :+1:

interface/src/app/volcano-plot/volcano-plot.js
rzelayafavila: Agree that this should be a parameter.

interface/src/less/variables.less
```diff
@@ -15,5 +15,5 @@
  */
 
 // used for highlighting sample groups in sampleBin.tpl.html
-@sample-group-a: #b7baff;
-@sample-group-b: #fee7a9;
+@sample-base-group: #b7baff;
```
rzelayafavila: :+1: for our new names!

interface/src/app/volcano-plot/volcano-plot.js
```diff
@@ -4,50 +4,134 @@
...
+
+.component('volcanoPlotSelection', {
+  bindings: {
+    selectedNodes: '<',
+    sampleGroups: '<'   // network view needs this context for annotation
+  },
+  templateUrl: 'volcano-plot/volcano-plot-selection.tpl.html',
+  controller: ['$log', '$location', 'errGen', 'NodeService',
+    function VolcanoPlotSelectionCtrl($log, $location, errGen, NodeService) {
+      var $ctrl = this;
```
rzelayafavila: Ah, so we have ``var $ctrl`` instead of ``var ctrl`` or ``var self`` here to refer to the controller instance? Is there any danger with using a variable that starts with ``$`` as most of the variables/objects that start with ``$`` belong to Angular? By danger here I also mean potential of misleading whoever is reading the code into thinking that this is an Angular object.

interface/src/app/volcano-plot/volcano-plot.js
mhuyck: I had read somewhere that naming functions helps when you get a stack trace. I couldn't remember where I read that so I Googled around and found this blog post on the subject:
> http://www.therightcode.net/naming-functions-yes-but-not-all-the-time/
The logic there makes perfect sense: use named functions when it would help for stack traces, but don't do it in cases (such as these you've highlighted) where it doesn't actually help and causes redundancy in the code.

Will fix.

- [ ] don't use named functions on objects that already "name" them via a method

interface/src/app/volcano-plot/volcano-plot.js
mhuyck: I'm really liking the simplification provided by `.component`. It's easy to see why this is "the way of the future" for Angular. 😎

interface/src/app/volcano-plot/volcano-plot.js
mhuyck: Aha. Good catch! I had started to implement a full selector for MLModel as we have elsewhere, but realized the right place for that was back in the heatmap so not in scope for this pull request.

Although I have more work to do to parametrize the volcano plot, this could certainly be the first part. Will fix.

- [ ] make MLModel into a parameter for the page

interface/src/app/volcano-plot/volcano-plot.js
mhuyck: Hmm. Good question. What I need here is a closure variable to use for the callback from `getGenesForNodesPromise` below. Since this is a `.component`, the `$ctrl` gets instantiated automatically for the template context but not here inside the controller itself. I went ahead and used `$ctrl` for consistency, but if you prefer I could use something like `cbVolcanoPlotSelection`, following the naming I've used for callback closure variables elsewhere.

@rzelayafavila, @dongbohu: What approach do you two think would be better?

dongbohu: Since $ctrl has been used by angular internally, I would avoid the name with leading `$`.

mhuyck: Unless @rzelayafavila objects, then, I will:

- [ ] use `cbVolcanoPlotSelection` instead of `$ctrl` for callback closure variable

rzelayafavila: Sure, that sounds good to me, I do think it is better to avoid the leading ``$``.

mhuyck:
Okay, @rzelayafavila & @dongbohu, I've addressed everything now I think. Please let me know if you have any other comments.

interface/src/app/volcano-plot/view.js
```diff
@@ -25,13 +25,19 @@ angular.module('adage.volcano-plot.view', [
 .controller('VolcanoPlotViewCtrl', ['SampleBin', '$stateParams',
   // TODO make use of $stateParams (pulling from SampleBin for initial tests,
   //      but the right way to do this is to refactor so view.js pulls params
-  //      for sample-group-a and sample-group-b and does what's necessary to
-  //      make a plot from those lists)
+  //      for sample-base-group and sample-comp-group and does what's necessary
+  //      to make a plot from those lists)
   function VolcanoPlotViewCtrl(SampleBin, $stateParams) {
+    var ctrl = this;
+    console.log('VolcanoPlotView $stateParams:', $stateParams);
```
dongbohu: This seems to be a debug line. Do you keep it on purpose?

mhuyck: Oops! Yes, that is how I debugged the problem I was having. No, I did not mean to leave it there. Thank you for catching it!

PR #108, opened 2017-04-12
Use FDR to adjust raw p-values for the volcano plot

mhuyck:


interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -424,12 +425,25 @@ MathFuncts, errGen) {
...
         });
+
+        // use FDR on the raw p-values from nodeInfoSet to get adjustedPValues
+        var rawPValues = nodeInfoSet.map(function getRawPValue(nodeObject) {
+          return nodeObject.rawPValue;
+        });
+        var adjustedPValues = MathFuncts.multTest.fdr(rawPValues);
+
+        // map the adjustedPValues back into the nodeInfoSet
+        nodeInfoSet = nodeInfoSet.map(function(nodeObject, i) {
```
dongbohu: Since `nodeInfoSet` is modified in place, do you think it is better to use `forEach()` instead of `map()` here?

mhuyck: Aha. A great question. I feel like I was avoiding `forEach()` for some reason... let me see.

By the way, there's nothing wrong with being opinionated! It's being inflexible *and* opinionated where the trouble starts. 😄

mhuyck: I have no idea why I would avoid `forEach()`, actually. I think you're right that it would be more efficient because we can modify the elements of the array in place rather than copying into a new array. Will fix.

- [ ] use `.forEach()` instead of `.map()` to modify array in place

mhuyck:
Okay, updated now. Let me know if you spot anything else I should address.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -424,12 +425,25 @@ MathFuncts, errGen) {
...
+        // use FDR on the raw p-values from nodeInfoSet to get adjustedPValues
+        var rawPValues = nodeInfoSet.map(function getRawPValue(nodeObject) {
+          return nodeObject.rawPValue;
+        });
+        var adjustedPValues = MathFuncts.multTest.fdr(rawPValues);
+
+        // compute logsig from the adjustedPValues & update the nodeInfoSet
+        nodeInfoSet.forEach(function(nodeObject, i) {
+          nodeObject.logsig = -Math.log10(adjustedPValues[i]);
+          nodeInfoSet[i] = nodeObject;
```
dongbohu: This line seems redundant since `nodeInfoObject` is a reference to the original object in the array `nodeInfoSet`, right?

This is trivial, so please feel free to close this PR any time you feel comfortable.

dongbohu: If you remove this line, then the argument `i` of the function inside `forEach` can be removed too.

mhuyck: Aha. Good point! Modify in place really means *in place*. I've removed line 444 since it's redundant.

I can't remove the `i` from the callback function. I'm using it to pull the right element from the parallel array `adjustedPValues` at line 443.

dongbohu: You are right. I forgot that. After you remove line 444, please feel free to merge it. Thanks.

PR #109, opened 2017-04-12
Add ExpressionValueResource

dongbohu:
This PR defines the new resource class for `ExpressionValue` model in `api.py` and has it registered in `urls.py`.

### DIFF
--- a/adage/analyze/api.py
+++ b/adage/analyze/api.py
@@ -13,7 +13,7 @@ from haystack.query import SearchQuerySet
 from haystack.inputs import AutoQuery
 from models import (
     Experiment, Sample, SampleAnnotation, AnnotationType, MLModel, Node,
-    Activity, Edge, Participation
+    Activity, Edge, Participation, ExpressionValue
 )
 
 # Many helpful hints for this implementation came from:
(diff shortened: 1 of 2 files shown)


adage/analyze/api.py
```diff
@@ -423,3 +423,21 @@ class Meta:
             'node': ('exact', 'in', ),
             'gene': ('exact', 'in', ),
         }
+
+
+class ExpressionValueResource(ModelResource):
+    gene = fields.IntegerField(attribute='gene_id', null=False)
+    sample = fields.IntegerField(attribute='sample_id', null=False)
+
```
rzelayafavila: So I guess we don't need to add the ``value`` field her, right? Tastypie will just add it automatically and we don't need to worry about converting from ForeignKey object to IntegerField like in the other fields, right?

rzelayafavila: Ok, great, LGTM!

adage/analyze/api.py
```diff
@@ -423,3 +423,21 @@ class Meta:
             'node': ('exact', 'in', ),
             'gene': ('exact', 'in', ),
         }
+
+
+class ExpressionValueResource(ModelResource):
+    gene = fields.IntegerField(attribute='gene_id', null=False)
+    sample = fields.IntegerField(attribute='sample_id', null=False)
```
mhuyck: 👍 on avoiding a join. I think we've learned our lesson on that! 😄

mhuyck:
I know it's already merged, but LGTM also.

PR #110, opened 2017-04-13
Docstring error corrected in import_gene_sample_expr.py

dongbohu:
I just realized that I missed the management command name in the docstring of `import_gene_sample_expr.py`.

### DIFF
--- a/adage/analyze/management/commands/import_gene_sample_expr.py
+++ b/adage/analyze/management/commands/import_gene_sample_expr.py
@@ -13,7 +13,7 @@ The two required arguments are:
 
 For example, to load the sample-gene expression values for the organism
 "Pseudomonas aeruginosa" (whose taxonomy ID is 208964), the command will be:
-  python manage.py input_filename 208964
+  python manage.py import_gene_sample_expr input_filename 208964
 
 IMPORTANT:
 (1) Before running this command, please make sure that "django-organisms"


adage/analyze/management/commands/import_gene_sample_expr.py
```diff
@@ -13,7 +13,7 @@
 
 For example, to load the sample-gene expression values for the organism
 "Pseudomonas aeruginosa" (whose taxonomy ID is 208964), the command will be:
-  python manage.py input_filename 208964
+  python manage.py import_gene_sample_expr input_filename 208964
```
rzelayafavila: Is the ``input_filename`` file always supposed to be a certain type of file (like ``.tsv``, tab-separated values)? If so, would it be helpful to give it a ``.tsv`` ending (so making it ``input_filename.tsv``)? What type of file is the expression_filename that we use now?

dongbohu: The input file that Jie sent to me is `data/all-pseudomonas-gene-normalized.pcl`. Although it is tab delimited, it doesn't have "tsv" extension, so I didn't put any extension in this example command either.

rzelayafavila: Gotcha, ok, LGTM!

PR #112, opened 2017-04-17
Network fetch optimization

mhuyck:
This is a minimal set of changes that boosts the performance of fetching gene network data. A sample query to fetch edges with genes to draw a network done the original way took ~30 seconds. The same query now takes <2 seconds.


mhuyck:
Please see git commit log comments for details on how this works.

PR #113, opened 2017-04-19
New feature in gene-gene Network: Coloring the gene circles

dongbohu:
This PR adds the functionality of coloring the gene circles in the network. It also integrates the code in PR #112 by @mhuyck.

Coding style may not be the best due to the upcoming deadline. For example, in lines 356-465 in `network.js`, it would be better to chain the three API calls instead of nesting them. I will address them after the deadline.

This PR also uses my own d3.network.js instead of Aaron Wang's.

Demo page:
http://165.123.67.202/#/gene_network/?genes=6547&base_group=123,124&comp_group=125,126

### DIFF
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -7,12 +7,13 @@ angular.module('adage.gene.network', [
   'ui.bootstrap',
   'ngResource',
   'rzModule',
-  'adage.utils'
+  'adage.utils',
+  'adage.gene.resource'
 ])
 
 .config(['$stateProvider', function($stateProvider) {
   $stateProvider.state('gene_network', {
-    url: '/gene_network/?genes&mlmodel',
+    url: '/gene_network/?genes&mlmodel&base_group&comp_group',
     views: {
       main: {
         templateUrl: 'gene/network/network.tpl.html',
@@ -40,7 +41,7 @@ angular.module('adage.gene.network', [
   });
 }])
 
-.factory('EdgeService', ['$resource', 'ApiBasePath',
+.factory('Edge', ['$resource', 'ApiBasePath',
   function($resource, ApiBasePath) {
   // Possible parameters for this endpoint when making a query are:
   // {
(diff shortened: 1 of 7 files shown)


dongbohu:
My own d3.network.js is here:
https://github.com/dongbohu/d3-network/blob/master/d3.network.js

interface/src/app/gene/network/network.less
```diff
@@ -5,7 +5,7 @@ line.edge {
 circle.gene {
   stroke-width: 2px;
   stroke: #AAA;
-  fill: #FFF;
+  //fill: #FFF;
```
rzelayafavila: Hmmm, do we still want to keep this as a comment? How come?

interface/src/app/gene/network/network.js
```diff
@@ -40,7 +41,7 @@ angular.module('adage.gene.network', [
   });
 }])
 
-.factory('EdgeService', ['$resource', 'ApiBasePath',
+.factory('Edge', ['$resource', 'ApiBasePath',
```
dongbohu: I feel that having "Service" at the end of service name is redundant. I saw the usage of both. I don't know which one is better, but I prefer a shorter name.

interface/src/app/gene/network/network.js
```diff
@@ -126,52 +140,85 @@ angular.module('adage.gene.network', [
...
-        if (idx === -1) {
-          if (urlGenes.indexOf(geneIn.id.toString()) !== -1) {
-            geneIn.query = true;
+      // Function that returns the list of genes that will be queried.
+      var getQueriedGenes = function(edgeList) {
+        // The genes that will be queried on the server should be determined
+        // by both genesInURL and edgeList, because even if a gene in the URL
+        // doesn't have any edge in the database, it still should be displayed
+        // in the network.
+        var geneList = genesInURL.slice(0); // Copy all genes ID in the URL.
```
dongbohu: Sure. Thanks.

interface/src/app/gene/network/network.less
dongbohu: I commented this line out because the color of gene circles (including the default color) now is controlled completely by d3.network.js. So this color is hard coded. I may need to figure out a better way in the future.

PR #114, opened 2017-04-21
fabfile: Add new management command that loads express values into DB

dongbohu:


mhuyck:
This looks okay to me, but do you see why CircleCI is failing?

dongbohu:
My bad. I fixed it.

Issue #71, opened 2017-02-14
Gene Network view needs a "home" in the nav bar


mhuyck:
Searching on a gene in "GeneNetwork" and clicking the "View Genes as Network" button produces a page (`gene_network`) with none of the nav bar elements selected. This should not happen. Every view on the site should have a "home" so the user can develop a "sense of place" about where things live on the site. In this case, the "GeneNetwork" tab should probably stay highlighted.

One way to help address this would be to keep a small representation of what was selected on the main entry (`gene_search`) view drawn at the top of the `gene_network` page. This should take up very little room to allow free exploration of the network using as much of the window as possible, but it would also be helpful to retain a bit of the context of what is being viewed here. Some web sites have implemented this as "breadcrumbs," a feature which is available in Bootstrap, but it's not clear that's exactly the right approach here -- I mention it mainly to explain what I'm thinking. The implementation details should be spearheaded by whomever decides to tackle this issue.

dongbohu:
Fixed in PR #116.

PR #116, opened 2017-04-26
Fix of issue #71: Highlight "GeneNetwork" tab when network is shown

dongbohu:
This is a fix of issue #71. 

There are a few ways to solve this problem. The one I chose is the simplest but may not be the best. According to the ui-router doc, the best seems to be create a parent state of `gene`, then make `gene_search` and `gene_network` the child states. But that will change at least three files, so I chose the current approach. I think it is a good hack to know.

### DIFF
--- a/interface/src/app/app.js
+++ b/interface/src/app/app.js
@@ -29,8 +29,9 @@ angular.module('adage', [
   $resourceProvider.defaults.stripTrailingSlashes = false;
 }])
 
-.controller('AppCtrl', ['$scope', 'UserFactory',
-  function AppCtrl($scope, UserFactory) {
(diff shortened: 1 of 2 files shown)


rzelayafavila:
Great! Would you have a url to a live version of this to check out in a web browser?

dongbohu:
I found a bug in my fix. Let me work on it now.

dongbohu:
I will let you guys know when it is ready.

dongbohu:
I fixed the issue. Here is a demo site:
http://localhost/#/gene_network/?genes=10155
You can click other tabs (such as `Home`, `About`, etc) to test the highlight.

Issue #80, opened 2017-02-22
app/tribe_client/tribe_client.js


dongbohu:
```
...
    getPromise: function() {
      if (!promise) {
        promise = User.get({}, function(data) {  // line 18
          if (data.meta.total_count !== 0) {
            user = data.objects[0];
          }
        });  // line 22
      }
      return promise;
    },
...
```
On line 18, `User.get(...)` actually returns an instance of Resource class, not a promise. So it seems to make sense to change line 22 into:
```}).$promise;```

All the other places that uses
```UserFactory.getPromise().$promise```
can be simplified into:
```UserFactory.getPromise()```

Reference:
https://docs.angularjs.org/api/ngResource/service/$resource

rzelayafavila:
Ah true, good catch @dongbohu! This makes sense.

rzelayafavila:
Fixed with PR #117

PR #117, opened 2017-04-27
Fix of issue #80: Changes to UserFactory's getPromise function so it actually returns an http promise.

rzelayafavila:
This is a fix for issue #80

### DIFF
--- a/interface/src/app/tribe_client/tribe_client.js
+++ b/interface/src/app/tribe_client/tribe_client.js
@@ -10,32 +10,27 @@ angular.module('adage.tribe_client', [
 
   return {
     getUser: function() {
-      // only call this when the promise completes
+      // Only call this function after the promise completes
+      // to be able to get an actual user (instead of null).
       return user;
     },
+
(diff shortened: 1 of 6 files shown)


interface/src/app/tribe_client/tribe_client.js
```diff
@@ -10,32 +10,27 @@ angular.module('adage.tribe_client', [
...
+        }).$promise;
       }
       return promise;
     },
+
     resetPromise: function() {
       // reset the promise in case we need to check the user again
       promise = null;
     },
-    setUser: function(newUser) {
```
dongbohu: So the method `setUser()` is removed because it is not being used anywhere?

rzelayafavila: Exactly - yes, sorry, I should have mentioned this in the PR comments. This method was not being used anywhere and in fact I was thinking that it might not be the right way to go about implementing that functionality, even if we ever wanted it.

PR #139, opened 2017-05-03
New module to search a Signature (node): Issue #111 

dongbohu:
This PR implements the new feature requested in issue #111.

### DIFF
--- a/interface/src/app/node/node_search.js
+++ b/interface/src/app/node/node_search.js
@@ -1,0 +1,64 @@
+/**
+ * "adage.node_search" module.
+ */
+
+angular.module('adage.node_search', [
+  'ui.router',
+  'ui.bootstrap',
+  'ngResource',
+  'adage.mlmodel.components',
+  'adage.utils'
+])
+
+.config(['$stateProvider', function($stateProvider) {
+  $stateProvider.state('node_search', {
+    url: '/node/search',
+    views: {
+      main: {
+        templateUrl: 'node/node_search.tpl.html',
+        controller: 'NodeSearchCtrl'
+        // See comments below on why "controllerAs" is not used.
+      }
+    },
+    data: {pageTitle: 'Node Search'}
+  });
+}])
+
+.factory('Node', ['$resource', 'ApiBasePath',
+  function($resource, ApiBasePath) {
+    return $resource(ApiBasePath + 'node');
+  }
+])
+
+// "controllerAs" syntax is not used here because we are using "$scope" to
+// watch the mlModel selected by user, and a combination of both seems weird.
+.controller('NodeSearchCtrl', ['$scope', 'Node', '$log', 'errGen',
+  function NodeController($scope, Node, $log, errGen) {
+    $scope.statusMessage = 'Machine learning model not selected yet.';
+    $scope.mlModel = null;
+    $scope.nameFilter = '';
+
+    $scope.$watch('mlModel', function() {
+      if ($scope.mlModel === null) { // Skip the initialization watch
+        return;
+      }
+      $scope.statusMessage = 'Connecting to the server ...';
+      var modelID = $scope.mlModel.id;
+      Node.get(
+        {mlmodel: modelID, limit: 0},
+        function success(response) {
+          $scope.nodes = response.objects;
+          $scope.nodes.sort(function(n1, n2) { // Sort nodes by name
+            return (n1.name < n2.name) ? -1 : (n1.name === n2.name) ? 0 : 1;
+          });
+          $scope.statusMessage = '';
+        },
+        function error(err) {
+          var message = errGen('Failed to get nodes: ', err);
+          $log.error(message);
+          $scope.statusMessage = message + '. Please try again later.';
+        }
+      );
+    });
+  }
+]);
--- a/interface/src/app/node/node_search.js
+++ b/interface/src/app/node/node_search.js
@@ -2,12 +2,11 @@
  * "adage.node_search" module.
  */
 
-angular.module('adage.node_search', [
+angular.module('adage.nodeSearch', [
   'ui.router',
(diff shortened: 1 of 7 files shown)


dongbohu:
Live demo: http://165.123.67.202/#/node/search

You can ignore the two new issues reported by codeclimate.

interface/src/app/node/node_search.js
```diff
@@ -0,0 +1,64 @@
...
+        templateUrl: 'node/node_search.tpl.html',
+        controller: 'NodeSearchCtrl'
+        // See comments below on why "controllerAs" is not used.
+      }
+    },
+    data: {pageTitle: 'Node Search'}
+  });
+}])
+
+.factory('Node', ['$resource', 'ApiBasePath',
```
rzelayafavila: I think there is some repetition between the code here and the 'NodeInfo' factory that already exists in the ``node.js`` file. Should we consolidate them into one?

interface/src/app/node/node_search.js
```diff
@@ -0,0 +1,64 @@
...
+  'ui.router',
+  'ui.bootstrap',
+  'ngResource',
+  'adage.mlmodel.components',
+  'adage.utils'
+])
+
+.config(['$stateProvider', function($stateProvider) {
+  $stateProvider.state('node_search', {
+    url: '/node/search',
```
rzelayafavila: Should this 'node_search' state be a child state of the 'node' state that already exists? What are the advantages/disadvantages?

interface/src/app/node/node_search.js
```diff
@@ -0,0 +1,64 @@
...
+        return;
+      }
+      $scope.statusMessage = 'Connecting to the server ...';
+      var modelID = $scope.mlModel.id;
+      Node.get(
+        {mlmodel: modelID, limit: 0},
+        function success(response) {
+          $scope.nodes = response.objects;
+          $scope.nodes.sort(function(n1, n2) { // Sort nodes by name
+            return (n1.name < n2.name) ? -1 : (n1.name === n2.name) ? 0 : 1;
```
rzelayafavila: A lot seems to be going on in this line, since we are using two ternary operators - I would suggest either breaking it up or adding a comment above it explain what the logic is.

interface/src/index.html
```diff
@@ -82,6 +82,11 @@
                 GeneNetwork
               </a>
             </li>
+            <li ng-class="{active: inNodeStates()}">
+              <a ui-sref="node_search">
+                Signature
```
rzelayafavila: :+1:

interface/src/app/node/node_search.js
dongbohu: My original plan was to create an abstract state as parent and two child states for node information and node search. But I gave up for two reasons:

1) The two states "node" and "node_search" do not really share anything, so a parent state doesn't seem to be helpful. (I feel that their relationship is similar to the one between `gene_search` and `gene_network`.)

2) The file `node.js` has been pretty long now (475 lines). Adding the new `node_search` state to it will make the file longer. Keeping them separate will make it easier to maintain.

Does this make sense?

interface/src/app/node/node_search.js
dongbohu: It is true. But the URLs of the two resource classes are slightly different, so I didn't bother.

interface/src/app/node/node_search.js
dongbohu: Sure. (Back to the good old time when I was a C/C++ programmer, I used this operator a lot. You probably can still see its popularity from today's linux kernel source code.)

interface/src/app/node/node_search.js
rzelayafavila: Yeah, I think that is sensible. I think that the code for it wouldn't necessarily need to go in the ``node.js`` file, but if they do not share much in common and it would not be helpful then it makes sense not to do it.

interface/src/app/node/node_search.js
rzelayafavila: Hmm are they really different though? It looks like the NodeInfo one just has the ``:id`` parameter for requests that are non-GET. It also looks like it does not use the ApiBasePath, which it probably should, right? Wouldn't it make sense to just keep one of them?

interface/src/app/node/node_search.js
rzelayafavila: Yeah, I see and use ternary operators every now and then in JavaScript as well. I think that the chaining of the two operators together makes it a little harder to read, thus making a comment here helpful.

interface/src/app/node/node_search.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * "adage.node_search" module.
+ */
+
+angular.module('adage.node_search', [
```
rzelayafavila: Quick question - should we use underscore when naming this module rather than camelCase? Do camelCase rules in javascript apply to module names as well?

interface/src/app/node/node_search.js
dongbohu: I got some idea how to do it. Let me try.

interface/src/app/node/node_search.js
dongbohu: I don't know. Let me do some research and get back to you.

dongbohu:
@rzelayafavila: I think I addressed all your comments. I did some research and it seems that most angular developers prefer the module name in lowerCamelCase, so I renamed the module name from "node_search" to "nodeSearch".

In the new "resources.js", the main reason why Node resource service is named "NodeInfo" instead of "Node" is because `sampleBin.js` (written by @mhuyck) is referring to it. I am going to create a new issue to remind ourselves that. When his current PR is done, one of us can change the service name into "Node", and probably also move `NodeInfoSet` service into `node/resources.js` too.

PR #140, opened 2017-05-03
Backend changes to add ParticipationType and link it to existing Participation objects in the database. Also adding a couple of new management commands and modifying an existing one.

rzelayafavila:
Hey everyone, this is kind of a heavier PR than I anticipated since it involved making a data migration to create a ParticipationType in the database (if one did not already exist) and adding it as the ForeignKey to all the existing Participation objects in the database.

This is the approach I used based on these docs:
https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations
and this Stack Overflow question & responses:
http://stackoverflow.com/questions/36280253/default-value-for-foreign-key-in-django-migrations-addfield

Let me know if this is a sensible approach. Thanks!

### DIFF
--- a/adage/analyze/management/commands/create_or_update_participation_type.py
+++ b/adage/analyze/management/commands/create_or_update_participation_type.py
@@ -5,8 +5,8 @@ command-line arguments. First argument after the command is the 'name'
 argument and the second is the 'description' argument.
 
 Example usage:
-python manage.py create_or_update_participation_type "High weight genes" \
-    "A high weight gene in a signature is a genes whose weight is ..."
+python manage.py create_or_update_participation_type "High-weight genes" \
+    "High-weight genes are those that most strongly influence the node's ..."
 """
 
 from __future__ import print_function
--- a/adage/analyze/management/commands/create_or_update_participation_type.py
+++ b/adage/analyze/management/commands/create_or_update_participation_type.py
@@ -1,0 +1,58 @@
+"""
+This management command will save new ParticipationType objects to the
+database with the 'name' and 'description' that are passed in the
+command-line arguments. First argument after the command is the 'name'
+argument and the second is the 'description' argument.
+
+Example usage:
+python manage.py create_or_update_participation_type "High weight genes" \
+    "A high weight gene in a signature is a genes whose weight is ..."
+"""
+
+from __future__ import print_function
+from django.core.management.base import BaseCommand, CommandError
+from analyze.models import ParticipationType
+
+import logging
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+
+class Command(BaseCommand):
+    help = ("Creates a participation type in the database for the "
+            "participation of genes in a node.")
+
+    def add_arguments(self, parser):
+        parser.add_argument('name', type=str)
+        parser.add_argument('description', type=str)
+
+    def handle(self, **options):
+        try:
+            (name, description) = (options['name'], options['description'])
+
+            if not name or name.isspace():
+                raise Exception("Name for participation type was not provided "
+                                "or was blank.")
+            elif not description or description.isspace():
+                raise Exception("Description for participation type was not "
+                                "provided or was blank.")
+            else:
+                try:
+                    existing_pt = ParticipationType.objects.get(name=name)
+                    existing_pt.description = description
+                    existing_pt.save()
+                    self.stdout.write(self.style.NOTICE(
+                        "Description for participation type with name '%s' "
+                        "has been updated." % name))
+
+                except ParticipationType.DoesNotExist:
+                    ParticipationType.objects.create(
+                        name=name, description=description)
+
+                    self.stdout.write(self.style.NOTICE(
+                        "Participation type '%s' created "
+                        "successfully." % name))
+
+        except Exception as e:
+            raise CommandError(
+                "Failed to create participation type with data: %s" % e)
--- a/adage/analyze/management/commands/create_or_update_participation_type.py
+++ b/adage/analyze/management/commands/create_or_update_participation_type.py
@@ -1,5 +1,5 @@
 """
-This management command will save new ParticipationType objects to the
+This management command will save a new ParticipationType object to the
 database with the 'name' and 'description' that are passed in the
 command-line arguments. First argument after the command is the 'name'
 argument and the second is the 'description' argument.
@@ -54,5 +54,5 @@ class Command(BaseCommand):
                         "successfully." % name))
 
         except Exception as e:
-            raise CommandError(
-                "Failed to create participation type with data: %s" % e)
+            raise CommandError("Failed to create or update participation "
+                               "type with data: %s" % e)
--- a/adage/analyze/management/commands/delete_participation_type.py
+++ b/adage/analyze/management/commands/delete_participation_type.py
@@ -5,7 +5,7 @@ ParticipationType objects with that 'name' in the database, a message
 will be printed to the terminal stating this.
 
 Example usage:
-python manage.py delete_participation_type "High weight genes"
+python manage.py delete_participation_type "High-weight genes"
 """
 
 from __future__ import print_function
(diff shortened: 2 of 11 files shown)


adage/analyze/models.py
```diff
@@ -210,6 +223,8 @@ class Participation(models.Model):
     """
     node = models.ForeignKey(Node, on_delete=models.PROTECT)
     gene = models.ForeignKey(Gene, on_delete=models.PROTECT)
+    participation_type = models.ForeignKey(ParticipationType,
+                                           on_delete=models.PROTECT)
 
     class Meta:
         unique_together = ('node', 'gene')
```
dongbohu: Please confirm that `unique_together` should include `participation_type` or not.

adage/analyze/management/commands/create_or_update_participation_type.py
```diff
@@ -0,0 +1,58 @@
+"""
+This management command will save new ParticipationType objects to the
```
dongbohu: To be more accurate, this command actually saves **a** new Participation object (instead of **objects**), right?

adage/analyze/management/commands/create_or_update_participation_type.py
```diff
@@ -0,0 +1,58 @@
...
+                    ParticipationType.objects.create(
+                        name=name, description=description)
+
+                    self.stdout.write(self.style.NOTICE(
+                        "Participation type '%s' created "
+                        "successfully." % name))
+
+        except Exception as e:
+            raise CommandError(
+                "Failed to create participation type with data: %s" % e)
```
dongbohu: I think "Failed to create or update ..." will be more accurate.

adage/analyze/management/commands/delete_participation_type.py
```diff
@@ -0,0 +1,46 @@
+"""
+Management command that only serves to delete ParticipationType objects in
```
dongbohu: Since the name is unique in the database, this command "serves to delete **a** ParticipationType object" instead of "... objects".

adage/analyze/management/commands/import_node_gene_network.py
```diff
@@ -73,12 +83,21 @@ def import_network(file_handle, ml_model_name):
     except MLModel.DoesNotExist:
         raise Exception("%s does NOT exist in the database" % ml_model_name)
```
dongbohu: I just realized that this error message could be a little more user-friendly if we change it into:
```
raise Exception("Input machine learning model (%s) does NOT exist in the database" % ml_model_name)
```
What do you think? (I apologize because I was the original author of this script).

adage/analyze/management/commands/import_node_gene_network.py
```diff
@@ -73,12 +83,21 @@ def import_network(file_handle, ml_model_name):
...
 
+    # Raise an exception if participation_type_name does not exist in
+    # the database.
+    try:
+        participation_type = ParticipationType.objects.get(
+            name=participation_type_name)
+    except ParticipationType.DoesNotExist:
+        raise Exception("%s does NOT exist in the database" %
```
dongbohu: What about:
```
raise Exception("Input participation type (%s) does NOT exist in the database" %
```
to give user a little more information?

adage/analyze/management/commands/import_node_gene_network.py
```diff
@@ -123,17 +141,26 @@ def check_and_import(file_handle, ml_model):
                 continue
 
             # Raise an exception if the combination of (node, gene)
-            # already exists in Particapation table.  Instead of relying
+            # already exists in Participation table. Instead of relying
```
dongbohu: Thank you for correcting my typo. I actually left two spaces at the end of a complete sentence on purpose.
See PEP8 at https://www.python.org/dev/peps/pep-0008/#id30
```You should use two spaces after a sentence-ending period.```
I noticed that our existing code didn't follow this rule very strictly. So it is totally okay with me if you guys want to use one space after the period.

File: adage/analyze/management/commands/import_node_gene_network.py [hunk 8]
```diff
@@ -123,17 +141,26 @@ def check_and_import(file_handle, ml_model):
...
             # on the IntegrityError exception implicitly, we raise an
             # explicit exception that includes the input file's line
             # number where the error is detected, and node name and gene
             # name involved.
+            # * Note: We do not need to check if the Participation's
+            # participation_type matches with our participation_type when
+            # we check if the Participation obj already exists,
+            # since a unique node AND a unique gene should yield at most
+            # *one* Participation object.
```
dongbohu: I am not sure whether this note is correct. According to what @cgreene said the other day on slack, I feel that it is possible that exactly same node and same gene may have different participation types and thus have multiple records in **Participation** table. (That is why I feel that the `unique_together` restriction in `models.py` file should include `participation_type` too.)

adage/analyze/management/commands/create_or_update_participation_type.py
rzelayafavila: Yep, you're right, that would be more accurate, I will change it.

adage/analyze/management/commands/create_or_update_participation_type.py
rzelayafavila: Ah yes, good call, I will change this.

adage/analyze/management/commands/delete_participation_type.py
rzelayafavila: Yep, just I agree that's more accurate, like in the other management command.

adage/analyze/management/commands/import_node_gene_network.py
rzelayafavila: Ah gotcha - good catch! I was going for the style of writing in English that just uses one space after a sentence-ending period, but I'm glad you pointed out that PEP8 states otherwise! I will change this.

adage/analyze/management/commands/import_node_gene_network.py
rzelayafavila: Sure, I think this sounds good, I will change it!

adage/analyze/management/commands/import_node_gene_network.py
rzelayafavila: Yes, I agree that that is more descriptive! I will change it.

adage/analyze/management/commands/import_node_gene_network.py
rzelayafavila: Ah ok, it's true, it is possible for one same gene and node to have different participation types, so I will change that here and in the ``unique_together`` restriction in ``models.py``. Good catch!

rzelayafavila:
Ok @dongbohu - I made the changes regarding the 'unique_together' restriction and your other comments. Thanks!

adage/analyze/migrations/0009_auto_20170503_1700.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
rzelayafavila: This file should also be code reviewed, because even though it is a migration file, it is a data migration file that I created manually to handle the non-null participation_type FK constraint when no ParticipationTypes have been added to the database.

rzelayafavila:
By the way @dongbohu as I just mentioned, the 'adage/analyze/migrations/0009_auto_20170503_1700.py' file also needs to be code reviewed since I manually made it to make a data migration. I should have mentioned this earlier - did you happen to go through it when you reviewed this PR?

dongbohu:
No, I haven't reviewed the migration file yet. I will review the whole PR again on Monday morning.

rzelayafavila:
Ok, thanks @dongbohu! I actually made one more commit since I talked to Casey about the high-weight genes description, let me know if that last commit looks good to you too! Thanks!

dongbohu:
@rzelayafavila: I reviewed the latest change and it looks good to me.

rzelayafavila:
Ok, thanks @dongbohu! Merging now.

PR #141, opened 2017-05-05
Heatmap select mlmodel

### LINKED ISSUE #115: Hard coded MLModel in analysis.js
In `app/analyze/analysis/analysis.js`, lines 55-56:
```
    // FIXME MLModel is hard-coded until heatmap has a selector added
    $state.go('volcano', {'mlmodel': 1});
```
Although Matt has labelled it as a **FIXME**, I want to emphasize the importance of this fix, because on my desktop, the default MLModel ID is 4 (for historical reasons), and this hard coded ID prevents me from testing the link to volcano plot.

mhuyck:
This set of changes fixes #115 by adding an MLModel selector to the heatmap page and giving it a couple of enhancements I needed to make it work well in that context.

Discussed with @cgreene modifying the overall flow to require users to select MLModel (on e.g. the home page) before proceeding down the path of analysis. I do think we need to streamline to user flow through the system, but doing so would have required many refactoring changes outside the scope of this PR, so I'm writing that up as a separate issue.


interface/src/app/mlmodel/selector.tpl.html
```diff
@@ -1,8 +1,8 @@
 <div class="alert alert-danger" role="alert" ng-if="$ctrl.errorMessage">
   {{$ctrl.errorMessage}}
 </div>
-<select
-  ng-options="option as option.title for option in $ctrl.availableMlModels"
-  ng-model="$ctrl.selectedMlModel">
-  <option value="">--- Please select a machine learning model ---</option>
+<select name="selectMlModel"
+  ng-options="opt.title for opt in $ctrl.availableMlModels track by opt.id"
```
dongbohu: I think `track by opt.id` is redundant here because `title` field in `models.py` is unique.

interface/src/app/mlmodel/selector.tpl.html
```diff
@@ -1,8 +1,8 @@
 <div class="alert alert-danger" role="alert" ng-if="$ctrl.errorMessage">
   {{$ctrl.errorMessage}}
 </div>
-<select
-  ng-options="option as option.title for option in $ctrl.availableMlModels"
-  ng-model="$ctrl.selectedMlModel">
-  <option value="">--- Please select a machine learning model ---</option>
```
dongbohu: Any reason why you delete this reminder line? I kind of like it.

mhuyck: It forces the select box to be *really* wide. If it is necessary on a particular page, I would prefer to see this sort of instruction appear in a form label or tool tip to be implemented by the parent view rather than in the reusable component itself.

interface/src/app/mlmodel/selector.tpl.html
mhuyck: I found that the select box would not properly respond to an already-selected mlmodel without this detail.

mhuyck:
Sorry for the lack of demo site... that would definitely be good for this PR. Should have up and running by the morning.

mhuyck:
Demo site now running at http://greenelab.fgtech.com/ .

interface/src/app/mlmodel/selector.tpl.html
dongbohu: Okay, that makes sense.

interface/src/app/mlmodel/selector.tpl.html
dongbohu: It is weird. Do you know why?

dongbohu:
Here is a screenshot from the demo site. What is the text below the note? Is it possible that part of the heatmap Y axis has been rendered before the mlmodel is selected? 
screenshot from 2017-05-09 11-43-18

mhuyck:
Ah, yes. Vega is rendering the y-axis even though there are no data to plot. Let me see if I can tidy that up. Might be too tricky for this PR.

interface/src/app/mlmodel/selector.tpl.html
mhuyck: I was following the documentation here: https://docs.angularjs.org/api/ng/directive/select#using-select-with-ngoptions-and-setting-a-default-value

I can try again without the `track by` part. Maybe that wasn't the change that actually made it work? I think it's a good idea to use `.id` for tracking, though. It just makes more sense to me than using the `.title`.

mhuyck: It definitely does not work if I remove `track by opt.id`. I suspect the reason is that the `select` directive may be trying to compare the entire `ng-model` *object* with what it finds in `$ctrl.availableMlModels` and it may be working with a copy of the object somewhere so they don't match. Using `track by opt.id` tells it exactly how to match the selection with the `ng-model` so it works. Just a hunch, though, without digging into the Angular source for the `select` directive.

dongbohu:
Thank you for your reply. If it is not easy to remove that little piece of Y-axis, please feel free to file an issue and deal with it some other time.

interface/src/app/mlmodel/selector.tpl.html
dongbohu: I think you right. See this doc:
https://docs.angularjs.org/api/ng/directive/ngOptions
The second part **Complex models (objects or collections)** talks about the advantage of using `track by` syntax.

mhuyck:
I found a quick fix: the heatmap is now hidden until an MLModel is selected. The Y-axes now shows briefly before the new data render, but it's better than it was.

mhuyck:
Not sure why CircleCI is failing. Looks to be the `pip install` step, which worked fine for me and on Codeship. There are no changes to pip packages here, either. Temporary glitch?

mhuyck:
Updated code and demo site. Let me know if you still approve and I'll merge.

I did a rebase, so that's why it looks like all of the changesets are new. This also means the demo site now shows the link from the volcano plot to the signature page.

dongbohu:
I have no idea why `circleci` is not working.

PR #147, opened 2017-05-08
[FEAT] Add a link from volcano plot to signature page

### LINKED ISSUE #137: Linking signatures in volcano plot and heatmap to signature page
Users should be able to examine a signature in its signature page after identifying an interesting signature in the volcano plot or in the heatmap.

mhuyck:
This fixes #137.


mhuyck:
Thanks!

Issue #138, opened 2017-05-03
Change "Node" to "Signature" (back end mostly)


dongbohu:
For historical reasons, **"signature"** has been called **"node"** in both adage server's backend and frontend. I wonder whether it is a good idea to change all occurrences of **node** (including the database table of backend and the related djangdo management commands and the URLs) into **signature** so that it will be consistent with @tj8901nm's R package and paper.  

One concern I have now is that some users may have saved bookmarks whose URL is `host://#/node/<id>`. After this change, the url will become `hostname/#/signature/<id>` and the original one will be broken. How easy is it to redirect all URLs that include `node` to the new ones?

dongbohu:
@rzelayafavila will take care of the `help` page.

dongbohu:
Excluded `node` occurrences in PR #149:
* All variables and string literals that are related to API, such as **/api/v0/node/...** or **/api/v0/activity/?node=xxx...**;

* app/analyze/analysis/sampleBin.js:
```
sampleClust.orderedNodes()
```
(occurred twice in this file)

dongbohu:
Renaming on front end is done. The renaming on back end will include:
* adage/urls.py
* adage/analyze/api.py
* adage/analyze/models.py
* adage/analyze/tests.py
* adage/analyze/management/commands/create_or_update_participation_type.py
* adage/analyze/management/commands/import_activity.py
* adage/analyze/management/commands/import_node_gene_network.py
* migration files in `adage/analyze/migrations/` (should be generated automatically by Django)

The modification of `api.py` will affect the frond end again. Any front end files in `interface/src/` directory that refers to the API such as **/api/v0/node/...** or **/api/v0/activity/?node=xxx...;** should be modified too.

mhuyck:
An issue closely related to this came up in discussion with the Hogan lab: when viewing the list of Signatures at http://adage.greenelab.com/#/signature/search?mlmodel=1 it is a bit odd that they are all named "Node###neg" or "Node###pos" (still using the term "Node").

These particular names come from the data files we are loading. It should be a simple matter to rename those to something like "Sig###neg" and "Sig###pos" right? The labels appear in at least two data files, so we would need to be consistent and thorough when doing the update to avoid breaking any references.

Let me know if you think this should be broken out as a separate issue and I will do that.

dongbohu:
@mhuyck: The `Nodexxx` name format exists in the following two files in `adage-server/data/` directory:
```
[...]

PR #149, opened 2017-05-10
WIP: Fix of Issue 138 (Change "Node" to "Signature")

dongbohu:
This is a PR in progress which will fix issue #138.

So far it passes all unit tests but the heatmap isn't shown up correctly. I got the following error when loading the heatmap:
```
TypeError: Cannot read property '_id' of undefined
```
You can reproduce the error by going to:
http://165.123.67.202/#/analyze
Search and add some samples, then click the test tube icon on top right corner, select the machine learning model of "Emsemble Adage 300". Then you should be able to see the error message on the console. So I must have made some mistake when modifying the related modules.

All the other modules seem to be working fine.

@mhuyck and @rzelayafavila: Please play with the web UI and tell me anything that is not working. Thanks.


mhuyck:
Looks like an issue inside Vega. I'm looking now.

dongbohu:
Don't worry about the issues reported by codeclimate for now. I will fix them when this PR is done.

mhuyck:
Here is the problem: the Activity API returns JSON data that still refer to them as `node`s. The data from this API are loaded without modification into the dataset that Vega uses to draw.

> http://165.123.67.202/api/v0/activity/?mlmodel=4&sample=118&format=json

mhuyck:
Looking for a simple way to fix. I think we can revert a few of the changes in the   `$scope.heatmapSpec` and it will work again.

dongbohu:
Ah, I didn't know that. So I must have modified some places that shouldn't be.

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -91,36 +91,36 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
...
-        // compute minimum normalized value for each node (across samples)
-        'name': 'node_summary',
+        // compute minimum normalized value for each signature (across samples)
+        'name': 'signature_summary',
         'source': 'activity',
         'transform': [
           {
             'type': 'aggregate',
-            'groupby': ['node'],
+            'groupby': ['signature'],
```
mhuyck: change this back to `node`

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -91,36 +91,36 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
...
         'transform': [
           {
             'type': 'lookup',
-            'on': 'node_summary',
-            'onKey': 'node',
-            'keys': ['node'],
-            'as': ['node_summary']
+            'on': 'signature_summary',
+            'onKey': 'signature',
+            'keys': ['signature'],
```
mhuyck: revert lines 117 and 118 to say `node`

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -129,22 +129,22 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
             'as': ['sample_order']
           }, {
             'type': 'lookup',
-            'on': 'nodeOrder',
+            'on': 'signatureOrder',
             'onKey': 'data',
-            'keys': ['node'],
-            'as': ['node_order']
+            'keys': ['signature'],
```
mhuyck: revert this line to `node`

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -202,7 +202,7 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
         'properties': {
           'enter': {
             // TODO dynamic layout: compute these constants also (see above)
-            'text': {'value': 'node'},
+            'text': {'value': 'signature'},
```
mhuyck: this needs to revert to `node` also

dongbohu: I made these changes and I can see the heatmap now! But when I click "Cluster Signatures" button, I got this error:
```
TypeError: Cannot read property 'value' of undefined
```
I suspect this is due to some similar replacement. Can you take a look?

mhuyck: Okay, I'm on it.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -144,39 +144,39 @@ MathFuncts, errGen) {
...
-      //      by walking through the `firstSampleNodes` and constructing a
-      //      `nodeObject` for each. [outer .map()]
-      var retval = firstSampleNodes.map(function(val) {
-        var nodeObject = {
+      // (2a) next, we build a new array (`retval`) comprised of
+      //      `signatureObject`s by walking through the `firstSampleSignatures`
+      //       and constructing a signatureObject` for each. [outer .map()]
+      var retval = firstSampleSignatures.map(function(val) {
+        var signatureObject = {
           'id': val.node,
```
dongbohu: I think we should keep `node` on this line, right?

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -144,39 +144,39 @@ MathFuncts, errGen) {
...
-            // (2b) the array of activity for each node is built by plucking the
-            //      activity `.value` for each sample within this node from the
-            //      `activityCache` [inner .map()]
+            // (2b) the array of activity for each signature is built by
+            //      plucking the activity `.value` for each sample within this
+            //      signature from the `activityCache` [inner .map()]
             function(sampleId) {
-              // FIXME: counting on array order to match node order here
+              // FIXME: counting on array order to match signature order here
               return this.activityCache.get(sampleId)[val.node - 1].value;
```
dongbohu: I also keep `node` on this line. Not sure whether it is correct.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: Yes. That is right.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: Also need to keep it as `node` here, for the same reason as before.

interface/src/app/analyze/analysis/analysis.js
mhuyck: Hmm. This one might be a bug that's just now showing up because the way your database was loaded is different from mine. If my hunch is correct, you would never have been able to run this clustering the data from your database before.

I'm digging deeper, but might not resolve it before I have to do school pick-up. Will have to clone your branch and try to replicate.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: It's this line that's biting us. See that `FIXME` above? I figured this would cause a problem some day. Today is the day it gets fixed. ✊

mhuyck: I think I see what to do. Might be simpler to fix than I feared, but will still take some testing.

dongbohu: Thanks!

mhuyck: This problem is actually separate from the renaming you did. If you want to proceed with this PR, let's file a separate issue. I will probably need to change several lines that are scattered around so it would be easier to fix that way anyway.

mhuyck: I will need help from your database to test my fix when it's ready, though, so don't do anything crazy like rebuild your dev database or anything, okay?

mhuyck:
Okay, I've logged a separate issue for the bug so it doesn't need to hold this up any more. If you haven't found any other issues from this set of changes, I'd say you've completed the UI portion of renaming from "node" to "signature" so this is ready to merge.

mhuyck:
There might be one codeclimate line length issue that you want to fix but I'll leave that up to you. I think we're softening our 80-char line length rule anyway.

dongbohu:
Thanks. I fixed the two lines that are longer than 80 char. Will take another look and if everything is okay, I will merge it.

PR #151, opened 2017-05-15
Changes to heatmap and volcano plot to fix #150

### LINKED ISSUE #150: Heatmap node clustering fails under some conditions
While working through PR #149 with @dongbohu, it became clear that the implementation of node clustering has a flaw that is revealed by the database in @dongbohu's development server.

Node clustering fails to run and the JavaScript Console throws the following error:

```
[Error] Error: undefined is not an object (evaluating 'this.activityCache.get(b)[a.node-1].value')
http://165.123.67.202/static/assets/adage-0.7.0b.js:41:1328
map@[native code]
http://165.123.67.202/static/assets/adage-0.7.0b.js:41:1273
map@[native code]
getSignatureObjects@http://165.123.67.202/static/assets/adage-0.7.0b.js:41:1206
http://165.123.67.202/static/assets/adage-0.7.0b.js:41:2835
g@http://165.123.67.202/static/assets/adage-0.7.0b.js:28:9827
http://165.123.67.202/static/assets/adage-0.7.0b.js:28:9999
$digest@http://165.123.67.202/static/assets/adage-0.7.0b.js:28:16130
http://165.123.67.202/static/assets/adage-0.7.0b.js:28:17733
f@http://165.123.67.202/static/assets/adage-0.7.0b.js:26:27932
http://165.123.67.202/static/assets/adage-0.7.0b.js:26:29390
	(anonymous function) (adage-0.7.0b.js:27:20333)
	g (adage-0.7.0b.js:28:9911)
	(anonymous function) (adage-0.7.0b.js:28:10000)
	$digest (adage-0.7.0b.js:28:16131)
	(anonymous function) (adage-0.7.0b.js:28:17734)
	f (adage-0.7.0b.js:26:27933)
	(anonymous function) (adage-0.7.0b.js:26:29391)
```

mhuyck:
This should fix the heatmap issue that @dongbohu and I uncovered last week. Still need to tidy up but filing as a WIP so we can test with Dongbo's database.


mhuyck:
Hah. There's a conflict because I haven't yet rebased to incorporate @dongbohu's "Node" to "Signature" renaming yet. Will fix that before merging, of course.

dongbohu:
Since `Activity` in model.py has been changed, do I need to rebuild the activity table too?

mhuyck:
I thought I would need to do that, too, but when I ran `makemigrations` it didn't change anything. I suspect that the existing `unique_together` constraint had already made those indexes implicitly.

dongbohu:
The local web server on my desktop is using the latest code, but it doesn't seem to be working:
http://165.123.67.202/#/home
Feel free to play with it.

mhuyck:
I will have a look. Although you won't need to change the database, you will need to update the backend using `sudo supervisorctl restart adage`. Did you do that?

dongbohu:
Yes, I did.

mhuyck:
I don't see the MLModel selector on the page above the heatmap. Are you sure you have the right code?

mhuyck:
Note that you will need to make a fresh clone of the `heatmap-node-order-fix` from *my* repo since we have not yet merged with `greenelab/master`. You cannot just do a `pull` to get your server updated.

dongbohu:
I found the reason: I messed up the new version with another version that was cloned from your repo. Let me fix it.

mhuyck:
Tested this fix with @dongbohu's database and it works. Now tidied up and ready for full review so removing the WIP label.

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -161,17 +161,25 @@ MathFuncts, errGen) {
...
             function(sampleId) {
-              // FIXME: counting on array order to match signature order here
-              return this.activityCache.get(sampleId)[val.node - 1].value;
+              var cachedActivity = this.activityCache.get(sampleId);
+              if (cachedActivity[index].node !== val.node) {
+                // ensure we're pulling out the right node (aka signature)
+                $log.error(
+                  'getSignatureObjects: signature IDs do not match. First ' +
+                  ' sample = ', val, ', but sample ' + sampleId + ' =',
+                  cachedActivity[index]
```
dongbohu: In this error message, both `val` and `cachedActivity[index]` are objects, do you think it makes more sense to only print out `val.node` and `cachedActivity[index].node`?

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -432,32 +441,42 @@ MathFuncts, errGen) {
...
+              //      plucking the activity `.value` for each sample within the
+              //      `index`th signature from the `activityCache`
+              //      [inner .map()]
+              var cachedActivity = cbSampleBin.activityCache.get(sampleId);
+              if (cachedActivity[index].node !== signatureId) {
+                // ensure we're pulling out the right node
+                $log.error(
+                  'mapSignaturesToSignatureInfo: signature IDs do not match.' +
+                  ' First sample = ' + signatureId + ', but sample ' +
+                  sampleId + ' =', cachedActivity[index]
```
dongbohu: Same question on whether to include the unmatched signature ID only in the log.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: I did that on purpose. It's true that we're just checking the IDs to be sure they match, but if they *don't* match then I think I'll want to see the extra information to help diagnose what is going wrong. I expect this to be a rare error condition, so if it does arise I want it to be easy to track down the source of trouble.

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: This case is slightly less verbose (note that the "first sample," in this case, is including just the `signatureId`. I would have done this the same way as the other error for the same reasons, but we are mapping an array that has already been translated from objects to IDs, so I no longer have the full object.

Anyway, let me know if you still think this is too much to include in the log, given that I expect this will be a rare error condition.

interface/src/app/analyze/analysis/sampleBin.js
dongbohu: Sure. It makes sense.

Issue #152, opened 2017-05-15
Volcano plot fails to load with long Node IDs


mhuyck:
Testing some changes on @dongbohu's database revealed a latent bug that will appear when Node IDs grow beyond 3-4 digits. The volcano plot in our test instance failed to load with the following error:

```
[...]

Issue #148, opened 2017-05-09
Got 'Failed to get genes: : 400 Bad Request. Please try again later.' while trying to "View in Gene Network"


mhuyck:
Steps to reproduce:

1. Search for 'anr' under Analyze tab
2. Experiment `E-GEOD-17296` will be at the top. Add its samples to the sample bin (click `+` at right)
3. Select all 4 samples with genotype "WT" for the base group (purple)
4. Select all 4 samples with genotype "∆anr" for the comparison group (yellow)

![screen shot 2017-05-09 at 9 28 04 am](https://cloud.githubusercontent.com/assets/1151166/25853175/dc52de18-3499-11e7-8cb0-96b3b11e89dc.png)

5. Click "Volcano Plot" button and select "Node41pos" (the node shown below near the upper right of the plot)

![screen shot 2017-05-09 at 9 21 51 am](https://cloud.githubusercontent.com/assets/1151166/25853210/00cf2350-349a-11e7-8548-b6ee1b16a3ba.png)

6. Click the "View in Gene Network" button and the "Failed to get genes" message appears.

cgreene:
URL longer than allowed? Just a guess...

mhuyck:
This is very similar to bug #146, but the error appears in a slightly different place so the message is different.

mhuyck:
Based upon what I see in the console log, @cgreene's hunch appears to be exactly what it is. I suspect the same issue with #146.

mhuyck:
The URL it's trying to get is this, which is 4432 characters long:

```
[...]

Issue #146, opened 2017-05-07
Got 'failed to get gene expression value: : 502 Bad gateway.' when analyzing E-MEXP-2593


mhuyck:
Report from Deb Hogan:

> I am analyzing E-MEXP-2593 and picked two nodes to look at in the gene network and got this.

> I think that this was Dao's request as well, but it would be nice if you could type in a node in the  Experiments with the Highest Range page.

> One node at a time worked.

mhuyck:
I see the string 'failed to get gene expression value' exactly once in the codebase: `app/gene/network/network.js`, currently line 446.

dongbohu:
`E-MEXP-2593` is the accession of an experiment. What does it mean by `picked two nodes to look at in the gene network`?

mhuyck:
I agree that's confusing. I think she is trying to select nodes in a volcano plot, but can't get the error that way. I will meet with Deb and document the steps to reproduce it.

mhuyck:
Got steps to reproduce via Deb just now.

1. search for `E-MEXP-2593` in the Analyze tab
2. add those samples to the SampleBin (click `+` at right of search result)
3. select the following groups and click "volcano plot":

<img width="1372" alt="screen shot 2017-05-15 at 1 18 24 pm" src="https://cloud.githubusercontent.com/assets/1151166/26070347/4b2ddcca-3972-11e7-9452-02d873c76273.png">

4. select the two signatures above the line (`Node226neg` and `Node297neg`)
5. click "View in Gene Network" and see the error

mhuyck:
Note: the actual error is "Bad Gateway" and appears to arise from the `nginx` layer. When testing directly against the `django-admin runserver` backend, the server happily processes the request and returns the result to the browser.

While we could, in theory, make adjustments to `nginx` settings, this would likely only defer the problem rather than resolve it. The defaults for `nginx` are set up the way they are for good reasons, and other browsers (I am testing with Safari at the moment) will not handle extremely long URIs gracefully.

I mention this here because reproducing the error requires some subtlety.

mhuyck:
Just now got the fix for this working... bug fix should roll out to the server in a few days after code is reviewed and updated.

mhuyck:
Fixed in PR #155

PR #155, opened 2017-05-23
Network uri bug

mhuyck:
Fix for #146. These changes allow us to use a POST to work around the length limitations of GET.

This was the cleanest way I could find to make this happen. The technique here should be easy to generalize to other cases where it will need to be applied, such as #148 and #152.


kurtwheeler:
LGTM :+1:

adage/analyze/api.py
```diff
@@ -5,7 +5,7 @@
 from organisms.api import OrganismResource
 from genes.api import GeneResource
 from tastypie import fields, http
-from tastypie.resources import Resource, ModelResource
+from tastypie.resources import Resource, ModelResource, convert_post_to_VERB
```
dongbohu: This is an interesting approach. I am surprised that tastypie doc never mentioned it. Can you add at least one test case in `tests.py` to confirm that it works? (I forgot to do that when adding `ExpressionValueResource` class. My bad.)

interface/src/app/gene/network/network.js
```diff
@@ -304,7 +315,7 @@ angular.module('adage.gene.network', [
             },
             function error(response) {
               var message = errGen(
-                'Failed to get signature info for gene edge: ', response);
```
dongbohu: Thank you for correcting this typo.

interface/src/app/gene/network/network.js
```diff
@@ -441,22 +459,22 @@ angular.module('adage.gene.network', [
                   drawNetwork();
                 },
                 function error(response) {
-                  var message = errGen('Failed to get gene expression value: ',
```
dongbohu: :+1:

interface/src/app/gene/network/network.js
```diff
@@ -441,22 +459,22 @@ angular.module('adage.gene.network', [
...
+                  var message = errGen('Failed to get gene expression value',
                                        response);
                   $log.error(message);
                   self.statusMessage = message + '. Please try again later.';
                 }
               );
             },
             function error(response) {
-              var message = errGen('Failed to get genes: ', response);
```
dongbohu: :+1:

interface/src/app/gene/network/network.js
```diff
@@ -441,22 +459,22 @@ angular.module('adage.gene.network', [
...
+              var message = errGen('Failed to get genes', response);
               $log.error(message);
               self.statusMessage = message + '. Please try again later.';
             }
           );
         },
         function error(response) {
-          var message = errGen('Failed to get edges: ', response);
```
dongbohu: :100:

interface/src/app/gene/network/network.js
```diff
@@ -390,9 +401,16 @@ angular.module('adage.gene.network', [
...
 
-              ExpressionValue.get(
-                {'gene__in': geneID.join(), 'sample__in': sampleID.join(),
-                  'order_by': 'gene'},
+              ExpressionValue.post(
+                {},
+                // Angular does not serialize POST data the way we would expect
+                // so we have to do it manually here. For more details, see
+                // https://github.com/angular/angular.js/issues/6039#issuecomment-113502695
+                $httpParamSerializerJQLike({
```
dongbohu: I found this link helpful in understanding `$httpParamSerializerJQLike` service in AngularJS:
https://docs.angularjs.org/api/ng/service/$httpParamSerializerJQLike

adage/analyze/api.py
mhuyck: Let me see if I can make an `api_test` test case do anything useful. I had thought about a test case for this but then thought we won't quite have the right testing infrastructure in place until we are doing E2E calls with a live server. That would obviously be the gold standard, but maybe there's something useful I can get out of `api_test`.

- [ ] see if an `api_test` test case would be helpful here

interface/src/app/gene/network/network.js

The discussion at Angular issue 6039 was more enlightening for me than the documentation. It seems there are some on the Angular team who agree that the framework could do more to help here and others who seem to prefer that "correcting" the default behavior would break too many things. It seems that we are left to do this sort of manual patch for the foreseeable future.

adage/analyze/api.py
dongbohu: True. If you can create a test case whose POST request is longer than 4KB and confirm that data is retrieved from database successfully, it will be good enough. Thanks.

mhuyck: Sure. I got a few basic tests written up yesterday. I will be sure to include the case you describe, too.

mhuyck: I've added several tests now, aimed at exercising the ExpressionValue API the way we have tested our other APIs.

PLEASE NOTE: although there is a test included to ensure that we can retrieve data using the POST API method that would normally fail a GET, this test is somewhat non-informative because the failure actually takes place in the nginx reverse-proxy layer, which cannot be tested without a complete E2E test infrastructure.

mhuyck:
Codeclimate is complaining about cyclomatic complexity in the function that generates test data for the ExpressionValue API. I would argue that this sort of complexity is *exactly* what we want in our tests. Bring on a million monkeys typing random garbage!

Heh. Anyway, if any of my human reviewers think the test code needs a tweak here or there, I'm all ears.

adage/analyze/tests.py
```diff
@@ -907,3 +928,100 @@ def test_experiment_filter_in_sample(self):
...
+        if MLModel.objects.exists():
+            ml_model = MLModel.objects.first()
+        else:
+            ml_model = factory.create(MLModel)
+
+        # Create genes:
+        for i in range(num_genes):
+            Gene.objects.create(entrezid=(i + 1),
+                                systematic_name="sys_name #" + str(i + 1),
+                                standard_name="sys_name #" + str(i + 1),
```
dongbohu: Typo: should be "std_name". This part of code seems to be copied from `create_edges()` function. Line #169 of this updated file has the same typo (my bad), can you correct it too? Thanks.

adage/analyze/tests.py
```diff
@@ -907,3 +928,100 @@ def test_experiment_filter_in_sample(self):
...
+                                organism=organism)
+
+        # Note: no need to create samples. See setUp() above.
+
+        # Now, for each sample, randomly select Genes and create
+        # ExpressionValue objects for them
+        self.expressionvalue_total = 0
```
dongbohu: The variable name `expressionvalue_total` is a little misleading. (I originally thought it was the sum of expression values instead of the number of expression value records in the database). Maybe we can change it into  "expression_value_counter" or something like that?

adage/analyze/tests.py
```diff
@@ -907,3 +928,100 @@ def test_experiment_filter_in_sample(self):
...
+            len(ev_resp['objects'])
+        )
+
+    def test_expressionvalue_big_post(self):
+        """
+        Test that we can use POST to retrieve more ExpressionValue records
+        that would fit in the ~4k-char limit for a GET
+        """
+        # we select 1100 genes because representing those gene IDs will require
+        # >= (9*1 + 90*2 + 900*3 + 101*4)chars + (1100-1)delimiters = 4392 char
```
dongbohu: Great. Thank you!

adage/analyze/tests.py
mhuyck: Good call...

adage/analyze/tests.py
mhuyck: Nice catch. I'll fix that.

mhuyck:
I think this is ready to go. Can you confirm That I got everything you noted @dongbohu?

mhuyck:
Thanks for the heads up on `requirements.txt`. I will rebase with caution but it will probably be a match anyway.

Issue #130, opened 2017-04-28
experiments with the highest range should be hoverable


cgreene:

cgreene:
Not a dropdown/activity list - hover should give sample annotations.

dongbohu:
After discussion with @casey, we decide to get rid of the "+" sign, and make the vega-lite widget (which shows the activity values) clickable. Clicking it will lead to the page of annotation page of all samples that are related to this experiment.

dongbohu:
Fixed in #156.

Issue #83, opened 2017-02-23
highRangeExp in node.js


dongbohu:
Quote from @rzelayafavila in Pull Request #77:
```
I think that this large highRangeExp directive can actually also be divided into 
smaller directives/components, where one is the directive/component for the 
Node page quadrant that holds all of the Highest Range Experiments, and it
calls smaller directives/components for each experiment itself. That way, 
the experiment in each component only knows information about itself and
the samples it contains. 
```

dongbohu:
Fixed in #156.

PR #156, opened 2017-05-24
Fix of Issues #130 and #83

dongbohu:
This PR is filed to fix issue #130 (and #83 as a byproduct) in **Signature** (aka. Node) page. I touched more code than expected:

1. Removed the samples in each experiment.

2. Made the vega-lite plot clickable. When it is clicked, the page will go to annotations of ALL samples in current experiment.

3. Refactored all resource classes into `resources.js`.

4. Used `$resource` to replace `$http` service.

5. Moved the code of vega-lite specification out of `highRangeExp` directive, and created a new service *EmbedSpecService* for it. It makes the code more readable.

6. Added "data_source" field in sample annotations web UI.

7. Removed redundant resource class in `sampleBin.js`, which refers to `signature/resources.js`.

8. Tweaked the API paths in signature module's test cases.

### DIFF
--- a/interface/src/app/signature/resources.js
+++ b/interface/src/app/signature/resources.js
@@ -13,6 +13,19 @@ angular.module('adage.signature.resources', [
 
 .factory('Participation', ['$resource', 'ApiBasePath',
   function($resource, ApiBasePath) {
-    return $resource(ApiBasePath + 'participation/');
+    return $resource(ApiBasePath + 'participation');
   }
-]);
+])
+
+.factory('Activity', ['$resource', 'ApiBasePath',
+  function($resource, ApiBasePath) {
+    return $resource(ApiBasePath + 'activity');
+  }
+])
+
+.factory('Experiment', ['$resource', 'ApiBasePath',
+  function($resource, ApiBasePath) {
+    return $resource(ApiBasePath + 'experiment');
+  }
+])
+;
--- a/interface/src/app/signature/signature.js
+++ b/interface/src/app/signature/signature.js
@@ -92,8 +92,71 @@ angular.module('adage.signature', [
   }]
 )
 
-.directive('highRangeExp', ['$http', '$log', '$state', 'ActivityDigits',
-  function($http, $log, $state, ActivityDigits) {
+// A service that defines the specifications of embedded vega-lite widget.
+// It is used by "highRangeExp" directive.
+.service('EmbedSpecService', [function() {
+  // sample2index: a dictionary that maps a sample's ID to the index
+  // of this sample's activity value in "EmbedSpecService.spec.data.values".
+  // It will make the toggling of "highlight" flag faster.
+  sample2index = {};
+
+  this.mode = 'vega-lite';
+  this.actions = false;
+  // Method that sets vega-lite specification:
+  this.setSpec = function(activities) {
+    var activityArr = [];
+    var samples = Object.keys(activities);
+    samples.forEach(function(sampleID, index) {
+      activityArr.push({
+        'val': activities[sampleID],
+        'highlight': false
+      });
+      sample2index[sampleID] = index;
+    });
+
+    var vlSpec = {
+      'data': {values: activityArr},
+      'mark': 'tick',
+      'encoding': {
+        'x': {
+          'field': 'val',
+          'type': 'quantitative',
+          'axis': {
+            'grid': false,
+            'title': '',
+            'format': 'r'
+          }
+        },
+        'color': {
+          'field': 'highlight',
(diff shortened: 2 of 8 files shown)


dongbohu:
Demo site:
http://165.123.67.202/#/signature/11344

interface/src/app/signature/signature.spec.js
```diff
@@ -153,27 +153,27 @@ describe('<high-range-exp> directive', function() {
...
-      {id: 13, signature: 123, sample: 1013, value: 0.0790896289324697},
-      {id: 14, signature: 123, sample: 1014, value: 0.0692046195604449},
-      {id: 15, signature: 123, sample: 1015, value: 0.078560557941843},
-      {id: 16, signature: 123, sample: 1016, value: 0.0743736434104904},
-      {id: 17, signature: 123, sample: 1017, value: 0.07871175918013},
-      {id: 18, signature: 123, sample: 1018, value: 0.0614160606502279},
-      {id: 19, signature: 123, sample: 1019, value: 0.0661371872868119},
-      {id: 20, signature: 123, sample: 1020, value: 0.0783672660119669},
-      {id: 21, signature: 123, sample: 1021, value: 0.0769023791604038}
+      {id: 1, node: 123, sample: 1001, value: 0.0854334209211516},
```
dongbohu: I changed the field name in this test case back to `node`, because adage's back end is still using `node` instead of `signature` as field name. We will change these fields when `node` field at the back end is renamed to `signature`.

rzelayafavila: :+1:

interface/src/app/signature/resources.js
```diff
@@ -13,6 +13,19 @@ angular.module('adage.signature.resources', [
 
 .factory('Participation', ['$resource', 'ApiBasePath',
   function($resource, ApiBasePath) {
-    return $resource(ApiBasePath + 'participation/');
+    return $resource(ApiBasePath + 'participation');
```
dongbohu: By default the trailing slashes will be removed. So I removed it in the test cases to make it consistent.

interface/src/app/signature/signature.js
```diff
@@ -92,8 +92,71 @@ angular.module('adage.signature', [
...
+// A service that defines the specifications of embedded vega-lite widget.
+// It is used by "highRangeExp" directive.
+.service('EmbedSpecService', [function() {
+  // sample2index: a dictionary that maps a sample's ID to the index
+  // of this sample's activity value in "EmbedSpecService.spec.data.values".
+  // It will make the toggling of "highlight" flag faster.
+  sample2index = {};
+
+  this.mode = 'vega-lite';
+  this.actions = false;
```
dongbohu: `mode` and `actions` are properties only required by the vega-lite specification.

dongbohu:
@rzelayafavila: About your question on the location of the link to sample annotations page, it is a good question. I feel that since the sample annotations page is related to the samples (the red lines in the vega-lite activity distribution plot), it makes more sense to add the link to the vega-lite plot (instead of to the experiment name). Does this make sense?

dongbohu:
On the other hand, if the URL points to a page of experiment information, then the experiment name will be definitely a better place to put the link onto.

interface/src/app/signature/signature.js
rzelayafavila: Oh ok, I see :+1:

interface/src/app/signature/resources.js
rzelayafavila: Ok

rzelayafavila:
@dongbohu - In response to your first answer, are the samples always the same for any given experiment on the next page? If so, maybe we can add it to the parent element that contains both the vega-lite plot and the experiment name (this is only if both are always linked to the same page)?

dongbohu:
@rzelayafavila: I added the tooltip for the vega-lite widget. Please take a look at the code and demo site and let me know what you think. Thanks. 

Just so you know: I am also working on the workflow redesign, so some of the other demo site pages (such as signature search) is not working now.

rzelayafavila:
Hi @dongbohu - I checked it out and it looks good! Do you know why the tests are failing in CircleCI though?

rzelayafavila:
Hmmm, it looks like it is complaining about the urllib3 version.

dongbohu:
Yeah, I remembered that too.

dongbohu:
@rzelayafavila: I updated `requirements.txt`, also added some comments in `high_range_exp.tpl.html` to explain why the tooltip is placed on the left side instead of right side. Please let me know if you have any other concerns.

Issue #119, opened 2017-04-28
Create and implement a front-end selector for ParticipationType (sub-issue of Issue #58)


rzelayafavila:
Sub-issue of issue #58 

This front-end selector would be tied to whatever ParticipationTypes are added to the backend (issue #118 ).

As per the comments on #58, we should:
*Change the High-weight genes section of the Node page so that it will show the participation type. We may need new widget on web UI to allow users to select participation type that they are interested in.

dongbohu:
Being addressed in PR #160.

rzelayafavila:
Resolved with merging of PR #160

PR #160, opened 2017-06-05
Front end changes for the recently added ParticipationType feature

rzelayafavila:
With these changes, users should be able to select the ParticipationType they want for genes participating in the signatures. If they select a different ParticipationType using the new selector widget, this change should propagate through the participatory-genes and the enriched-genesets directives in the main signature page (/#/signature/<signature_id>). This PR addresses issue #119 .

I am running a test server on my machine at this address, so you can check it out and play with the interface:
http://165.123.67.154:8000/#/signature/400


interface/src/app/signature/participation-type-selector.tpl.html
```diff
@@ -0,0 +1,7 @@
+<div class="alert alert-danger" role="alert" ng-if="$ctrl.errorMessage">
+  {{$ctrl.errorMessage}}
+</div>
+<select name="selectedParticipationType"
```
cgreene: It might be good to put the description of the currently selected participation type immediately after the dropdown. Otherwise, the user is unlikely to understand what the different participation types mean.

rzelayafavila: Ok, I think I will put the description in the signature page template itself rather than here so that how the description is displayed is not hard-coded into this component.

rzelayafavila: Ok, done! If you have time to check it out, this is what it looks like now: http://165.123.67.154:8000/#/signature/400

cgreene: LGTM :+1: - also, I know this isn't in your code but can you change the text to `High-weight genes are those that most strongly influence the signature's activity, and we have found that they often reveal the underlying process or processes captured by the signature.`

The (es) bit looks a little weird, and this way our default load (the HW gene explanation is in some migration, maybe, or is a default?) into a new instance of a server will be something that would be easier to read.

rzelayafavila: Ok, changed it in my local DB and opened a PR for the change to the data migration file here: 
https://github.com/greenelab/adage-server/pull/162

interface/src/app/signature/resources.js
```diff
@@ -11,6 +11,12 @@ angular.module('adage.signature.resources', [
   }
 ])
 
+.factory('ParticipationType', ['$resource', 'ApiBasePath',
+  function($resource, ApiBasePath) {
+    return $resource(ApiBasePath + 'participationtype/');
```
dongbohu: Can you remove the trailing `/` in the API so that it will be consistent with others?

interface/src/app/signature/signature.js
```diff
@@ -23,8 +23,39 @@ angular.module('adage.signature', [
   });
 }])
 
-.controller('SignatureCtrl', ['Signature', '$stateParams', '$log',
-  function SignatureController(Signature, $stateParams, $log) {
+.component('participationTypeSelector', {
```
dongbohu: Do you think this component may be used in any other modules? If yes, it will be better to put in a separate module instead of in `adage.signature`.

interface/src/app/signature/signature.js
```diff
@@ -23,8 +23,39 @@ angular.module('adage.signature', [
...
+    // limit to how many objects it returns).
+    var apiReturnLimit = 0;
+
+    ParticipationType.get({limit: apiReturnLimit},
+      function success(response) {
+        self.availableParticipationTypes = response.objects;
+        self.selectedParticipationType = response.objects[0];
+      },
+      function error(response) {
+        var errMessage = errGen('Failed to get ParticipationTypes: ',
```
dongbohu: No need to put `: ` at the end of the first argument of `errGen()`, whose implementation (in `app/utils.js`) will insert ": ".

interface/src/app/signature/signature.js
```diff
@@ -23,8 +23,39 @@ angular.module('adage.signature', [
...
+                                response);
+        $log.error(errMessage);
+        self.errorMessage = errMessage + '. Please try again later.';
+      }
+    );
+  }]
+})
+
+.controller('SignatureCtrl', ['Signature', '$stateParams', '$log', 'errGen',
```
dongbohu: Pre-warning: My current PR also modified this function. So when you merge your PR, you will have to resolve the conflicts.

interface/src/app/signature/signature.js
```diff
@@ -37,56 +68,66 @@ angular.module('adage.signature', [
...
+                  }
+                  $scope.genes.push(
+                    {sysName: sysName, stdName: stdName, desc: desc,
+                      entrezID: entrezID});
+                }
+                $scope.hypoPercentage = Math.round(numHypo / n * 100);
+                $scope.queryStatus = '';
+              },
+              function error(response) {
+                var errMessage = errGen('Failed to get participatory genes: ',
```
dongbohu: Again, the trailing ": " in the context argument is redundant.

interface/src/app/signature/signature.js
```diff
@@ -37,56 +68,66 @@ angular.module('adage.signature', [
...
+        self.statusMessage = errMessage + '. Please try again later.';
       }
     );
-    self.organism = 'Pseudomonas aeruginosa';
     self.genes = [];
   }
 ])
 
-.directive('highWeightGenes', ['Participation', '$log',
+.directive('participatoryGenes', ['Participation', '$log',
```
dongbohu: I noticed that you used `errGen` service later. Can you add it in dependency list?

interface/src/app/signature/signature.js
```diff
@@ -37,56 +68,66 @@ angular.module('adage.signature', [
...
   function(Participation, $log) {
```
dongbohu: Need `errGen` as a parameter of this function too.

interface/src/app/signature/resources.js
rzelayafavila: Yes, for sure :+1:

interface/src/app/signature/signature.js
rzelayafavila: Ok, thanks for the heads up! I will do a merge of your changes after those have been merged to the master branch and resolve any conflicts that come out.

dongbohu:
By the way, I took a quick look at the two new issues reported by codeclimate. They seem to easy to fix. So you might as well fix them.

interface/src/app/signature/signature.spec.js
```diff
@@ -21,11 +21,15 @@ describe('<high-weight-genes> directive', function() {
...
-    parentScope.signatureID = 123;
-    var testHTML = '<high-weight-genes signature-id="{{signatureID}}" ' +
-        'genes="ctrl.genes"></high-weight-genes>';
-    var uri = '/api/v0/participation?limit=0&node=' + parentScope.signatureID;
+    parentScope.signatureId = 123;
+    parentScope.selectedParticipationType = null;
+
+    var testHTML = '<participatory-genes signature-id="{{signatureId}}" ' +
+        'genes="genes" selected-participation-type=' +
+        '"selectedParticipationType"></participatory-genes>';
```
mhuyck: I might wrap this to be a little more readable as follows:

```javascript
    var testHTML = '<participatory-genes ' +
      'signature-id="{{signatureId}}" ' +
      'genes="genes" ' +
      'selected-participation-type="selectedParticipationType">' + 
    '</participatory-genes>';
```

At any rate, that's how I'd wrap it in an HTML template. But this is a minor quibble.

interface/src/app/signature/signature.spec.js
```diff
@@ -46,11 +50,28 @@ describe('<high-weight-genes> directive', function() {
...
     $httpBackend.expectGET(uri).respond({objects: mockGenes});
+
     parentScope.$digest(); // Start Angular's digest cycle manually.
 
-    // Get isolate scope of <high-weight-genes> directive:
-    var elementScope = element.isolateScope();
+    // Check to make sure that the properties from the parentScope have
+    // been inherited to the child scope (elementScope in this case).
+    expect(elementScope.signatureId).toBe('123');
+    expect(elementScope.selectedParticipationType.id).toBe(1);
```
mhuyck: Nice tests. 👍

interface/src/app/signature/signature.tpl.html
```diff
@@ -5,12 +5,25 @@
 
 <h3 class="text-warning" ng-bind="ctrl.statusMessage"></h3>
 
+<h4>
```
mhuyck: I really think this should just be a paragraph `<p>` block. The contents here are not a (short, succinct) section header, which is what an `<h4>` should be.

interface/src/app/signature/signature.tpl.html
```diff
@@ -5,12 +5,25 @@
...
+  <participation-type-selector
+      selected-participation-type="ctrl.selectedParticipationType">
+  </participation-type-selector>
+  <br>
+  <br>
+  {{ctrl.selectedParticipationType.description}}
```
mhuyck: Hint: if it's a "description" you're showing, it's body text, not a header.

interface/src/app/signature/signature.tpl.html
```diff
@@ -5,12 +5,25 @@
...
+</h4>
```
mhuyck: `</p>`

interface/src/app/signature/signature.tpl.html
```diff
@@ -5,12 +5,25 @@
...
 <div ng-if="!ctrl.statusMessage">
+
   <div class="container">
     <div class="row">
       <div class="col-xs-6">
-        <high-weight-genes signature-id="{{ctrl.id}}" genes="ctrl.genes">
-        </high-weight-genes>
+        <participatory-genes signature-id="{{ctrl.id}}" genes="ctrl.genes"
+            selected-participation-type="ctrl.selectedParticipationType">
+        </participatory-genes>
```
mhuyck: Ok, here is the template where that wrapping I suggested would make this more readable:

```html
        <participatory-genes
          signature-id="{{ctrl.id}}"
          genes="ctrl.genes"
          selected-participation-type="ctrl.selectedParticipationType">
        </participatory-genes>
```

I like this way because it's easier to scan visually.

interface/src/app/signature/signature.tpl.html
```diff
@@ -21,7 +34,8 @@ <h3 class="text-warning" ng-bind="ctrl.statusMessage"></h3>
     </div>
     <div class="row">
       <div class="col-xs-6">
-        <enriched-genesets organism="{{ctrl.organism}}" genes="ctrl.genes">
+        <enriched-genesets organism="{{ctrl.organism}}" genes="ctrl.genes"
+            selected-participation-type="ctrl.selectedParticipationType">
```
mhuyck: I'd wrap as describe above.

rzelayafavila:
Ok thanks @mhuyck - that is a great point with the watch statement - I will use the onChange function instead. Also the formatting changes you mention make sense, so I will change those this morning. Thanks!

mhuyck:
The syntax for invoking a callback is a little funky (at least in Angular 1.x). For an example, see what I did in `app/mlmodel/`. The parameter is declared in `mlModelComponents.js`, just as you have it, and it is invoked by the component in the `ng-change` parameter of the `<select>` in `selector.tpl.html`: https://github.com/greenelab/adage-server/blob/master/interface/src/app/mlmodel/selector.tpl.html .

Note especially that this invocation passes parameters as an object, which behaves as an associative array: `{value: $ctrl.selectedMlModel}`. Without those curly braces, parameters are not passed properly.

interface/src/app/signature/signature.spec.js
rzelayafavila: Sure! I can do that to make it more readable.

interface/src/app/signature/signature.spec.js
rzelayafavila: Thanks!

rzelayafavila:
Hi @mhuyck - I actually did a deep dive today to try to figure out the best solution for the ``onChange`` vs. ``$watch`` debate. It turns out that there isn't really a good solution using ``ng-change`` and binding that to a function on the parent scope, because the code we need to run is actually in another child directive of that parent controller (a sibling directive if you will!). There isn't a really straightforward way of handling a binding in that direction, and it makes sense, because you want parent and children directives/controllers to be loosely bound, and the parent should not need to know anything about the child directive and the code it contains. One approach would be to put all of that code in the parent controller (SignatureCtrl) and bind all of the elements that come out of that code to the child directive, but I have a strong feeling that all of that code does not belong in that parent controller.

Some solutions online suggest using ``$broadcast`` from a parent controller to tell any and all child controllers that something has happened (https://stackoverflow.com/questions/19038778/angularjs-how-to-call-child-scope-function-in-parent-scope), and then use ``$scope.$on`` in the child controllers to run some code when receiving that signal. However, I know we have turned away from this approach in the past. Do you think this is a better approach than the ``$watch`` approach though? For now I decided to remove the ``onChange`` binding where it was not being used, and stick with the ``$watch``.

rzelayafavila:
I think I have made changes for all the other stuff you mentioned though @mhuyck!

mhuyck:
Got it. I hadn't noticed there was a sibling involved... that does make it a little less obvious how to handle. I'm totally fine with keeping it as a `$watch` for now, since that's working. My guess is that a `$broadcast` approach would be a step backwards, both from a performance perspective and from a code maintainability perspective... that way just feels akin to using a global variable to solve a problem that should be handled by a local variable.

rzelayafavila:
Ok, thank you @mhuyck!

PR #161, opened 2017-06-06
Workflow Redesign

dongbohu:
This is by far the biggest PR that I have ever created in adage. It redesigns the workflow and forces users to select a valid machine learning model before any data can be displayed on the UI. The demo site is here:
http://165.123.67.202/#/home

Here is a summary of how this functionality is implemented (for reviewers who will read the code):

* A `MlModelTracker` service is defined in `app/utils.js` to keep track of currently selected machine learning model.

* To display the current machine learning model (which could be "unset") on web UI, a `mlModelView` component is defined in `app/mlmodel/mlModelComponents.js`. Note that it is for display ONLY.

* For URLs which include `mlmodel` parameter, the input `mlmodel` parameter is validated by another component `mlModelValidator`, which is defined in `app/mlmodel/mlModelComponents.js` too.

* Most HTML template files that are related to a ui-router state (except `Home`, `About`, `Download`, whose contents are static) are using `<ml-model-view>` tag to display the current model, and `<ml-model-validator>` tag to validate the `mlmodel` parameter in the URL. `signature.tpl.html` is another exception: it only uses  `<ml-model-view>`, because the URL doesn't need `mlmodel`. The input signature ID already determines the mlmodel.

* Any page that is using `<ml-model-validator>` also passes in a `isValidModel` boolean flag, which will be updated by  `<ml-model-validator>`. Initially the flag is false, if this flag is updated to true, then the input `mlmodel` is valid, and the other widgets in the HTML template will show up; otherwise these widgets will be hidden. That is why the main body of these HTML templates are enclosed by this:
```
<div ng-if="ctrl.isValidModel">
...
</div>
```

* I also cleaned up the HTML templates that correspond to the tabs on navbar so that their headers are all in this format:
```
<div class="page-header">
  <h1>Tab Title</h1>
  <ml-model-view>...</ml-model-view>
  <ml-model-validator ...></ml-model-validator>
</div> 
```
(As stated earlier, some pages don't have `<ml-model-view>` and `<ml-model-validator ...>` tags.)

If you have any questions or concerns, please let me know.

### DIFF
--- a/interface/src/app/about/about.tpl.html
+++ b/interface/src/app/about/about.tpl.html
@@ -1,8 +1,8 @@
 <div class="page-header">
   <h1>About ADAGE</h1>
 </div>
-<h3>
-  <p>
+
+<h4>
     The ADAGE webserver allows users without programming expertise to analyze
     signatures extracted from large compendia of gene expression data via
     unsupervised machine learning methods. We primarily designed the server to
@@ -12,10 +12,12 @@
     <a href="http://biorxiv.org/content/early/2017/04/10/078659">eADAGE</a>;
     however, the server was designed to be general and can be used for other
     types of machine learning models as well.
-  <!--
-  <p>
+</h4>
+
+<!--
+<h4>
     If you use the server for any type of model, please cite the
     [ADAGE Signature Analysis paper](Link TBD). Please also cite the paper that
     reports the machine learning model that you use.
-    -->
-</h3>
+</h4>
+-->
--- a/interface/src/app/about/about.tpl.html
+++ b/interface/src/app/about/about.tpl.html
@@ -2,7 +2,7 @@
   <h3>About ADAGE</h3>
 </div>
 
-<p>
+<p class="h4">
   The <strong>ADAGE</strong> (<strong>A</strong>nalysis using
   <strong>D</strong>enoising <strong>A</strong>utoencoders of
   <strong>G</strong>ene <strong>E</strong>xpression) web server allows users
@@ -18,8 +18,8 @@
 
 <!--
 <p>
-    If you use the server for any type of model, please cite the
-    [ADAGE Signature Analysis paper](Link TBD). Please also cite the paper that
-    reports the machine learning model that you use.
+  If you use the server for any type of model, please cite the
+  [ADAGE Signature Analysis paper](Link TBD). Please also cite the paper that
+  reports the machine learning model that you use.
 </p>
 -->
--- a/interface/src/app/about/about.tpl.html
+++ b/interface/src/app/about/about.tpl.html
@@ -1,8 +1,6 @@
-<div>
-  <h1 class="page-header">
-    About ADAGE
-  </h1>
-  <p>
-    This is where some explanatory text about ADAGE should go.
-  </p>
+<div class="page-header">
+  <h1>About ADAGE</h1>
 </div>
+<p>
+  This is where some explanatory text about ADAGE should go.
+</p>
--- a/interface/src/app/about/about.tpl.html
+++ b/interface/src/app/about/about.tpl.html
@@ -1,6 +1,21 @@
 <div class="page-header">
   <h1>About ADAGE</h1>
 </div>
-<p>
-  This is where some explanatory text about ADAGE should go.
-</p>
+<h3>
+  <p>
+    The ADAGE webserver allows users without programming expertise to analyze
+    signatures extracted from large compendia of gene expression data via
+    unsupervised machine learning methods. We primarily designed the server to
+    support models constructed by
+    <a href="http://msystems.asm.org/content/1/1/e00025-15">ADAGE</a>
+    and
+    <a href="http://biorxiv.org/content/early/2017/04/10/078659">eADAGE</a>;
+    however, the server was designed to be general and can be used for other
+    types of machine learning models as well.
+  <!--
+  <p>
+    If you use the server for any type of model, please cite the
+    [ADAGE Signature Analysis paper](Link TBD). Please also cite the paper that
+    reports the machine learning model that you use.
+    -->
+</h3>
--- a/interface/src/app/about/about.tpl.html
+++ b/interface/src/app/about/about.tpl.html
@@ -1,8 +1,8 @@
 <div class="page-header">
-  <h1>About ADAGE</h1>
+  <h3>About ADAGE</h3>
 </div>
 
-<h4>
+<p>
     The ADAGE webserver allows users without programming expertise to analyze
     signatures extracted from large compendia of gene expression data via
     unsupervised machine learning methods. We primarily designed the server to
@@ -12,12 +12,12 @@
     <a href="http://biorxiv.org/content/early/2017/04/10/078659">eADAGE</a>;
     however, the server was designed to be general and can be used for other
     types of machine learning models as well.
-</h4>
+</p>
 
 <!--
-<h4>
+<p>
     If you use the server for any type of model, please cite the
     [ADAGE Signature Analysis paper](Link TBD). Please also cite the paper that
     reports the machine learning model that you use.
-</h4>
+</p>
 -->
--- a/interface/src/app/about/about.tpl.html
+++ b/interface/src/app/about/about.tpl.html
@@ -3,15 +3,17 @@
 </div>
 
 <p>
-    The ADAGE webserver allows users without programming expertise to analyze
-    signatures extracted from large compendia of gene expression data via
-    unsupervised machine learning methods. We primarily designed the server to
-    support models constructed by
-    <a href="http://msystems.asm.org/content/1/1/e00025-15">ADAGE</a>
-    and
-    <a href="http://biorxiv.org/content/early/2017/04/10/078659">eADAGE</a>;
-    however, the server was designed to be general and can be used for other
-    types of machine learning models as well.
+  The <strong>ADAGE</strong> (<strong>A</strong>nalysis using
+  <strong>D</strong>enoising <strong>A</strong>utoencoders of
+  <strong>G</strong>ene <strong>E</strong>xpression) web server allows users
+  without programming expertise to analyze signatures extracted from large
+  compendia of gene expression data via unsupervised machine learning methods.
+  We primarily designed the server to support models constructed by
+  <a href="http://msystems.asm.org/content/1/1/e00025-15">ADAGE</a>
+  and
+  <a href="http://biorxiv.org/content/early/2017/04/10/078659">eADAGE</a>;
+  however, the server was designed to be general and can be used for other
+  types of machine learning models as well.
 </p>
 
 <!--
--- a/interface/src/app/analyze/analysis/analysis.js
+++ b/interface/src/app/analyze/analysis/analysis.js
@@ -15,7 +15,7 @@ angular.module('adage.analyze.analysis', [
 
 .config(['$stateProvider', function($stateProvider) {
   $stateProvider.state('analysis-detail', {
-    url: '/analysis-detail',
+    url: '/analysis-detail?mlmodel',
     views: {
       main: {
         controller: 'AnalysisCtrl',
@@ -26,8 +26,18 @@ angular.module('adage.analyze.analysis', [
   });
 }])
 
-.controller('AnalysisCtrl', ['$scope', '$log', '$q', '$state', 'SampleBin',
-function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
+.controller('AnalysisCtrl', ['$scope', '$log', '$q', '$state', '$stateParams',
+  'SampleBin',
+function AnalysisCtrl($scope, $log, $q, $state, $stateParams, SampleBin) {
+  $scope.isValidModel = false;
+  // Do nothing if mlmodel in URL is falsey. The error will be taken
+  // care of by "<ml-model-validator>" component.
+  if (!$stateParams.mlmodel) {
+    return;
+  }
+
+  $scope.modelInUrl = $stateParams.mlmodel;
+  SampleBin.getActivityForSampleList($scope.modelInUrl);
   $scope.analysis = {
     status: '',
     // TODO these exampleCols are temporarily hard-coded until a column chooser
@@ -38,12 +48,7 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
       {'typename': 'medium'},
       {'typename': 'temperature'},
       {'typename': 'treatment'}
-    ],
-    updateMlModel: function(value) {
-      if (value !== null && value.id !== null) {
-        SampleBin.getActivityForSampleList(value.id);
-      }
-    }
+    ]
   };
 
   // give our templates a way to access the SampleBin service
@@ -58,11 +63,7 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
   };
 
   $scope.showVolcanoPlot = function() {
-    if (SampleBin.selectedMlModel.id === null) {
-      $scope.analysis.status = 'Please select a machine learning model.';
-      return;
-    }
-    $state.go('volcano', {'mlmodel': SampleBin.selectedMlModel.id});
+    $state.go('volcano', {'mlmodel': $scope.modelInUrl});
   };
 
   // these options are important for making ngSortable work with tables
--- a/interface/src/app/analyze/analysis/analysisDetail.tpl.html
+++ b/interface/src/app/analyze/analysis/analysisDetail.tpl.html
@@ -15,59 +15,60 @@
 
 <div ng-if="isValidModel">
 
-<status-bar msg-text="analysis.status"></status-bar>
-<!-- TODO need a tool to select and order annotation columns -->
-<table class="table table-hover">
-  <thead>
-    <tr>
-      <th> </th>
-      <th>
-        <span class="row-control">
-          <i class="fa fa-bars" aria-hidden="true"></i>
-        </span>
-        ML Data Source
-      </th>
-      <th>id</th>
-      <th>Sample Name</th>
-      <th ng-repeat="at in analysis.exampleCols">{{
-        at.typename
-      }}</th>
-    </tr>
-  </thead>
-  <tbody ng-model="sb.heatmapData.samples" as-sortable="sortableOptions">
-    <tr ng-repeat="id in sb.heatmapData.samples"
-        ng-mouseover="showControls = true" ng-mouseleave="showControls = false"
-        ng-class="'sample-' + sb.sampleToGroup[id]" as-sortable-item>
-      <td class="row-control">
-        <input type="radio" ng-model="sb.sampleToGroup[id]"
-            name="{{id}}-group" value="base-group"
-            class="sample-base-group" aria-label="group a">
-        <input type="radio" ng-model="sb.sampleToGroup[id]"
-            name="{{id}}-group" value="comp-group"
-            class="sample-comp-group" aria-label="group b">
-        <input type="radio" ng-model="sb.sampleToGroup[id]"
-            name="{{id}}-group" value="other"
-            class="sample-other" aria-label="other">
-        &nbsp;
-        <button type="button" class="btn row-control"
-            ng-click="sb.removeSample(id)" aria-label="Delete">
-          <i class="fa fa-times" aria-hidden="true"></i>
-        </button>
-      </td>
-      <td as-sortable-item-handle>
-        <span class="row-control">
-          <i class="fa fa-bars" aria-hidden="true"></i>
-        </span>
-        {{sb.sampleData[id].ml_data_source}}
-      </td>
-      <td>{{id}}</td>
-      <td>{{sb.sampleData[id].name}}</td>
-      <td ng-repeat="at in analysis.exampleCols">{{
-        sb.sampleData[id].annotations[at.typename]
-      }}</td>
-    </tr>
-  </tbody>
-</table>
+  <status-bar msg-text="analysis.status"></status-bar>
+  <!-- TODO need a tool to select and order annotation columns -->
+  <table class="table table-hover">
+    <thead>
+      <tr>
+        <th> </th>
+        <th>
+          <span class="row-control">
+            <i class="fa fa-bars" aria-hidden="true"></i>
+          </span>
+          ML Data Source
+        </th>
+        <th>id</th>
+        <th>Sample Name</th>
+        <th ng-repeat="at in analysis.exampleCols">{{
+          at.typename
+          }}</th>
+      </tr>
+    </thead>
+    <tbody ng-model="sb.heatmapData.samples" as-sortable="sortableOptions">
+      <tr ng-repeat="id in sb.heatmapData.samples"
+          ng-mouseover="showControls = true"
+          ng-mouseleave="showControls = false"
+          ng-class="'sample-' + sb.sampleToGroup[id]" as-sortable-item>
+        <td class="row-control">
+          <input type="radio" ng-model="sb.sampleToGroup[id]"
+                 name="{{id}}-group" value="base-group"
+                 class="sample-base-group" aria-label="group a">
+          <input type="radio" ng-model="sb.sampleToGroup[id]"
+                 name="{{id}}-group" value="comp-group"
+                 class="sample-comp-group" aria-label="group b">
+          <input type="radio" ng-model="sb.sampleToGroup[id]"
+                 name="{{id}}-group" value="other"
+                 class="sample-other" aria-label="other">
+          &nbsp;
+          <button type="button" class="btn row-control"
+                  ng-click="sb.removeSample(id)" aria-label="Delete">
+            <i class="fa fa-times" aria-hidden="true"></i>
+          </button>
+        </td>
+        <td as-sortable-item-handle>
+          <span class="row-control">
+            <i class="fa fa-bars" aria-hidden="true"></i>
+          </span>
+          {{sb.sampleData[id].ml_data_source}}
+        </td>
+        <td>{{id}}</td>
+        <td>{{sb.sampleData[id].name}}</td>
+        <td ng-repeat="at in analysis.exampleCols">{{
+          sb.sampleData[id].annotations[at.typename]
+          }}</td>
+      </tr>
+    </tbody>
+  </table>
 
   <p>Note: scroll to the right to see all signatures in the heatmap. Zooming
      feature to come in a future update.</p>
--- a/interface/src/app/analyze/analysis/analysisDetail.tpl.html
+++ b/interface/src/app/analyze/analysis/analysisDetail.tpl.html
@@ -3,11 +3,11 @@
 -->
 
 <div class="page-header">
-  <h1>Samples
+  <h3>Samples
     <small>(drag <i class="fa fa-bars" aria-hidden="true"></i> to reorder,
       click <i class="fa fa-times" aria-hidden="true"></i> to delete,
       choose <input type="radio" disabled> to set group)</small>
-  </h1>
+  </h3>
   <ml-model-view></ml-model-view>
   <ml-model-validator model-id="modelInUrl" is-valid-model="isValidModel">
   </ml-model-validator>
--- a/interface/src/app/analyze/analysis/analysisDetail.tpl.html
+++ b/interface/src/app/analyze/analysis/analysisDetail.tpl.html
@@ -1,11 +1,20 @@
 <!--
   TODO each sample should be built using a reusable Angular directive
 -->
-<h1>Samples
-  <small>(drag <i class="fa fa-bars" aria-hidden="true"></i> to reorder,
-    click <i class="fa fa-times" aria-hidden="true"></i> to delete,
-    choose <input type="radio" disabled> to set group)</small>
-</h1>
+
+<div class="page-header">
+  <h1>Samples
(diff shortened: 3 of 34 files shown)


dongbohu:
Don't worry about the codeclimate issues. I will fix them later.

dongbohu:
@rzelayafavila and @mhuyck: To make this review speedy, you may want to start from the modules that you wrote originally. Matt may not necessarily need to review the modules that were written by Rene, and vice versa. But the modules whose original author was me should be reviewed by at least one of you.

interface/src/app/home/home.tpl.html
```diff
@@ -1,6 +1,32 @@
-<div>
+<div class="page-header">
   <h1>ADAGE: Home</h1>
-  <p class="lead">
-    This shell is the framework into which the ADAGE server will grow.
-  </p>
+</div>
+
+<h3>Step 1: Choose a machine leanring model</h3>
```
rzelayafavila: Typo: ``learning`` instead of ``leanring``

interface/src/app/mlmodel/mlModelComponents.js
```diff
@@ -30,4 +31,45 @@ angular.module('adage.mlmodel.components', [
...
 })
 
+.component('mlModelView', {
+  templateUrl: 'mlmodel/view.tpl.html',
+  controller: ['GlobalModelInfo', function(GlobalModelInfo) {
+    this.modelInfo = GlobalModelInfo;
+  }]
+})
+
+.component('mlModelValidator', {
```
rzelayafavila: I'm not sure if this should be a component or just a service that is used by other components. What are the advantages of having it be a component that gets injected at the template level?

interface/src/app/about/about.tpl.html
```diff
@@ -1,8 +1,6 @@
...
-  </h1>
-  <p>
-    This is where some explanatory text about ADAGE should go.
-  </p>
+<div class="page-header">
+  <h1>About ADAGE</h1>
 </div>
+<p>
+  This is where some explanatory text about ADAGE should go.
+</p>
```
dongbohu: Good idea. Can we copy/paste some texts from @tj8901nm paper and put it here?

interface/src/app/home/home.tpl.html
dongbohu: Thanks.

interface/src/app/about/about.tpl.html
cgreene: Text incoming!

interface/src/app/mlmodel/mlModelComponents.js
dongbohu: This component not only validates the input ML model ID, it is also responsible for generating an error message on the web UI. Please read the long comments that I wrote 6 days ago and see whether it makes sense to you.

interface/src/app/about/about.tpl.html
cgreene: How about:

`The ADAGE webserver allows users without programming expertise to analyze signatures extracted from large compendia of gene expression data via unsupervised machine learning methods. We primarily designed the server to support models constructed by [ADAGE](https://doi.org/10.1128/mSystems.00025-15) and [eADAGE](https://doi.org/10.1101/078659); however, the server was designed to be general and can be used for other types of machine learning models as well.

cgreene: Guess the code block doesn't format that nicely. I'll remove the backticks.

dongbohu:
About the variable name `GlobalModelInfo`, originally I had `ml` in its name, but it seemed a little bit too long, so I deleted it. I agree that the current name is a little misleading (which is one of the reasons why I wrote a long post six days ago right after I submitted this PR). If you guys prefer having "Ml", surely I can add it back.

interface/src/app/about/about.tpl.html
dongbohu: @casey: I copied the texts above but commented out the second paragraph. Once the ADAGE Signature Analysis paper's link it available, we can set it and show the second paragraph.
http://165.123.67.202/#/about

cgreene: Sounds good :+1:

interface/src/app/mlmodel/mlModelComponents.js
rzelayafavila: Yes, I read these comments, I was just wondering whether this was the best approach, or we should split the functionality of the mlModelValidator into an mlModelValidator service and a component that displays an error message. I am ok with this approach if you feel it is the best way.

rzelayafavila: Also, I can't seem to find the mlmodel/validator.tpl.html file for this component - is it included in this PR?

dongbohu: I see. This component is small. Unless other modules may also need a service to validate a ML model, I tend to keep it as is.

dongbohu: @rzelayafavila: My bad. I added the two html template files.

mhuyck:
Using the test server: I type a typical search (`pa14`) using the new page flow and get results back, but the console log is filled with 415 errors. Selecting an experiment or sample from the search results leads to an error "Query Failed".

mhuyck:
More info: the "Query Failed" appears to arise from `http://165.123.67.202/api/v0/experiment/?accession=E-GEOD-9592`, which is throwing a `BAD REQUEST` error 400.

mhuyck:
I think this must be the issue: a service was recently added in `src/app/signature/resources.js` with a new factory named `Experiment`, which is masking the original `Experiment` service in `src/app/analyze/experiment/experiment.js`.

The quick and dirty fix is to rename the new `Experiment` something like `SignatureExperiment` everywhere it's needed. That should unmask the service that the Analyze tab is looking for.

A longer term resolution to this issue is to complete the refactoring I described in #76 and reinforce a more component-based architecture. That way it would be obvious that all `Experiment`-related code belongs in the source tree under a reusable component by that name.

dongbohu:
@mhuyck: I think the error you found is because of a new commit that I didn't have in my local repo.

mhuyck:
I'm looking at the demo server on your machine @dongbohu, so it must be in your repo.

rzelayafavila:
@mhuyck @dongbohu - Yeah, I was actually wondering if the not allowing filtering by 'accession' (which came back in that error response) was intentional or not?

dongbohu:
@mhuyck: I think I fixed it. Can you confirm?

mhuyck:
What's happening to produce that "filtering by accession" error is that the new `Experiment` service is taking the query from the Sample or Experiment page and passing the `accession` as a parameter to the `.../experiment/` REST API. The original `Experiment` service passes that parameter to the "detail" endpoint, which is at `.../experiment/accession`, and then Tastypie knows how to handle it. There are other features on that currently masked `Experiment` service that are unavailable, so we definitely need to do some form of change to the new service to unmask the original.

The changeset that introduced the new service is here:

https://github.com/greenelab/adage-server/commit/05542515f734a80c6f562f7dc172a481575876f6

mhuyck:
Still getting 415 errors, but not the "Query Failed" now. What did you change, @dongbohu?

rzelayafavila:
Ah gotcha @mhuyck, thanks

dongbohu:
@mhuyck: I renamed `Experiment` service in `signature/resources.js` into `SignatureExperiment` so that it won't overwrite the `Experiment` service in your `analyze` module.

mhuyck:
Okay, good. Let me see if I can figure out where those other errors are coming from.

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -38,12 +48,7 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
       {'typename': 'medium'},
       {'typename': 'temperature'},
       {'typename': 'treatment'}
-    ],
```
mhuyck: Need to restore this line

interface/src/app/analyze/analysis/analysis.js
```diff
@@ -38,12 +48,7 @@ function AnalysisCtrl($scope, $log, $q, $state, SampleBin) {
...
-    updateMlModel: function(value) {
-      if (value !== null && value.id !== null) {
-        SampleBin.getActivityForSampleList(value.id);
-      }
-    }
+    ]
```
mhuyck: and delete this line. That should make the `...detail.showing` errors disappear.

mhuyck: No, wait... this is wrong. I'm reading the diff backwards! Let me keep looking.

dongbohu: I got a lot of errors like this on the console:
```
TypeError: Cannot read property 'showing' of undefined
    at Object.item_style (adage-0.7.0b.js:41)
    at fn (eval at compile (adage-0.7.0b.js:30), <anonymous>:4:345)
    at o.$digest (adage-0.7.0b.js:28)
    at o.$apply (adage-0.7.0b.js:28)
    at i (adage-0.7.0b.js:27)
    at u (adage-0.7.0b.js:27)
    at XMLHttpRequest.x.onload (adage-0.7.0b.js:27)
```
Does this ring a bell for you?

mhuyck: This is what I'm investigating. Never seen it before and it is not happening on my local instance. I'm not seeing any changes in the code directly related to that `detail.showing` error in this PR, so I'm looking at some more comprehensive diffs to see what has changed.

mhuyck: The diff does not tell me anything I don't already know, so I can't yet see what's causing those errors. I'm still investigating. Pulling up a copy of @dongbohu's fork so I can try to reproduce here.

dongbohu: Note that in my new version, the `updateMlModel()` function has been removed completely. Instead, 
```
SampleBin.getActivityForSampleList($scope.modelInUrl);
``` 
is called directly before $scope.analysis is assigned.

mhuyck: I noticed that, but I don't think this chain of errors is related to that. I'm following up a lead I just found. Standby...

mhuyck: Ugh. I'm getting `npm` errors trying to get my local clone of your dev branch to build.

@dongbohu: Can you try something on your instance to see if it helps? Let me point out the line I'd like to change...

dongbohu: @mhuyck: Sure.

interface/src/app/analyze/analyze.js
```diff
@@ -31,9 +31,19 @@ angular.module('adage.analyze', [
...
+.controller('AnalyzeCtrl', ['$scope', '$stateParams', '$log', '$location',
   '$anchorScroll', 'Sample',
-  function AnalyzeCtrl($scope, $log, $location, $anchorScroll, Sample) {
+  function AnalyzeCtrl($scope, $stateParams, $log, $location, $anchorScroll,
+                       Sample) {
+    $scope.isValidModel = false;
+    // Do nothing if mlmodel in URL is falsey. The error will be taken
+    // care of by "<ml-model-validator>" component.
+    if (!$stateParams.mlmodel) {
+      return;
```
mhuyck: Try removing this line. I think that by not allowing the controller to finish executing, we're preventing the initialization code below from ever running.

dongbohu: I commented out this block but get the same error.

mhuyck: Rats. I'll need to try again to get your fork to build. If I can trace into the code I can see what's happening.

dongbohu: Line 53 (in my PR) is referring to `$scope.detail.showing` directly before `$scope.detail` is assigned to any valule, do you think this is the reason of that error?

dongbohu: I wonder whether we can change line 53 from:
```
 if ($scope.detail.showing &&
```
into:
```
 if ($scope.detail && $scope.detail.showing &&
```

dongbohu: @mhuyck: I made the change and the error is gone, but I am not sure whether this change introduces any other issues.

mhuyck: Nice! I was able to get a build by copying `node_modules` from my working dev instance. Let me confirm if that does the trick here.

interface/src/app/analyze/analyze.tpl.html
```diff
@@ -1,44 +1,52 @@
...
-        </li>
-      </ul>
+<div class="page-header">
+  <h1>Analyze</h1>
+  <ml-model-view></ml-model-view>
+  <ml-model-validator model-id="modelInUrl" is-valid-model="isValidModel">
+  </ml-model-validator>
+</div>
+
+<div ng-if="isValidModel">
```
mhuyck: This must be the source of the problem. I think the `ng-if` here prevents the template from referring to `detail.showing`, which must have somehow been initializing that variable.

Weird bug!

dongbohu: I figured out another solution: on line 8 (my PR) of `analyze.tpl.html`, use `ng-show` instead of `ng-if`, it works too. You can test it.

Rene had a similar issue sometime ago. I think this is one of the key differences between `ng-show` and `ng-if`.

rzelayafavila: Right - that makes sense. ng-show will still include the element in the DOM, but just apply some kind of 'hidden' CSS styling to it, whereas ng-if will actually remove the element from the DOM if the condition is false.

dongbohu: @mhuyck: I changed "ng-if" to ng-show" and added some comments. Code has been committed.

dongbohu:
@mhuyck: Please continue playing with the web UI (if you are not done) and let me know any problems you found. This is very very helpful. Thank you.

interface/src/app/analyze/analyze.tpl.html
mhuyck: Ok, checking.

mhuyck:
@dongbohu: can you do a rebase on this code and update your test instance? I'm seeing the error in the Volcano Plot that I just fixed. This would be an excellent test that my GET to POST conversion there works on your database.

interface/src/app/about/about.tpl.html
```diff
@@ -1,8 +1,21 @@
...
-  <h1 class="page-header">
-    About ADAGE
-  </h1>
-  <p>
-    This is where some explanatory text about ADAGE should go.
-  </p>
+<div class="page-header">
+  <h1>About ADAGE</h1>
 </div>
+<h3>
```
mhuyck: This `h3` tag is causing all of the text to appear bold, as if the page were yelling. Can we reserve these tags for actual headers (brief titles) only? I recommend simple `<p>` tags instead.

interface/src/app/about/about.tpl.html
```diff
@@ -1,8 +1,21 @@
...
+  <p>
+    The ADAGE webserver allows users without programming expertise to analyze
+    signatures extracted from large compendia of gene expression data via
+    unsupervised machine learning methods. We primarily designed the server to
+    support models constructed by
+    <a href="http://msystems.asm.org/content/1/1/e00025-15">ADAGE</a>
+    and
+    <a href="http://biorxiv.org/content/early/2017/04/10/078659">eADAGE</a>;
+    however, the server was designed to be general and can be used for other
+    types of machine learning models as well.
```
mhuyck: Please include a closing `</p>` tag as well. There are some circumstances when it is allowed to be omitted, but it is much easier to write correct HTML by not doing that.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

interface/src/app/about/about.tpl.html
```diff
@@ -1,8 +1,21 @@
...
+  <!--
+  <p>
+    If you use the server for any type of model, please cite the
+    [ADAGE Signature Analysis paper](Link TBD). Please also cite the paper that
+    reports the machine learning model that you use.
+    -->
```
mhuyck: closing `</p>` tag.

mhuyck:
@dongbohu: I think I see some of the changes from my code so you must have done a rebase. Thank you! I am still getting an error, though, which would happen if you didn't restart the Django web service. On production, the command would be `sudo supervisorctl restart adage`. This will allow me to fully test the Volcano Plot.

dongbohu:
@mhuyck: Thanks. I have restarted the backend.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -1,11 +1,20 @@
...
+    <small>(drag <i class="fa fa-bars" aria-hidden="true"></i> to reorder,
+      click <i class="fa fa-times" aria-hidden="true"></i> to delete,
+      choose <input type="radio" disabled> to set group)</small>
+  </h1>
+  <ml-model-view></ml-model-view>
+  <ml-model-validator model-id="modelInUrl" is-valid-model="isValidModel">
+  </ml-model-validator>
+</div>
+
+<div ng-if="isValidModel">
```
mhuyck: Can we make this an `ng-show` and move it down between line 70 and 71? Putting this all the way up here removes features from the page that don't depend upon having an MLModel selected.

interface/src/app/utils.js
```diff
@@ -17,4 +17,24 @@ angular.module('adage.utils', [])
 
 // Number of digits in activity value shown on frontend:
 .constant('ActivityDigits', 5)
+
+// Service of global machine learning model, which is shared by multiple
+// modules.
```
mhuyck: I understand why you've put this here, but I think my preference would be to reinforce the component architecture by defining this service in a file within the `src/app/mlmodel` directory and then list that file as a *dependency* for `adage.utils` above.

interface/src/app/about/about.tpl.html
dongbohu: I was using <p> initially, but the font size seems a little too small, and it looks bad with the <h1> header.

interface/src/app/about/about.tpl.html
dongbohu: Okay.

interface/src/app/about/about.tpl.html
dongbohu: Okay.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: Sure.

interface/src/app/utils.js
dongbohu: I thought of this option. But that means I have to either create a new module in `mlmodel` directory, or put it in `mlmodel/mlComponents.js` or `mlmodel/resources.js`, neither filename seems to be a good fit. On the other hand, I feel that it is a service that is shared by all modules, so I put it in `app/utils.js`.

interface/src/app/about/about.tpl.html
dongbohu: Tried `<p>` and the font is definitely too small to me. I changed it to `<h4>`.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: Question: If `mlmodel` in the URL is absent or invalid, do we really want to show data on the web interface? I thought we decided to force the user to select a machine learning model before any data is shown. What do you think, @casey?

cgreene: @dongbohu - tag here is @cgreene / @casey bugs someone else ;)

I'm imagining a situation where a user searches and selects datasets from multiple organisms. How will we show those results? I can't think of a sensible way to do that. Restricting by `mlmodel` first prevents us from getting stuck in a situation where the user is selecting multiple experiments that we can't sensibly show together (if we facet the search results by the `mlmodel`'s `organism`. Though it's not a problem with one organism in the production instance, it will be. Seems good to handle this now.

interface/src/app/utils.js
dongbohu: Creating a new module in `mlmodel` directory for this service will affect many modules on front end. Honestly I don't feel comfortable to do it before I leave in a few hours. If you have a strong preference on it, please add an issue and we can deal with it later. Thanks.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: @cgreene Thanks. Will use the correct tag.

cgreene:
Random question: Is our <p> styling out of line with the headers that we've been using? I see the <h3> / <h4> discussions b/w @mhuyck and @dongbohu, and it seems to me that if we need our text to be a header then maybe we should change the CSS. I'm wondering if we should go with <p> for its semantic meeting and create an issue to reexamine font sizing in our CSS.

dongbohu:
@casey: Are you talking about the styling files in in `src/less/` directory? I checked the files there but don't see any spec on text in `<p>`.

Back to the text size in `about` page, I just feel that the paragraph is so short, but the regular font seems a little too small (especially when it is compared with the `<h1>` title).

cgreene:
@dongbohu : what about a lower level `h` title and then `p` text? I like that `p` has a meaning and it seems like the appropriate meaning in this case so I am reluctant to move to a header. We can continue to expand the `about` page (actually - would it make sense to merge with the "Help" page?).

dongbohu:
@cgreene: I consider `about` as a place that holds a summary of ADAGE software (and maybe some tech information such as the backend/frontend frameworks that we are using, if necessary), and `help` page a place for end users. It also depends on how much information we want to put on these two pages. If `about` will eventually contain only two short paragraphs, maybe we can merge them together; otherwise we probably should keep them separate.

dongbohu:
I changed the title from `h1` to `h3`, and the text is regular.
http://165.123.67.202/#/about
What do you think?

cgreene:
At the moment, I only have two short paragraphs planned. I think that we will have to see what our users want to see when they click there. Maybe @mhuyck can ask in his next time with them?

dongbohu:
One thing that bothers me is that the titles of all the other pages are using `h1`, but `about` is using `h3` now.

dongbohu:
If nobody objects, I am going to change all titles from `h1` to `h3`.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: I see the point now. In this case, we have a bunch more work to do. Nothing in the Experiment or Sample models connects them to an organism, so there is no way for the current system design to enforce consistency on what is added to the sample list.

interface/src/app/utils.js
mhuyck: Oh, I thought it would just be a small change. If we make the service a dependency of `adage.utils` then it will be available everywhere as if it were included right here. Maybe I'm missing something. I will make an issue.

mhuyck:
The place to tweak the CSS styles is in the appropriate `.less` file, but now is not the time to monkey with the CSS. It will take some fiddling to get right. I'll file an issue for us to follow up on semantic HTML markup and organized CSS formats.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
cgreene: @mhuyck : gotcha - yea - a "sample" should probably tie back to an organism. @kurtwheeler's data refinery should have the organism information stored, so this will be something that i'm pretty sure we can get input

interface/src/app/utils.js
dongbohu: @mhuyck : Practically your approach will work, but I would prefer adding the dependency of this new module in the module that uses this `MlModelTracker` directly, otherwise when you do unit test, you will have to always remember to include `adage.utils` module, even though you may not really need it in the test.

dongbohu:
Do you guys (especially @cgreene) think it is a good idea to add the full name of ADAGE on `about` page? so that the first sentense will be something like:
```
The ADAGE (Analysis using Denoising Autoencoders of Gene Expression) web server allows users ...
```
I remember when I first joined the lab, I was very confused about what it really stands for.

interface/src/app/utils.js
mhuyck: If I want to add a feature or fix a bug related to the `MLModelTracker`, the logical place to look for that code is in the `app/mlmodel` directory. I would never think to look in `app/utils.js`. This is part of what I was trying to advocate for in our discussion last Wednesday.

There is nothing wrong with declaring that `adage.utils` depends upon the module that contains `MLModelTracker` *and* having each page include the same dependency. The Angular dependency injection system has no trouble sorting that out.

Doing it the way I suggest will help us to avoid bugs like the `'accession' field does not allow filtering` one we found earlier. The structure of our repository should be intuitive enough that every developer knows without question which part of the source tree holds what code.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -1,71 +1,75 @@
...
-      <td ng-repeat="at in analysis.exampleCols">{{
-        sb.sampleData[id].annotations[at.typename]
-      }}</td>
-    </tr>
-  </tbody>
-</table>
-<p>To proceed with further analysis, please select a machine learning model.</p>
-<label>Machine learning model:</label>
-<ml-model-selector
-  selected-ml-model="sb.selectedMlModel"
```
mhuyck: I expected that taking the `SampleBin` (`sb`, here) out of the loop on the `selectedMlModel` would cause serious trouble, but I'm not seeing it break on the demo server.

Just the same, I'm going to file an issue to update the `SampleBin` to defer to the `MlModelTracker`, since that is where we are now maintaining the state of that selection.

dongbohu:
@cgreene, @mhuyck and @rzelayafavila : Just noticed that bootstrap supports this kind of font spec:
```
<p class="h4">
  blah blah
</p>
```
So I changed `about` page from a simple `<p>` into `<p class="h4">`:
http://165.123.67.202/#/about
What do you think?

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -1,19 +1,31 @@
...
             var self = this;
+            self.isValidModel = false;
+            // Do nothing if mlmodel in URL is falsey. The error will be taken
+            // care of by "<ml-model-validator>" component.
+            if (!$stateParams.mlmodel) {
+              return;
+            }
+
+            self.modelInUrl = $stateParams.mlmodel;
+            self.selectedMlModel = MlModelTracker;
```
mhuyck: This is a cool design, by the way. I like that the service itself is an object with properties that yield the relevant MLModel details.

interface/src/app/mlmodel/validator.tpl.html
```diff
@@ -0,0 +1,5 @@
+<div class="alert alert-danger" role="alert" ng-if="$ctrl.errMessage">
+  {{$ctrl.errMessage}}
+  <br>
+  Please click <a href="#/home">Home</a> to reset it.
```
mhuyck: Can't change this now, but I see no reason to force the user to head back to the Home page to change the model. I'll file an issue to streamline this.

interface/src/index.html
```diff
@@ -98,6 +103,19 @@
...
+          <div class="col-sm-1"></div>
+          <div class="col-sm-11">
+            <span>Current Model: </span>
+            <span class="text-danger" ng-bind="modelInfo.title || 'Not Set'">
+            </span>
+            <span>(Click <a href="#/home">Home</a> to change)</span>
+          </div>
+        </div>
+        End of machine learning model info part
+        -->
```
mhuyck: This looks like debug code commented out. Did you mean to leave it in?

mhuyck:
I think the `class=` modifications can stay in for now, but we'll take them all out when handling issue #168. It is not good practice.

interface/src/app/mlmodel/validator.tpl.html
dongbohu: @mhuyck: I discussed this with @cgreene and we actually did this on purpose. Please feel free to file an issue. We can discuss further when I am back from vacation. Thanks.

interface/src/index.html
dongbohu: @mhuyck: You are right. I forgot to remove these comments. Since I have already merged the PR, I am going to file an issue to remove these lines later.

mhuyck:
Just to clarify (I was typing fast earlier and my comment didn't come through all the way): there is nothing wrong with using `class=` on an HTML tag to indicate a meaningful classification for the markup. The part that I'm objecting to is the use of `class="h4"` purely to change the way the element is styled. That's what CSS is for. If it should be an `h4`, the markup should use an `<h4>` tag to indicate that.

There is a separation of concerns for design logic that is no less important than separation of concerns in programming logic. Keeping it all straight pays dividends as a more understandable and maintainable system.

interface/src/app/mlmodel/validator.tpl.html
mhuyck: This works fine as an initial implementation -- you had a ton of work as it was -- and the overall workflow is much better. I don't want to lose sight of the fact that the flow could be even better.

Note that I prioritized the fix as "before multi-organism." Maybe it should say "before multi-MLModel." That's when I think it will really start to matter.

PR #162, opened 2017-06-12
Correcting description for the default ParticipationType that gets created in data migration file.

rzelayafavila:
Made changes based on comments in PR #160


rzelayafavila:
Ok, thanks @cgreene!

PR #174, opened 2017-06-20
Django genes: POST instead of GET

mhuyck:
Fix for Issue #148.

This is a WIP since the `django-genes` PR is not yet merged (so v0.13 referenced in requirements.txt does not yet exist). Once that is merged, this can be fully tested and reviewed.


mhuyck:
CircleCI is complaining (of course) because it can't find v0.13 of `django-genes`.

mhuyck:
Okay, `django-genes` is now updated, so this is no longer a WIP. Hmm... now, how to get CircleCI to retry.

interface/bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "adage",
-  "version": "0.7.0b",
+  "version": "0.8.0b",
```
rzelayafavila: :+1:

fabfile/adage_server.py
```diff
@@ -191,8 +191,11 @@ def import_data_and_index():
         CONFIG['data']['gene_history_file'])
     run('python manage.py import_gene_network  %s "Ensemble ADAGE 300"' %
         CONFIG['data']['gene_network_file'])
-    run('python manage.py import_node_gene_network %s "Ensemble ADAGE 300"' %
-        CONFIG['data']['node_gene_network_file'])
+    run('python manage.py create_or_update_participation_type'
+        ' "High weight genes" "Genes that most strongly influence the'
+        ' activation of a signature."')
```
rzelayafavila: Nice! I wonder if the second part of that, which is in the data migration file that automatically adds the ``High weight genes`` ParticipationType if no other PT object exists in the database, would be useful to have here. It reads "and we have found that they often reveal the underlying process or processes captured by the signature."" What do you think?

interface/src/app/gene/resource.js
```diff
@@ -1,22 +1,29 @@
...
+angular.module('adage.gene.resource', [
+  'ngResource',
+  'adage.utils'
+])
 
-.factory('Gene', ['$resource', function($resource) {
+.factory('Gene', ['$resource', 'ApiBasePath', function($resource, ApiBasePath) {
   return $resource(
-    '/api/v0/gene/:id', {
+    ApiBasePath + 'gene/:id', {
```
rzelayafavila: :+1:

interface/src/app/gene/resource.js
```diff
@@ -1,22 +1,29 @@
-angular.module('adage.gene.resource', ['ngResource'])
+angular.module('adage.gene.resource', [
+  'ngResource',
+  'adage.utils'
```
rzelayafavila: :+1:

interface/src/app/gene/resource.js
```diff
@@ -1,22 +1,29 @@
...
-.factory('Gene', ['$resource', function($resource) {
+.factory('Gene', ['$resource', 'ApiBasePath', function($resource, ApiBasePath) {
   return $resource(
-    '/api/v0/gene/:id', {
+    ApiBasePath + 'gene/:id', {
       id: '@id'
     }, {
-
+      post: {
+        url: ApiBasePath + 'gene/',
```
rzelayafavila: Hmm, I don't know if we need to define the url again here since we already defined it (and it should be the same as the one) in line 8. I think the ``/:id`` part should not matter if we do not send the ``id`` as a parameter to this ``post`` action.

interface/src/app/gene/network/network.js
```diff
@@ -372,7 +372,12 @@ angular.module('adage.gene.network', [
           // Collect a list of distinct genes for query.
           var geneList = getQueriedGenes(edgeList);
           // Now retrieve the Gene objects using the geneList.
-          Gene.get({'pk__in': geneList.join(), 'limit': 0},
+          Gene.post(
+            {},
+            $httpParamSerializerJQLike({
```
rzelayafavila: Nice! It's nice that there is already that other part of the code using $httpParamSerializerJQLike in this controller, so it is already being injected!

fabfile/adage_server.py
mhuyck: - [ ] add remainder of description for 'High weight genes' that I missed

interface/src/app/gene/resource.js
mhuyck: Good question. I think if the `:id` stays as it is in the GET then if it does get passed during a POST, the back-end will get tripped up since we're now hitting the "detail" REST endpoint rather than the "list" REST endpoint. This will mean the Tastypie logic we've added into `django-genes` to do the POST back to GET conversion won't get triggered and we'll get an error.

We aren't using the `:id` parameter during POST so I haven't tested that case explicitly, but I'm pretty sure it could foul things up.

interface/src/app/gene/network/network.js
mhuyck: Heh... yeah, I searched up to add it and found it was already there. Did a double-take. 😄

interface/src/app/gene/resource.js
mhuyck: I just saw the part of your comment that says "if we do not send the `id` as a parameter..."

You are absolutely right. I guess what I'm doing is protecting against future changes to the code that might go and add that parameter in. I'd rather have the Gene service keep working properly  under that condition rather than leave this issue to be discovered and fixed later.

mhuyck:
Rebased to incorporate #176.

mhuyck:
Okay, if CircleCI agrees, then I think this is ready to go. Let me know what you think, @rzelayafavila.

interface/src/app/gene/resource.js
rzelayafavila: Ok, that sounds like a good idea!

fabfile/adage_server.py
```diff
@@ -191,8 +191,13 @@ def import_data_and_index():
         CONFIG['data']['gene_history_file'])
     run('python manage.py import_gene_network  %s "Ensemble ADAGE 300"' %
         CONFIG['data']['gene_network_file'])
-    run('python manage.py import_node_gene_network %s "Ensemble ADAGE 300"' %
-        CONFIG['data']['node_gene_network_file'])
+    run('python manage.py create_or_update_participation_type'
+        ' "High weight genes" "High-weight genes are those that most strongly '
+        'influence the signature\'s activity, and we have found that they '
+        'often reveal the underlying process or processes captured by the '
+        'signature."')
```
rzelayafavila: 👍

Issue #145, opened 2017-05-07
Take absolute value of all activity values?


mhuyck:
Should we be taking the absolute value of all activity values before performing calculations? The original design kept negative activity values in the database while separating positive and negative node/signatures.

This question arises in the context of the volcano plot. Currently, the difference in activity for a node between two groups of samples is computed by taking the mean activity for each group and subtracting. The question is: should we take the absolute value first?

mhuyck:
Am I remembering right that @dhu created a separate data file for loading "absolute value" activities as a separate MLModel?

mhuyck:
Tagging @tj8901nm, who called this to my attention.

dongbohu:
Actually @tj8901nm created a new activity file in which all activity values are positive, but when deploying the database, we still used the original activity sheet. Should we import both activity files into the database (with different machine learning models)?

mhuyck:
Right, that is the question I'm asking. To put it simply: should we change the *code* to take absolute values or should we use a specially-prepared activity file? I thought the decision was to leave the code alone but handle this via the revised activity file.

If we proceed down the activity file route then we need to make another decision: keep both the original (signed) activity data *and* the new (unsigned) activity data, or do we remove the original data to avoid confusion?

cgreene:
I think we should use a specially prepared activity file. In the production instance we should probably keep only the new unsigned data. I think that having multiple models in our dev instance will be nice, because we expect multiple models in the not too distant future.

mhuyck:
This is tagged as intermediate priority, but I noticed that @tj8901nm's paper mentions (under *Active signature detection* in **Methods**) that "we ignore the signs of weight values."

Should that bump the priority, @cgreene? If I can get a copy of that activity file, I can load it when we deploy our updates today.

cgreene:
I don't have the activity file. Can you ask @tj8901nm? I agree with an importance bump. It would be ideal to get this loaded when updates are deployed.

mhuyck:
Message sent. If we don't hear back, it occurs to me that the new activity file would be just the same as the one we have just with all of the minus signs stripped out, so I could create it myself. Do you see any problem with doing it that way, @cgreene?

cgreene:
I do not see how that could be the wrong answer, but I am quite wary of changing the model file without some confirmation from @tj8901nm.

mhuyck:
Ok, fair enough.

mhuyck:
Fixed in #176

PR #175, opened 2017-06-20
Activity absval

mhuyck:
Changes to resolve Issue #145.


mhuyck:
Oops. Need to rebase first so this doesn't include the other PIR in progress...

PR #176, opened 2017-06-20
[FIX] Use the absolute value activity file by default.

mhuyck:
These changes resolve Issue #145.


adage/adage/config.py.template
```diff
@@ -69,8 +69,9 @@ DATA_CONFIG.update({
         DATA_CONFIG['data_dir'], 'PseudomonasAnnotation.tsv'),
     'activity_file': os.path.join(
         DATA_CONFIG['data_dir'],
-        'all-pseudomonas-gene-normalized_HWActivity_perGene_with_net300_100'
-        'models_k=300_seed=123_ClusterByweighted_avgweight_network_ADAGE.txt'),
+        'all-pseudomonas-gene-normalized_HWActivity_perGene_with_net300_'
+        '100models_1_100_k=300_seed=123_ClusterByweighted_avgweight_'
+        'network_ADAGE'),
```
rzelayafavila: Ok, this makes sense - does this filename need a ``.txt`` tacked on at the end since the filename below has that?

mhuyck: Sure does. Thanks! I missed that.

PR #181, opened 2017-06-29
Case insensitive search

### LINKED ISSUE #123: Can we make the search ignore capitalization

mhuyck:
This fixes #123, which is an issue with tokenization in the search engine that crops up with camelCase words. Also adds a test to ensure this problem does not come back and another test for the related issue whose fix caused this problem to arise.


mhuyck:
The only CodeClimate issue that looks semi-valid is the first one, but that's just picking up a non-ascii character in the newly-added test data, and that's covered by the `# coding: utf-8` line at the top of the file... something I'd expect `radon` to be able to handle.

CircleCI is going to need some help with Elasticsearch, I suspect, akin to what I did to make Codeship happy.

mhuyck:
The CircleCI fix is looking a little less straightforward than what I did for Codeship, mostly because building and testing takes place inside of Docker on CircleCI.

CircleCI *does* provide an Elasticsearch service that's disabled by default. I could enable that, but it's going to launch ES on the *host* VM, I'd expect. So then we'd need to tweak the haystack settings a bit more to contact ES on the host rather than assume ES is running at `127.0.0.1` (localhost) as we do now.

I suppose another option is to install and launch ES inside of the Docker image, which would be a better match for our current deployment configuration, but doesn't feel like heading in the right direction.

Going to stick with tweaking haystack settings first. It might be simpler that it looks.

mhuyck:
Sure enough. That was *way* simpler than it looked. Just enabled the Elasticsearch service on the CircleCI host. I keep forgetting that Docker is not a full-fledged VM, so the localhost IP `127.0.0.1` is *exactly* what we want for contacting the service... same as PostgreSQL.

adage/adage/settings.py
```diff
@@ -82,12 +82,18 @@
...
                         "lowercase",
                         "stop",
                         "snowball"
                     ]
                 },
+            },
+            'filter': {
+                'adage_word_delimiter': {
+                    'type': 'word_delimiter',
+                    'split_on_case_change': False
```
rzelayafavila: Nice! I think this makes sense. I was wondering though, any reason for using double quotes for strings above and single quotes here?

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,107 @@ def test_expressionvalue_big_post(self):
...
+
+@override_settings(HAYSTACK_CONNECTIONS=TEST_INDEX)
+class SearchIndexTestCase(ResourceTestCaseMixin, TestCase):
+    searchURI = '/api/v0/search/'
+    experiments = [
+        {
+            'accession':    'E-GEOD-31227',
+            'name':         'Expression data of Pseudomonas aeruginosa isolates'
+                            ' from Cystic Fibrosis patients in Denmark',
+            'description':  'CF patients suffer from chronic and recurrent '
```
rzelayafavila: Hmm, something about the spacing here feels funky. Is there a double space between `'description':` and `'CF patients...'`? Also is the PEP 8 linter cool with the indenting of the lines after this? I feel like it might want them to be aligned with either `'CF....'` or `'description'`. Minor things, but what do you think?

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,107 @@ def test_expressionvalue_big_post(self):
...
+            'accession':    'E-GEOD-17296',
+            'name':         'Transcription profiling of Pseudomonas aeruginosa '
+                            'roxSR and anr mutant strains under aerobic '
+                            'conditions',
+            'description':  'To assess the role of two redox-sensitive '
+                'transcriptional regulators, RoxSR and ANR, in Pseudomonas '
+                'aeruginosa under aerobic conditions, microarray analysis was '
+                'performed. Transcriptome profiles of roxSR mutant and anr '
+                'mutant aerobically grown in LB medium were determined by '
+                'Affymetrix GeneChip at both the exponential phase and early '
```
rzelayafavila: Same questions about spacing in these two other experiments.

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,107 @@ def test_expressionvalue_big_post(self):
...
+        self.assertValidJSONResponse(respLower)
+        jsonMixed = self.deserialize(respMixed)
+        jsonLower = self.deserialize(respLower)
+        self.assertEqual(
+            jsonMixed['meta']['total_count'],
+            jsonLower['meta']['total_count']
+        )
+        self.assertEqual(
+            jsonMixed['objects'],
+            jsonLower['objects']
```
rzelayafavila: :+1:

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,107 @@ def test_expressionvalue_big_post(self):
...
+        """
+        respMixed = self.api_client.get(self.searchURI, data={'q': 'mexR'})
+        self.assertValidJSONResponse(respMixed)
+        respLower = self.api_client.get(self.searchURI, data={'q': 'mexr'})
+        self.assertValidJSONResponse(respLower)
+        jsonMixed = self.deserialize(respMixed)
+        jsonLower = self.deserialize(respLower)
+        self.assertEqual(
+            jsonMixed['meta']['total_count'],
+            jsonLower['meta']['total_count']
```
rzelayafavila: :+1:

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,107 @@ def test_expressionvalue_big_post(self):
...
+        """
+        respPA14 = self.api_client.get(self.searchURI, data={'q': 'PA14'})
+        self.assertValidJSONResponse(respPA14)
+        self.assertEqual(
+            self.deserialize(respPA14)['objects'][0]['pk'],
+            'E-GEOD-24262'
+        )
+
+    def tearDown(self):
+        call_command('clear_index', interactive=False, verbosity=0)
```
rzelayafavila: Oh wow I hadn't realized it, but we probably need to call this after we create test indexes, huh?

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,107 @@ def test_expressionvalue_big_post(self):
...
+                '(PAO6261) strains were cultivated aerobically in LB in '
+                'Erlenmeyer flasks, and total RNAs were extracted at both the '
+                'exponential phase (OD600 = 0.3) and early stationary phase '
+                '(OD600 = 1.4). The experiment was performed in duplicate '
+                'independent cultures.'
+        }
+    ]
+
+    def setUp(self):
+        haystack.connections.reload('default')
```
rzelayafavila: I forget - why do we need this line again?

adage/adage/settings.py
mhuyck: Ahh... nice catch. I saw that when I started editing the page and meant to correct it. This file hasn't been touched in a while, apparently, and was "grandfathered" in before we mandated PEP8 compliance. Will fix that.

- [ ] fix string delimiter inconsistency in settings.py

adage/analyze/tests.py
mhuyck: Also a good catch. Not following PEP8, I think. I was making it match the sample experiment I copied from the original `Experiment` tests. Will fix.

- [ ] bring test Experiment data into PEP8 compliance

adage/analyze/tests.py
mhuyck: Uh, well, I got this line from the Tribe tests. I was assuming this reload was needed to be sure we use the Haystack connection we've changed in the `@override_settings` decorator above. I didn't try without it to see if it was actually necessary. Should I?

adage/analyze/tests.py
mhuyck: This is another one I got from the Tribe code you pointed out for me... unless there's somewhere else you need it, I think you're covered! 😄

adage/analyze/tests.py
mhuyck: While this style was not strictly against PEP8 rules, it does fall squarely into the "pet peeve" category here (see last example): https://www.python.org/dev/peps/pep-0008/#pet-peeves .

Fixed.

adage/analyze/tests.py
mhuyck: Please let me know if you think this line should change, but I'm guessing we need it.

I did try the test without that line and things worked, but I also did a Google search on `haystack.connections.reload` and found a *lot* of examples of people using that line for testing Haystack. One of the hits was the actual source code for Haystack and it seems that what this `.reload` is doing is purging the relevant entry from a connection pool so the next use will presumably reload it.

So although it appears to work in our case without that line, it appears to be there for certain conditions that we might eventually hit in our testing.

adage/adage/settings.py
```diff
@@ -72,22 +72,28 @@
...
+            'analyzer': {
+                'adage_snowball': {
+                    'type': 'custom',
+                    'tokenizer': 'standard',
+                    'filter': [
+                        'standard',
+                        'adage_word_delimiter',
+                        'lowercase',
+                        'stop',
+                        'snowball'
```
rzelayafavila: :+1: for single quotes!

adage/analyze/tests.py
```diff
@@ -1025,3 +1034,106 @@ def test_expressionvalue_big_post(self):
...
+            'accession': 'E-GEOD-24262',
+            'name': 'PA14_mexR vs. wildtype planktonic cells in minimal '
+                'medium with C-30',
+            'description': 'Mutations that made the cells insensitive to the '
+                'QS inhibition by C-30 were identified in mexR, a multi-drug '
+                'resistance operon repressor. Gene expression of mexR mutant '
+                'relative to wild-type at the presence of C-30 was examined. '
+                'Strains: PA14_mexR and wildtype. Medium: OS minimal medium + '
+                '0.1% Adenosine as carbon source. Compounds: 50 µM C-30 added '
+                'at OD600=0.25. Time: 2 hr. Temp: 37 ºC. Cell type: Planktonic '
```
rzelayafavila: Hmm, Code Climate complains that this line is one character too long, which the linter on my text editor agrees with. Shall we move the trailing space to the next line?

adage/analyze/tests.py
rzelayafavila: Haha ok, nice!

adage/analyze/tests.py
rzelayafavila: Yes, sure, let me dig in a little bit more into this today, I think you are right about purging the relevant entry. I will let you know by today if I think it is totally irrelevant and we should dispose of it though.

adage/analyze/tests.py
mhuyck: Missed that, thanks. Looks like my editor stopped warning me about these... will have to check into that.

- [ ] correct line >80 chars
- [ ] fix linter settings on my editor

adage/analyze/tests.py
rzelayafavila: Ah ok -  seems from my research that we almost definitely want this here (as well as calling the 'clear_index' command below), so that stuff doesn't get persisted between tests. https://stackoverflow.com/questions/23569583/haystack-with-elasticsearch-unit-test-gets-different-results-when-run-in-isolat , http://bwreilly.github.io/blog/2013/07/21/testing-search-haystack-in-django/

mhuyck: Right... those two links came up in my search, too. It would be nice to have an authoritative solution in the form of Haystack documentation, but looking over the source code clinched it for me.

rzelayafavila: Yeah - like you said, I wish the documentation itself provided some guidance on that!

PR #182, opened 2017-06-29
Changing Signature page layout, switching to expandable sections

rzelayafavila:
Removing the grid-like structure of the Signature page, and instead putting the three existing directives into consecutive sections, and adding anchors to them. Also adding links that take the users to these anchors using Angular's '$location.hash()' and '$anchorScroll'. This PR addresses issue #131 

The Signature page now looks like this: http://165.123.67.154:8000/#/signature/404
Let me know how it looks!

### DIFF
--- a/interface/src/app/signature/enriched_genesets.tpl.html
+++ b/interface/src/app/signature/enriched_genesets.tpl.html
@@ -1,12 +1,13 @@
 <h4>
-  <b>Enriched gene sets: </b>&nbsp;
-  <button type="button" class="btn btn-default"
-    ng-click="goToAnchor({anchor: 'top-of-page'})">
-    Back to Top
-  </button>
+  <b>Enriched gene sets: </b>
 </h4>
+<p>
+  Explore canonical (or curated) gene sets that significantly overlap
+  with this signature's associated genes; can help with interpretation of
+  the underlying process or processes.
+</p>
 
-<table id="enrichment-table" class="table table-striped table-bordered table-condensed">
+<table id="enrichment-table" class="table table-striped">
   <thead>
     <tr>
       <th>Pathways/Processes/Diseases</th>
@@ -16,11 +17,28 @@
     </tr>
   </thead>
 
-  <tr ng-repeat="geneset in enrichedGenesets">
+  <tr ng-repeat="geneset in enrichedGenesets | limitTo: numGenesetsShown">
     <td><a href="{{geneset.url}}">{{geneset.name}}</a></td>
     <td>{{geneset.dbase}}</td>
     <td>{{geneset.pValue}}</td>
     <td>{{geneset.genes}}</td>
+  </tr>
+
+  <tr ng-click="setMode()" ng-show="topMode">
+    <td>
+      <button type="button" class="btn btn-default">Show all ...</button>
+    </td>
+    <td></td>
+    <td></td>
+    <td></td>
+  </tr>
+  <tr ng-click="setMode()" ng-hide="topMode">
+    <td>
+      <button type="button" class="btn btn-default">Only show top {{topNum}}</button>
+    </td>
+    <td></td>
+    <td></td>
+    <td></td>
   </tr>
 
 </table>
--- a/interface/src/app/signature/enriched_genesets.tpl.html
+++ b/interface/src/app/signature/enriched_genesets.tpl.html
@@ -1,5 +1,9 @@
 <h4>
-  <b>Enriched gene sets: </b>
+  <b>Enriched gene sets: </b>&nbsp;
+  <button type="button" class="btn btn-default"
+    ng-click="goToAnchor({anchor: 'top-of-page'})">
+    Back to Top
+  </button>
 </h4>
 
 <table id="enrichment-table" class="table table-striped table-bordered table-condensed">
--- a/interface/src/app/signature/high_range_exp.tpl.html
+++ b/interface/src/app/signature/high_range_exp.tpl.html
@@ -1,6 +1,10 @@
 <div ng-if="!queryStatus">
   <h4>
-    <b>Experiments with the Highest Range: </b>
+    <b>Experiments with the Highest Range: </b>&nbsp;
+    <button type="button" class="btn btn-default"
+      ng-click="goToAnchor({anchor: 'top-of-page'})">
+      Back to Top
+    </button>
   </h4>
   <button class="btn btn-primary" ng-click="setMode()"
           ng-hide="topMode">Show Top {{topNum}}</button>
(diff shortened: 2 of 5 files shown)


dongbohu:
It is a good starting point. Maybe you can also add a "Back to Top" button or link at the right side of each section's title (except the first section)?

I also wonder whether the UI will look better if the font of "sections: aaa bbb ccc" part could be larger and the two lines merge into one.

cgreene:
There's not really a clear hierarchy on the page, and the "sections" header is very small relative to other headers + the text is not aligned with the "participation type bit".

What did you decide about the floating nav bar that we were looking at on the bootstrap page? That seems like it might be a cleaner design than this approach. The overall number of links ends up being the same.

cgreene:
Can you sketch this out on a whiteboard, snap a picture, and add it to the issue discussion before you put time into making it? It may be more efficient to iron out things.

rzelayafavila:
@cgreene - I thought we decided to start with this approach and play with it first. Also, by sketching 'this' on a whiteboard do you mean this approach or the floating nav bar like the one on the bootstrap page?

cgreene:
I am under the impression that we agreed to split the sections out instead of going side-by-side as in the previous design. I don't think we need to revert that.

We're at the stage on this server where we need to consider how the design works for users. I'm not sure that this design puts our best foot forward either.

If you wanted to sketch out an example signature page on a whiteboard with a note as to what the hierarchy of importance is for each component/etc, then I think that will help us design a solution that meets the needs of our users and addresses the underlying issue (users didn't realize there was more at the bottom).

rzelayafavila:
Ok @dongbohu - I played around with it a little bit more and added the "Back to the Top" buttons. I am working on the sketch of what the overall structure of the page and its hierarchy should look like though, and I will post that on the issue related to this PR.

rzelayafavila:
Ok @dongbohu @mhuyck @cgreene: I implemented the solution I posted about in issue #131. You can play with it here: http://165.123.67.154:8000/#/signature/404 . Let me know what you think!

cgreene:
I like this a lot better! There's something funky with the styling on the tables (they all look subtly different), but this does a nice job of putting the options at the front of the fold.

dongbohu:
In the second section (Enriched top genesets), the "only show top 3" line at the end probably shouldn't be in the table.

In the third section (Experiments with the Highest Range), there seems to be an extra blank line between the title and the "Click the activity ..." line.

rzelayafavila:
Ok @casey and @dongbohu, I think I got those things, let me know if I did and/or what else needs fixing!

rzelayafavila:
I am still playing around with styles and putting in/removing stuff, but let me know what other feedback you have!

cgreene:
I can confirm tables now look much more consistent. L&F is :+1: to me! :tada:

mhuyck:
I agree this is a big step in the right direction. My obsessive-compulsive CSS tweaking impulse is tingling a bit, but I think we should run this by our users to get their take before doing too much fine-tuning.

mhuyck:
I started looking at the code to see about finishing off this review, but it looks as if the latest demo code isn't pushed to your `changing-signature-page-layout`. Is that right, @rzelayafavila?

rzelayafavila:
Hey @mhuyck - yes, you were right, I just pushed the changes, and the live server is still up for exploring this UI. Let me know how it looks (both the UI and the code changes), and I can make more changes later today and tomorrow to wrap this up this weekend. Thanks!

interface/src/app/signature/high_range_exp.tpl.html
```diff
@@ -2,12 +2,10 @@
...
-          ng-hide="topMode">Show Top {{topNum}}</button>
-  <button class="btn btn-primary" ng-click="setMode()"
-          ng-show="topMode">Show All</button>
-
-  <br>
+  <p>
+    Explore gene expression experiments with the highest range of this
+    signature's activity.
+  </p>
   (Click the activity distribution widget to see sample annotations in this experiement.)
```
kurtwheeler: Shouldn't this parenthetical text be wrapped in a `<p>` tag?

Also I noticed Casey's comment about the tables not quite looking consistent. This table doesn't need a header column since it only has one column, but it might help the look a bit to have a line break here so the text isn't almost on top of the table.

kurtwheeler:
@rzelayafavila I looked this over in the absence of @dongbohu and @mhuyck and it looks good to me. I left one super minor comment which may or may not actually be a good idea. Your call.

interface/src/app/signature/high_range_exp.tpl.html
cgreene: I'm not sure this needs to be a parenthetical statement. It could just be the second sentence in the `<p>` above.

rzelayafavila: Gotcha, I'll change it, thank you both!

rzelayafavila: Ok, I changed it, and I agree - it looks better!

mhuyck:
I usually blow off the "similar code" found by Code Climate because it's just picking up on Angular boilerplate that we can't really simplify, but in this case I think it might be on to something. The three instances it's pointing out are a block of code for each of the new "expandable table" sections we've added.

It might behoove us to generalize the code for this into an actual "expandable table" component that we could reuse anywhere. This would require a bit of a refactoring and some thought about how to go about it in a clean way, but possibly worth following up. Out of scope for this PR, I'd say, but if you agree maybe we should file an issue on it so we can take a second look.

interface/src/app/signature/enriched_genesets.tpl.html
```diff
@@ -12,13 +17,30 @@
     </tr>
   </thead>
 
-  <tr ng-repeat="geneset in enrichedGenesets">
+  <tr ng-repeat="geneset in enrichedGenesets | limitTo: numGenesetsShown">
```
mhuyck: 👍 I always forget about Angular's awesome filters. This is a great way to implement this feature.

rzelayafavila:
Ok good call @mhuyck - I was thinking about it as well, and I agree that it is outside the scope of this PR, since it is kind of unclear what the best approach would be (I was considering the expanding table components as well!). I will make an issue for this.

interface/src/app/signature/enriched_genesets.tpl.html
rzelayafavila: Totally, thanks! They do help implementing stuff like this and making it concise and neat.

rzelayafavila:
Ok thanks everyone who helped with this PR! Merging.

Issue #134, opened 2017-04-28
gene query - is it getting only edges to the query gene or is it getting edges b/w any gene in the queries


cgreene:

dongbohu:
It is getting all edges that are related to any of the genes specified in the URL (in other words, a UNION Of all edges related).

cgreene:
Ok - Deb says that what she really wants it to return is:
* Edges to any genes specified in the URL.
* Edges between any genes that appear in the network.

This will let her see which things are small modules of connected genes. I expect this to be the case for most of our users.

dongbohu:
Got it.

dongbohu:
Fixed in PR #194.

Issue #184, opened 2017-06-29
UI should indicate the gene-gene network's cutoff value


dongbohu:
Currently the gene-gene network file that has been loaded into the database has a cutoff value of 0.4, but the slider of correlation values on the web UI has a valid range of [0, 1]. This is misleading. We should set the slider's minimum value at the cutoff value.

This issue is related to #183.

dongbohu:
I am trying to figure out the best way to solve this issue. Obviously we can hard code the cutoff value of 0.4 for now, but it won't work if a user loads another gene-gene network file with different cutoff value. 

How about adding a new field `gene_gene_network_cutoff` in `MLModel` table? The management command that loads the gene-gene network will need another argument which allows users to specify this cutoff value when creating the MLModel record.

mhuyck:
This sounds like a great plan. When you update the management command, please also update the `fabfile/adage_server.py` script that calls it during deployment so that step does not break.

dongbohu:
Fixed in PR #201.

PR #194, opened 2017-07-05
Fix of Issue #134: Include edges between related genes in gene-gene network

dongbohu:
This PR also includes a few minor corrections of typos or long line cutoff to make PEP8 happier.
Please see Issue #134 for more details.

### DIFF
--- a/adage/analyze/api.py
+++ b/adage/analyze/api.py
@@ -427,7 +427,14 @@ class EdgeResource(ModelResource):
                 raise BadRequest("Invalid gene IDs: %s" % genes)
             # "in" operator in Django supports both list and set.
             qset = Q(gene1__in=ids) | Q(gene2__in=ids)
-            object_list = object_list.filter(qset).distinct()
+            direct_edges = Edge.objects.filter(qset).distinct()
+            related_genes = set()
+            for e in direct_edges:
+                related_genes.add(e.gene1)
+                related_genes.add(e.gene2)
+            object_list = object_list.filter(
+                gene1__in=related_genes, gene2__in=related_genes
+            ).distinct()
         return object_list
 
 
--- a/adage/analyze/tests.py
+++ b/adage/analyze/tests.py
@@ -475,7 +475,8 @@ class APIResourceTestCase(ResourceTestCaseMixin, TestCase):
 
     def call_non_get_post_API(self, uri, data={}):
         '''
-        Helper method: assert that PUT, PATCH and DELETE methods are NOT allowed
+        Helper method: assert that PUT, PATCH and DELETE methods are NOT
+        allowed.
         '''
         # PUT
         resp = self.api_client.put(uri, data=data)
@@ -698,6 +699,19 @@ class APIResourceTestCase(ResourceTestCaseMixin, TestCase):
         self.call_get_API(uri)
         self.call_non_get_API(uri)
 
+    @staticmethod
+    def get_edge_union(gene_ids):
+        """Return all edges that are related to input gene_ids."""
+        qset = Q(gene1__in=gene_ids) | Q(gene2__in=gene_ids)
+        direct_edges = Edge.objects.filter(qset).distinct()
+        related_genes = set()
+        for e in direct_edges:
+            related_genes.add(e.gene1)
+            related_genes.add(e.gene2)
+        return Edge.objects.filter(
+            gene1__in=related_genes, gene2__in=related_genes
+        ).distinct()
+
     def test_edge_union(self):
         """
         Test edge unions via 'api/v0/edge/?genes=<id1,id2,...>' API.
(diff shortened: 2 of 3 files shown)


dongbohu:
Here is a demo of this PR:
http://165.123.67.202/#/gene_network/?genes=6546&mlmodel=4
Note that now there are edges between the genes that are directly related to gene 6546.

mhuyck:
For comparison, see the same gene in the currently-deployed site: http://adage.greenelab.com/#/gene_network/?genes=844&mlmodel=1

Nice fix, @dongbohu! Looking over the code now.

adage/analyze/api.py
```diff
@@ -427,7 +427,14 @@ def apply_filters(self, request, applicable_filters):
...
             qset = Q(gene1__in=ids) | Q(gene2__in=ids)
-            object_list = object_list.filter(qset).distinct()
+            direct_edges = Edge.objects.filter(qset).distinct()
+            related_genes = set()
+            for e in direct_edges:
+                related_genes.add(e.gene1)
+                related_genes.add(e.gene2)
+            object_list = object_list.filter(
+                gene1__in=related_genes, gene2__in=related_genes
+            ).distinct()
```
mhuyck: Let me see if I follow: before, we applied `qset` as a filter to remove `Edge` objects that did not have either `gene1` or `gene2`. The issue is raised because this is too restrictive: we are *also* filtering out edges that are present in the network between genes that share an edge with our search set in `genes` because those are not direct edges.

So, now, we start by creating a list of `direct_edges`, expand out to get all of the genes that participate in those edges (`related_genes`) and *then* we go back and filter our overall `Edge` list to include only those edges touching one of those `related_genes`.

This seems like a logical approach, but I don't think it matches what was requested. Why, for example, are we seeing many more *genes* than the original query, when we were just looking for more *edges*?

I think I can answer. The new `object_list.filter` applied at lines 435-437 is picking out all edges that touch one of the `related_genes`. Since we have already expanded our gene list to get `related_genes`, what we actually want here is to pick out *only* edges where both genes are in our list of `related_genes`. I think you could express this via a new `qset` similar to line 429:

```python
qset = Q(gene1__in=related_genes) & Q(gene2__in=related_genes)
```

and then we could use this new `qset` in place of the filter at lines 435-437:

```python
object_list = object_list.filter(qset).distinct()
```
See what I am saying? I'm taking a small leap in the syntax by assuming that the `&` operator will do what I expect because I haven't fully read up on `django.db.models.Q`, but do you follow my logic?

interface/src/app/gene/network/network.js
```diff
@@ -206,7 +206,7 @@ angular.module('adage.gene.network', [
         });
       };
 
-      // Function that sets edges in the netowrk.
+      // Function that sets edges in the network.
```

adage/analyze/tests.py
```diff
@@ -698,6 +699,19 @@ def test_one_edge(self):
...
+        """Return all edges that are related to input gene_ids."""
+        qset = Q(gene1__in=gene_ids) | Q(gene2__in=gene_ids)
+        direct_edges = Edge.objects.filter(qset).distinct()
+        related_genes = set()
+        for e in direct_edges:
+            related_genes.add(e.gene1)
+            related_genes.add(e.gene2)
+        return Edge.objects.filter(
+            gene1__in=related_genes, gene2__in=related_genes
+        ).distinct()
```
mhuyck: This follows the same logic implemented above, so if changed above it will also need to change here.

adage/analyze/tests.py
```diff
@@ -709,18 +723,16 @@ def test_edge_union(self):
         id2 = Gene.objects.last().id   # The last gene
 
         # The number of edges in the union of the first and last genes
-        # should be: num_gene1 + num_gene2 - 1, "-1" is due to double
-        # counting of the edge between the first and last genes.
+        # should be equal to num_gene1 * num_gene2.
```
mhuyck: This works out, but only because the test data set is a fully-connected network of 100 genes. What I think we're looking for is more subtle, as described above.

adage/analyze/tests.py
```diff
@@ -938,11 +948,6 @@ def create_expressionvalue_data(self, num_genes):
         else:
             organism = factory.create(Organism)
 
-        if MLModel.objects.exists():
-            ml_model = MLModel.objects.first()
-        else:
-            ml_model = factory.create(MLModel)
-
```
mhuyck: Nice catch.

adage/analyze/api.py
dongbohu: @mhuyck: Your proposal is actually equivalent to my current implementation. When there are multiple conditions inside `filter`, these conditions are AND-ed together. (What you suggested was my original version, but I changed it to the current version because I feel that using `filter` will be easier to understand.

mhuyck: I don't think so. My suggested implementation may not be correct, but please note the observation that we have more *genes* coming back in this new demo site. I am pretty sure the request was that we find all of the relevant *edges* without expanding our list of genes.

dongbohu: True. In http://adage.greenelab.com/#/gene_network/?genes=844&mlmodel=1
when the minimum correlation is set at 0.5, 8 edges are shown between the queried gene in the URL (dnaA) and 8 other genes.

In http://165.123.67.202/#/gene_network/?genes=6546&mlmodel=4
when the minimum correlation is set at 0.5, you can see the same 8 edges that originate from dnaA, but there are other edges whose correlations are no less than 0.5 and are between two genes that are directed related to dnaA. Are you saying that the network should only show the edges that are among those 8 genes? What do you think @cgreene?

cgreene: Let me see if I understand. The user request was that we add edges between the genes that were returned - not just the query genes. There are two ways to interpret this:

1. Query the gene (dnaA) -> This provides a list of related genes (up to X) -> query for all edges between any genes in X + dnaA -> Display any edges (+associated nodes) over the slider cutoff.

2. Query the gene (dnaA) -> This provides a list of related genes (up to X) -> query for all edges between any genes in X + dnaA -> Display any edges between genes that have an edge that is over the slider-cutoff with dnaA.

Did I get that framing correct?

dongbohu: @cgreene: yes, that is right.

mhuyck: Right. I guess (1) is the way it is implemented now, while (2) is the way I understood the request.

cgreene: quick slack conversation with @mhuyck led to that looks like we asked the user in question for clarification and 2 is consistent with what we're pretty sure she was asking for.

cgreene: follow-up to more slack conversations: We have a call today with the user. Will just ask her directly. :smile:

dongbohu: Upon Deb's request, the left slider of correlation is set at 0 for now. When issue #184 is fixed, it will be set at the cutoff value of the user-selected machine learning model.

PR #195, opened 2017-07-05
Changing header of ParticipationType selector in signature page to read 'Filter genes in this signature by:'.

rzelayafavila:
This makes the meaning of it more obvious to new users. Example of this new UI is running at: http://165.123.67.154:8000/#/signature/100

### DIFF
--- a/interface/src/app/signature/signature.tpl.html
+++ b/interface/src/app/signature/signature.tpl.html
@@ -6,7 +6,9 @@
 <h3 class="text-warning" ng-bind="ctrl.statusMessage"></h3>
 
 <p>
-  <b>Filter genes in this signature by:</b>
+  Each signature is comprised of genes with varying importance. Different
+  methods can be used to find the most relevant set of genes.
+  <b>Filter genes in this signature using:</b>
   <br>
   <participation-type-selector
     selected-participation-type="ctrl.selectedParticipationType">
--- a/interface/src/app/signature/signature.tpl.html
+++ b/interface/src/app/signature/signature.tpl.html
@@ -6,7 +6,7 @@
 <h3 class="text-warning" ng-bind="ctrl.statusMessage"></h3>
 
 <p>
-  <b>Participation Type of Genes in this Signature:</b>
+  <b>Filter genes in this signature by:</b>
   <br>
   <participation-type-selector
     selected-participation-type="ctrl.selectedParticipationType">


rzelayafavila:
Ok, thanks @dongbohu!

cgreene:
I feel like this missed something @gwaygenomics said though.

cgreene:
What was it? Something like:

Each signature is comprised of genes with various weights. Different methods find the most relevant set of genes. Filter genes in this signature by:

There was some additional value in being overly explanatory in this case, I think. Our users are not as familiar with these methods as we are.

gwaygenomics:
> There was some additional value in being overly explanatory in this case, I think. Our users are not as familiar with these methods as we are.

💯

> Each signature is comprised of genes with various weights. Different methods find the most relevant set of genes. Filter genes in this signature by:

After chatting with @rzelayafavila it sounded like he was going to try adding some similar explanatory language. Although I do think `weights` can also be a bit confusing. How about

"Each signature is comprised of genes with varying importance. Different methods find the most relevant set of genes."  Then, in bold: **"Filter genes in this signature by:"**

rzelayafavila:
Yep, I totally agree that there is value in adding that additional explanation to make it as clear as we can to new users.

rzelayafavila:
What do you think about using "varying importance" vs "varying weight", @cgreene?

cgreene:
importance > weight. More meaningful word 😄

cgreene:
`>` used as greater than - not arrow 👍

rzelayafavila:
Ok, cool, I will change this!

rzelayafavila:
Ok, I've changed it: http://165.123.67.154:8000/#/signature/400 Let me know if that (and the styling) looks good!

cgreene:
Feels like maybe having two lines disconnects "filter" from the previous sentence that describes filtering. What about merging those two paragraphs together?

rzelayafavila:
@cgreene I was wondering about that, and couldn't tell which worked better. I'll change it so that they are both in the same paragraph.

rzelayafavila:
Done.

cgreene:
Looks like you may need to edit to remove a linebreak too. Edited in chrome to see what it looked like. Thing together may help to unify that sentence + filter text.
screen shot 2017-07-06 at 11 05 56 am
screen shot 2017-07-06 at 11 06 19 am

Final wording request: `Filter... by` -> `Filter... using`?

cgreene:
"Different methods find" -> "Different methods can be used to find..."

Ok - actual final wording request 😉

cgreene:
With those changes I am 👍 on both interface & code.

rzelayafavila:
@cgreene Ok, I changed it, but now since the bold part is at the end of the sentence, it doesn't appear over the selector when viewing the page in a full window. Do you think this is ok?

cgreene:
I think that bringing it together with the sentence before was more important. Merge at will 👍

rzelayafavila:
Ok!

Issue #173, opened 2017-06-16
"List of genes functionality" was not noticed by a collaborator


cgreene:
In the gene network gene search, this was missed. Maybe we can provide a more effective signifier for this.

rzelayafavila:
Resolved with merging of PR #196

Issue #178, opened 2017-06-27
Adding samples to "shopping cart" is non-obvious


jaclyn-taroni:
Unclear how to move from Analyze search step to adding samples to visualization

mhuyck:
True. Thank you for capturing this. I'm going to tag issue #142 as well. That's an over-arching issue where I'm trying to organize concerns about analysis flow.

mhuyck:
Closed by #200. Further work in this area will continue under #142.

Issue #142, opened 2017-05-05
Simplify user flow for analyzing samples


mhuyck:
This issue is a placeholder for several smaller changes we need to make to streamline the steps a user must take to perform a typical analysis.

@tj8901nm has captured the current flow via a PDF posted to Slack. I propose to start by reviewing this and sketching a new flow that makes more sense to @tj8901nm and anyone from Deb's lab I can meet with on Monday.

tj8901nm:
[Figure 3.pdf](https://github.com/greenelab/adage-server/files/979810/Figure.3.pdf)
I just updated the flowchart. Red circles and arrows indicate how I imagine the analysis workflow. Currently, step 6 is missing. Also, step 6 and 7 are both on the volcano plot page. This might require users to go back once. One solution could be opening the node page in a new window. Other solutions are welcome.

mhuyck:
Nice. Thank you, @tj8901nm!

mhuyck:
One thing that Deb commented on as we worked to reproduce issue #146 is that it's not natural to remember to click on the SampleBin icon after adding an experiment. Her usual pattern is to find an experiment and begin defining sample groups right away. (So the "shopping cart" metaphor for collecting samples is not helping in the way I would have expected.)

dongbohu:
Being addressed by PR #161,

jaclyn-taroni:
Sketched out some thoughts about how this workflow might go. I agree with Deb that it's natural to want to define groups before you get to the "Analysis Detail" page.


![adage-server-142-thoughts](https://user-images.githubusercontent.com/19534205/28240335-ea57a140-694d-11e7-94ea-271339a6add5.png)

I'm far less sure about the second image. My concern there is that I completely missed the button in the upper righthand corner for going to the "Analysis Detail" page (part of the motivation for filing #178)

mhuyck:
We have made some minor improvements to this process with PR #200. Addressing the clumsiness in this workflow in a more substantial way will require refactoring the code behind these views to be more modular. I am keeping this issue open, but moving the milestone to August, when that refactoring work will be completed.

mhuyck:
Another bit of feedback about assigning samples to groups from the Hogan lab (paraphrasing): "it would be easier to assign samples to groups if we could *select* multiple samples [within the sample bin view] and then assign everything from the selection to a group all at once."

[...]

PR #196, opened 2017-07-11
Change the UI design of gene search to highlight most useful features, including switching to gene list search

rzelayafavila:
This PR solves issue #173, and should also address issue #197.  A sample of what this new UI looks like is running at http://165.123.67.154:8000/#/gene_search?mlmodel=1 if you would like to play around with it and let me know what you think. Thanks!

### DIFF
--- a/adage/requirements.txt
+++ b/adage/requirements.txt
@@ -12,7 +12,7 @@ six==1.10.0
 unicodecsv==0.14.1
 urllib3==1.21.1
 wsgiref==0.1.2
-django-genes>=0.13
+django-genes>=0.14
 tribe-client>=1.1.8
 defusedxml==0.5.0
 lxml==3.7.3
--- a/interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
+++ b/interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
@@ -1,30 +1,37 @@
 <form>
 
   <div class="control-group">
-    <div class="well">
-      <span ng-repeat="gene in selectedGenes">
-        <selected-gene-button gene="gene"></selected-gene-button>
-      </span>
+    <span ng-repeat="gene in selectedGenes">
+      <selected-gene-button gene="gene"></selected-gene-button>
+    </span>
 
-      <input type="text" ng-model="autocompleteQuery"
-        placeholder="Enter gene name"
-        uib-typeahead="result.standard_name for result in searchGenes($viewValue)"
-        typeahead-min-length="2"
-        typeahead-on-select="onSelect($item,$model,$label)"
-        typeahead-loading="loadingGenes">
+    <input type="text" ng-model="autocompleteQuery"
+      placeholder="Search gene name..."
+      uib-typeahead="result.standard_name for result in searchGenes($viewValue)"
+      typeahead-min-length="2"
+      typeahead-on-select="onSelect($item,$model,$label)"
+      typeahead-loading="loadingGenes">
 
-      <i ng-show="loadingGenes" class="fa fa-spinner"></i>
-      <div ng-show="noResults">
-        <i class="fa fa-remove"></i> No genes found for query
-      </div>
+    <i ng-show="loadingGenes" class="fa fa-spinner"></i>
+
+    <div ng-show="noResults">
+      <i class="fa fa-remove"></i> No genes found for query
     </div>
   </div>
 
   <p class="bg-danger">{{ errors }}</p>
 
+  <div class="control-group">
+    <button type="button" class="btn btn-sm" ng-click="switchToMany()">
+      I would like to copy and paste a list of genes instead.
+    </button>
+  </div>
+
+  <br>
   <div class="form-group">
-    <button class="btn btn-success" ng-click="sendToNetwork()">
+    <button class="btn btn-success btn-lg" ng-click="sendToNetwork()">
       View Genes as Network <i class="fa fa-check"></i>
     </button>
   </div>
+
 </form>
--- a/interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
+++ b/interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
@@ -26,7 +26,7 @@
 
   <div class="control-group">
     <button type="button" class="btn btn-sm" ng-click="switchToMany()">
-      I would like to copy and paste a list of genes instead.
+      I would like to paste a list of genes
     </button>
   </div>
 
--- a/interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
+++ b/interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
@@ -4,7 +4,10 @@
     <span ng-repeat="gene in selectedGenes">
       <selected-gene-button gene="gene"></selected-gene-button>
     </span>
+  </div>
+  <br>
 
+  <div>
     <input type="text" ng-model="autocompleteQuery"
       placeholder="Search gene name..."
       uib-typeahead="result.standard_name for result in searchGenes($viewValue)"
--- a/interface/src/app/gene/searchFew/autocomplete-search-panel.tpl.html
+++ b/interface/src/app/gene/searchFew/autocomplete-search-panel.tpl.html
@@ -1,16 +1,12 @@
 <div class="panel panel-default">
   <div class="panel-heading">
-    <h3>
-      Search for Genes
-    </h3>
+    <b>Type in gene name to search using autocomplete</b>
   </div>
   <div class="panel-body">
-    <autocomplete-search-form organism="{{organism}}" ml-model="{{mlModel}}">
+    <autocomplete-search-form
+      organism="{{organism}}"
+      ml-model="{{mlModel}}"
+      switch-to-many="switchToMany()">
     </autocomplete-search-form>
-  </div>
-  <div class="panel-footer">
-    <button type="button" class="btn btn-link" ng-click="switchToMany()">
-      I would like to copy and paste a list of genes instead.
-    </button>
   </div>
 </div>
--- a/interface/src/app/gene/searchFew/searchFew.js
+++ b/interface/src/app/gene/searchFew/searchFew.js
@@ -77,7 +77,8 @@ angular.module('adage.gene.searchFew', [
     ],
     scope: {
       organism: '@',
-      mlModel: '@'
+      mlModel: '@',
+      switchToMany: '&'
     },
     restrict: 'E',
     templateUrl: 'gene/searchFew/autocomplete-search-form.tpl.html'
--- a/interface/src/app/gene/searchFew/searchFew.js
+++ b/interface/src/app/gene/searchFew/searchFew.js
@@ -36,7 +36,6 @@ angular.module('adage.gene.searchFew', [
           $scope.loadingGenes = true;
           var qparams = {
             'query': val,
-            'limit': numResultsToReturn,
             'organism': $scope.organism
           };
 
@@ -63,7 +62,7 @@ angular.module('adage.gene.searchFew', [
               var geneResultsList = response.results.filter(function(result) {
                 return selectedIDs.indexOf(result.id.toString()) === -1;
               });
-              return geneResultsList;
+              return geneResultsList.slice(0, numResultsToReturn);
             }
           );
         };
--- a/interface/src/app/gene/searchFew/searchFew.js
+++ b/interface/src/app/gene/searchFew/searchFew.js
@@ -51,17 +51,21 @@ angular.module('adage.gene.searchFew', [
               } else {
                 $scope.noResults = false;
               }
-
-              return data.results;
             },
 
             // Error function
             function(responseObject, responseHeaders) {
               $scope.errors = 'Gene search is temporarily down';
             }
-          ).$promise.then(function(response) {
-            return response.results;
-          });
+          ).$promise.then(
+            function(response) {
+              var selectedIDs = Object.keys($scope.selectedGenes);
+              var geneResultsList = response.results.filter(function(result) {
+                return selectedIDs.indexOf(result.id.toString()) === -1;
+              });
+              return geneResultsList;
+            }
+          );
         };
 
         $scope.onSelect = function($item, $model, $label) {
--- a/interface/src/app/gene/searchMany/gene-search-form.tpl.html
+++ b/interface/src/app/gene/searchMany/gene-search-form.tpl.html
@@ -8,4 +8,8 @@
     ng-disabled="loadingSearchResults">
     Search <i class="fa fa-search"></i>
   </button>
+  <button type="button" class="btn btn-sm" ng-click="switchToFew()">
+    I would like to search individual<br>
+    genes using autocomplete instead.
+  </button>
 </form>
--- a/interface/src/app/gene/searchMany/gene-search-form.tpl.html
+++ b/interface/src/app/gene/searchMany/gene-search-form.tpl.html
@@ -9,7 +9,6 @@
     Search <i class="fa fa-search"></i>
   </button>
   <button type="button" class="btn btn-sm" ng-click="switchToFew()">
-    I would like to search individual<br>
-    genes using autocomplete instead.
+    I would like to search for one gene at a time
   </button>
 </form>
--- a/interface/src/app/gene/searchMany/gene-search-panel.tpl.html
+++ b/interface/src/app/gene/searchMany/gene-search-panel.tpl.html
@@ -1,23 +1,25 @@
 <div class="panel panel-default">
   <div class="panel-heading">
-    <h3>
-      Search for Genes
-    </h3>
+    <b>Copy and paste a list of genes into the text box</b>
   </div>
   <div class="panel-body">
-    <div class="col-xs-3">
-      <gene-search-form search-results="searchResults" queries="queries"
-      organism="{{organism}}">
-      </gene-search-form>
+    <div class="row">
+      <div class="col-xs-4">
+        <gene-search-form
+          search-results="searchResults"
+          queries="queries"
+          organism="{{organism}}"
+          switch-to-few="switchToFew()">
+        </gene-search-form>
+      </div>
+      <div class="col-xs-8">
+        <search-result-table search-results="searchResults" queries="queries">
+        </search-result-table>
+      </div>
     </div>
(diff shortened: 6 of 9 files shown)


cgreene:
Spiffy that clicking between the two keeps the genes selected!

cgreene:
I think the autocomplete view is looking pretty solid to me! I would love to hear other experiences.

The one thing that I noticed that we might consider changing is that as one adds genes, the search bar moves to the right (until it wraps). This is fine since it's annoying to add a lot of genes this way, but if you click back and forth w/ the batch add it can take a moment to find the search bar. You might consider having the search bar always positioned below the set of selected genes.

cgreene:
The copy/paste a set of genes view is a bit complex. I'm not sure exactly what to suggest there, but I think that bit can still be confusing. I don't usually think to look below the green button to see what genes are selected.

You might also consider using the responsive elements to lay things out. As I drag things around, sometimes they end up overlapped. See attached - not sure if this interface works at all on a phone:

screenshot from 2017-07-11 14-58-08

Also, do we need the header "Search for Genes to view as Network" along with the button as "View Genes as Network"? I actually didn't notice the header at first. I wonder if we need it? If you do feel that we need it, should we simplify the button?

cgreene:
Random thoughts. Would love to see what others experience :smile: :+1:

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -15,37 +15,38 @@ angular.module('adage.gene.searchMany', [
         'main': {
           templateUrl: 'gene/gene-network.tpl.html',
           controller: ['$stateParams', 'MlModelTracker', 'UserFactory',
-          function($stateParams, MlModelTracker, UserFactory) {
```
rzelayafavila: *Note: The following changes (from line 18-49) are simply to correct an indentation issue, which was pointed out in issue #197 , but code did not really change. A couple of documentation lines were changed to stay under the 80-char limit.

jaclyn-taroni:
I agree with @cgreene's comment about having the selected genes be listed above the green button on the "copy/paste genes" page

rzelayafavila:
Ok @cgreene, @jaclyn-taroni - I switched the button to be below the table of Selected Genes. Does that one bit look better?

jaclyn-taroni:
That change LGTM

rzelayafavila:
@cgreene : I also agree that the autocomplete text-input should go below the genes that have already been selected. I will change this.

rzelayafavila:
@cgreene Ok, done with that as well!

cgreene:
I think these are improvements. I can't entirely put my finger on it, but there's something that feels a bit odd about this flow. 

screenshot from 2017-07-11 16-22-40

A user pastes lists of genes into A. Clicks search (B). Then things show up at C. They add things with D. Genes show up at E. And then they finally click F to view.

I see that there is a horizontal rule to try to separate the top bit from the bottom bit, to help users understand that the top thing happens first, then the bottom thing. I wonder if there's a way to strengthen some signifiers to help users flow through this, or if there's a way to make this a bit more natural.

Side note: the things I marked with the number sign seem to have broken tooltips (the right tooltip shows up, but it seems to apply to the wrong element?) in chrome on linux.

rzelayafavila:
Aha, ok, working on those now @cgreene

rzelayafavila:
Ok @cgreene, I made the changes based on the suggestions from @jaclyn-taroni . I also removed the "Search Genes..." header, as you both suggested. A new version is up and running on the live instance I have on my computer. How do you think it looks? I am still working on solving the issues with the tooltips, but let me know how the other stuff looks!

interface/src/app/gene/searchMany/searchMany.js
dongbohu: :+1:

dongbohu:
Suppose a user types "dna" in the search box and select "dnaA", when he/she types "dna" again, can you remove "dnaA" from the list?

cgreene:
Wow! Awesome! Love the UI changes! This feels smooth. :+1: Was this one of the reusable components that also ends up automagically back in Tribe as well once you update it, or was that something we were working towards?

cgreene:
Just saw @dongbohu's point :point_up: - not sure how hard that is, but it'd be a level up to not include things in autocomplete that are already selected.

mhuyck:
I agree with the autocomplete observation... in fact, it's issue #69. So, if it's easy to do here then you get a twofer. If not, we've already got it documented.

interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
```diff
@@ -1,30 +1,40 @@
...
+    <button type="button" class="btn btn-sm" ng-click="switchToMany()">
+      I would like to copy and paste a list of genes instead.
+    </button>
+  </div>
+
+  <br>
   <div class="form-group">
-    <button class="btn btn-success" ng-click="sendToNetwork()">
+    <button class="btn btn-success btn-lg" ng-click="sendToNetwork()">
       View Genes as Network <i class="fa fa-check"></i>
```
mhuyck: I'm not really clear why we are using the checkmark or the green `btn-success` style. I do like that the button is at bottom right of the `searchMany` form. That and the `btn-lg` size set it apart enough to be found easily, in my opinion.

interface/src/app/gene/selected/selected-genes-panel.tpl.html
```diff
@@ -1,26 +1,15 @@
...
+<div>
+  <span ng-repeat="gene in selectedGenes">
+    <selected-gene-button gene="gene"></selected-gene-button>
+  </span>
 </div>
+<br>
+
+<span class="pull-right">
+  <button class="btn btn-success btn-lg" ng-click="sendToNetwork()">
+    View Genes as Network <i class="fa fa-check"></i>
```
mhuyck: As noted above: I think we can do without the checkmark or the `btn-success` style. The `pull-right` is perfect.

interface/src/app/gene/searchFew/autocomplete-search-panel.tpl.html
```diff
@@ -1,16 +1,12 @@
 <div class="panel panel-default">
   <div class="panel-heading">
-    <h3>
-      Search for Genes
-    </h3>
+    <b>Type in gene name to search using autocomplete</b>
```
mhuyck: I think this succinct instruction is just right. Even using the term "autocomplete" (which I complained about on the button above) is helpful in this context, as it gives an idea of what is about to happen *right now* when the feature is used.

mhuyck:
I'm getting a regression of bug #123 when I use the `searchMany` form. If you have done a rebase to include that code already, then I think you just need to rebuild search indexes on your demo server and that will be fixed.

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
```diff
@@ -1,23 +1,27 @@
...
-    <div class="col-xs-9">  
-      <search-result-table search-results="searchResults" queries="queries">
-      </search-result-table>
-    </div>
-  </div>
-  <div class="panel-footer">
-    <button type="button" class="btn btn-link" ng-click="switchToFew()">
-      Switch to searching with autocomplete.
-    </button>
+
```
mhuyck: this extra space looks out of place to me

rzelayafavila:
@mhuyck Hmm, it looks like I am up to speed with integrating/rebasing all the previous merges, so I will try to reubild the search indexes, and I'll let you know when that is done, to see if it works.

cgreene:
@mhuyck / @rzelayafavila - is it possible to quickly run a test to see if the tests pass before doing a full rebuild?

rzelayafavila:
@cgreene : ah yes, there is, I just ran it and it just failed.

rzelayafavila:
Am rebuilding the test indexes so that the tests will pass

interface/src/app/gene/searchMany/searchMany.js
```diff
@@ -15,37 +15,38 @@ angular.module('adage.gene.searchMany', [
...
+              //
+              // *Note: This boolean value wasn't getting properly changed
+              // in the child directives when it was passed as a primitive.
+              // It needed to be either set as a property inside a new object
+              // (which was then placed in the $scope object), or made part of
+              // the controller instance object when using 'controllerAs'
+              // syntax. We chose to follow this last approach as it is
+              // considered a best practice. For more information on this, see:
+              // https://github.com/angular/angular.js/wiki/Understanding-Scopes
+              self.autocomplete = true;
```
mhuyck: Thank you for the excellent comments here. Since this view is "in charge" of the overall layout, I'd say this is exactly the level at which the property should be tracked.

I am a little confused about the naming, though. Does `searchMany` refer specifically to the "paste a list of genes" feature? If so, I think the overarching view should be factored apart from the sub-components and given a name that reflects that, such as `adage.gene.search`, to distinguish from the `adage.gene.searchMany` component. Maybe too much to ask from this PR... if you think so, let's create a separate issue.

interface/src/app/gene/selected/selected-genes-panel.tpl.html
```diff
@@ -1,26 +1,15 @@
...
-        <table class="table table-striped"> 
-          <tr ng-repeat="gene in selectedGenes">
-            <td>
-              <selected-gene-button gene="gene"></selected-gene-button>
-            </td>
-          </tr>     
-        </table>
-      </div>
-    </div>
-  </div>
```
mhuyck: *Thank you* for simplifying this. That chain of `<div>`s made me dizzy!

interface/src/app/gene/selected/selected-genes-panel.tpl.html
```diff
@@ -1,26 +1,15 @@
...
+    <selected-gene-button gene="gene"></selected-gene-button>
+  </span>
 </div>
+<br>
+
+<span class="pull-right">
+  <button class="btn btn-success btn-lg" ng-click="sendToNetwork()">
+    View Genes as Network <i class="fa fa-check"></i>
+  </button>
+</span>
```
mhuyck: This is a nice. Much tighter and more component-like now. I think I'd still prefer to see things like headers (lines 1-3) deferred to the view that assembles the components. That keeps overall layout-oriented concerns where they can be best addressed while making the component as compact and re-usable as possible. Maybe to be addressed as a separate refactoring issue, as discussed above.

rzelayafavila:
@mhuyck Yeah, I just needed to rebuild the index. Want to check it out again and let me know if the issue is solved?

mhuyck:
Yep, that got it!

rzelayafavila:
Whoops, hit the `Close and comment` instead of the `Comment` button by accident :stuck_out_tongue:

mhuyck:
So annoying how close those two buttons are. I suppose it's there because not everyone merges right after "finishing" feedback on the PR. It makes more sense to me to keep the conversation open as long as the code isn't yet merged.

mhuyck:
Let me just add: I like this flow a lot. So much that I'm thinking of how we can use some of these ideas in the workflow for Experiment and Sample search to build our sample list for analysis. I won't be able to work it into #178, but possibly for #142. Tagging those here for future reference.

interface/src/app/gene/searchMany/gene-search-panel.tpl.html
rzelayafavila: I agree, I'll remove it

dongbohu:
So we will address #69 later?

mhuyck:
I think that's up to @rzelayafavila to tell us if the change is small enough to fit into this PR.

mhuyck:
If further changes are made in this PR I'd be happy to take a look at them. (I think it should be implicit that if non-trivial changes are made after approval is granted then reviewers should have a chance to get another look.)

rzelayafavila:
@dongbohu and @mhuyck : I was just talking to Casey, because there were a few things I was addressing in this PR before finishing it up (among them some of the changes from Matt's review). I'll give issue #69 a shot this afternoon, but if it looks like it will take too much time, I'll just leave it alone (as we said, it is already documented as an issue).

interface/src/app/gene/searchFew/autocomplete-search-form.tpl.html
```diff
@@ -1,30 +1,40 @@
...
+    <div ng-show="noResults">
+      <i class="fa fa-remove"></i> No genes found for query
     </div>
   </div>
 
   <p class="bg-danger">{{ errors }}</p>
 
+  <div class="control-group">
+    <button type="button" class="btn btn-sm" ng-click="switchToMany()">
+      I would like to copy and paste a list of genes instead.
```
rzelayafavila: Hmm, this is an interesting idea - would you like to make an issue for it?

interface/src/app/gene/searchMany/gene-search-form.tpl.html
```diff
@@ -8,4 +8,8 @@
     ng-disabled="loadingSearchResults">
     Search <i class="fa fa-search"></i>
   </button>
+  <button type="button" class="btn btn-sm" ng-click="switchToFew()">
+    I would like to search individual<br>
+    genes using autocomplete instead.
```
rzelayafavila: Simplified.

interface/src/app/gene/selected/selected-genes-panel.tpl.html
rzelayafavila: I know, me too!!

interface/src/app/gene/searchMany/searchMany.js
rzelayafavila: I agree with this, @mhuyck , I will create a separate issue for refactoring this.

interface/src/app/gene/selected/selected-genes-panel.tpl.html
cgreene: If this styling gets removed, also remove the blue from the searchMany search button to avoid a mismatch in levels of emphasis.

rzelayafavila: Aha, I see what you are saying about the green and the checkmark - I was talking to @jaclyn-taroni though, and I think we might stick with this design for now, but can always change it later depending on user feedback.

rzelayafavila:
Ok @cgreene @jaclyn-taroni @dongbohu @mhuyck : I managed to implement the feature that filters out gene search results that have already been selected. You can play with them and let me know how they look! Thanks!

rzelayafavila:
I added a change for the autocomplete search so that genes that weren't showing up before now show up if some of the search results that originally showed up have been selected. I also opened this PR in django-genes to make sure this happens smoothly: https://bitbucket.org/greenelab/django-genes/pull-requests/24/quick-change-making-the-maximum-number-of/diff

interface/src/app/gene/selected/selected-genes-panel.tpl.html
mhuyck: 👌

mhuyck:
Hmm. Is the demo still at http://165.123.67.154:8000/#/gene_search?mlmodel=1 ? It's hanging for me at the moment.

rzelayafavila:
@mhuyck: Good question - no, I had stopped it briefly to run some tests, but it is back up and running if you want to try it again!

interface/src/app/gene/searchFew/searchFew.js
```diff
@@ -51,17 +50,21 @@ angular.module('adage.gene.searchFew', [
...
-          });
+          ).$promise.then(
+            function(response) {
+              var selectedIDs = Object.keys($scope.selectedGenes);
+              var geneResultsList = response.results.filter(function(result) {
+                return selectedIDs.indexOf(result.id.toString()) === -1;
+              });
+              return geneResultsList.slice(0, numResultsToReturn);
+            }
+          );
```
mhuyck: This is a nice simple fix that improves #69.

It's a nice catch, also, that you want to let new entries into the autocomplete list as those that are added get filtered out here. I think there's a flaw in this approach, however: you'll never be able to add more than the top `GENES_API_RESULT_LIMIT` search results.

Here's how I tested:

1. perform an autocomplete search using `PA`
2. add the top entry
3. repeat 1 & 2 fifteen times
4. typing `PA` into the autocomplete search returns no more results, even though there are still more to be found (confirmed by typing `PA1` and getting more results)

I see two ways around this.

A rigorous fix would include sending the list of genes filtered at the front end into the autocomplete API endpoint and have the back-end filter those out before building a new list of `GENES_API_RESULT_LIMIT` results to send back to the front end. This would be a complete fix but requires more updates to the whole stack (including another revision to the `django-genes` API).

A more pragmatic approach would be to specify a much higher `GENES_API_RESULT_LIMIT` (maybe 100?) and recognize that most users probably won't be using the autocomplete approach to enter more than a few dozen genes.

rzelayafavila: @mhuyck I agree, but I think this isn't really a bug, but rather an enhancement that can be done, since this method already returns 5 more results than what it used to (as it only shows 10 at a time).

I have thought about this, and I came up with similar implementations as the ones you were mentioning (especially the first one). In terms of the second one, yeah, we could send in the 'limit' parameter with this request here and just ask for however many genes we want. However, I think that in either case this is an enhancement that we can track as an issue and not really a bug or part of this PR. What do you think?

rzelayafavila: I can open an issue for this.

mhuyck: Ok. If we have this tracked as an issue I'm totally fine with dealing with it that way. I do think a user would consider this a bug if they encountered it... the distinction between "feature limitation" and "bug" doesn't really matter if you can't do what you expected. 🤓

dongbohu: Agree that this is more like an enhancement. I think that setting `GENES_API_RESULT_LIMIT` to 20 will be probably enough for most use cases.

adage/requirements.txt
```diff
@@ -12,7 +12,7 @@ six==1.10.0
 unicodecsv==0.14.1
 urllib3==1.21.1
 wsgiref==0.1.2
-django-genes>=0.13
+django-genes>=0.14
```
dongbohu: :+1:

rzelayafavila:
Ok, merging this! Thank you so much everyone for all your help with this PR. This actually ended up solving many more issues than it set out to (including some in django-genes!), but it is finallly done! Thanks!

Issue #93, opened 2017-03-03
Clicking search result in `analyze` triggers glitchy auto-scroll


mhuyck:
This used to work properly, but now the point to which the search results scroll seems to have reverted to the top of the browser window rather than just under the navbar as intended.

mhuyck:
I'm looking at #178, and this glitch is not helping with the flow on that issue. Going to tackle this before trying out some ideas there.

mhuyck:
Hah. Well, this issue is also entangled with #63. Turns out that the auto-scroll behavior was due to the use of `$location.hash` in `analyze.js`. Going to rip out this feature entirely and make it work again properly in the context of #63.

mhuyck:
Found the simple solution I was expecting for this. PR coming shortly.

mhuyck:
I may still change this behavior as I work through #63, but this is a good solid improvement so it makes sense to put it in now.

mhuyck:
Fixed in #199

PR #199, opened 2017-07-12
[FIX] adjust CSS to resolve issue #93

mhuyck:
Beautifully simple fix that resolves the issue where it arose: in CSS.
Courtesy of: https://github.com/twbs/bootstrap/issues/1768#issuecomment-46519033


mhuyck:
Hmm. Where did CircleCI go?

interface/src/app/analyze/analyze.js
```diff
@@ -58,9 +54,7 @@ angular.module('adage.analyze', [
       },
 
       scroll_to_id: function(id) {
-        $log.info('scroll_to_id called with: ' + id);
```
dongbohu: Is this supposed to be debug line?

mhuyck: Aha, nice catch. I'll take that out.

mhuyck: Heh. I *did* take that out! Note that it's got a `-` in that diff view.

mhuyck: Ah... but I'm noticing that I removed the `$anchorScroll` entirely but it's still getting injected above. I'll fix that too.

dongbohu: Yeah, I noticed that this line is removed. I just want to understand the exact reason. Thanks.

mhuyck: Right. It had been left in there from when I was initially building the feature. There is no reason for it to stay... it was just for debugging.

mhuyck:
Thanks for the speedy review, @dongbohu!

I see that CircleCI is back. I'll make sure it's happy with my update removing `$anchorScroll` before merging.

Issue #63, opened 2017-02-02
Experiment and Sample pages need to be parametrized


mhuyck:
The `experimentDetail.tpl.html` and `sampleDetail.tpl.html` pages use `href` properties in `<a>` tags, but would work more harmoniously with Angular's `ui.router` handling state parameters.

mhuyck:
Step 1: re-factor to allow `Experiment`s and `Sample`s to be viewed on their own.
Step 2: proper parameterization is now totally trivial.

mhuyck:
Reminder for during refactoring: need to migrate from snake_case in `analyze.js` (and elsewhere?) to camelCase.

PR #200, opened 2017-07-13
Highlight add sample

mhuyck:
This is a work in progress to address #178. Code is not ready for review, but pushing this so I can get feedback once I spin up a demo.


mhuyck:
Demo now up and running at http://greenelab.fgtech.com .

These are smaller tweaks aimed at addressing the specific issue from issue #178. I'd like to do some larger changes but they will require refactoring, which I plan to tackle in #142. @jaclyn-taroni, can you tell me if these tweaks help at all?

jaclyn-taroni:
@mhuyck I think the "add" button needs to be larger. It's probably too verbose but I think it would make sense if it said something like "analyze these samples."

I have some broader comments about this flow but I can save them for #142. Let me know what would be helpful/where we should talk about them.

mhuyck:
I can bump the size of the button up. I agree that "analyze these samples" would be too verbose. Part of the challenge with this current design is that any button to take action must be repeated for every row. Too big or verbose and it will clutter up the whole screen. What if I included a "tool tip" with more of an explanation when the cursor hovers over the button? That would provide the explanation without adding visual clutter.

This solution is really a band-aid for the current layout. I would love to hear your broader comments over at #142. Do you want to write up your thoughts and post them there? We could also meet via appear.in to get that conversation started and I could do the work of capturing your feedback and starting a new mock-up.

jaclyn-taroni:
How about bigger button, different color (white?) with the + sign glyph, that says something like "analyze these samples" in the hover?

jaclyn-taroni:
Part of my issue with the + sign earlier is that I completely missed it.

mhuyck:
Agree that the original `+` was too subtle. So you mean the text of the button should say `+ add`, not just the + glyph, right? (If I make it `+` and white then we're almost back where we started...)

What do you think of the yellow highlight when clicking on a search hit? The idea is to call attention to the way to take action on what's selected. The non-selected buttons are intentionally subdued a little (they stand out more now that they have borders) to avoid having the repetition of the grid draw attention away from the primary task of finding the experiment of sample in the search results.

jaclyn-taroni:
> Agree that the original + was too subtle. So you mean the text of the button should say + add, not just the + glyph, right? (If I make it + and white then we're almost back where we started...)

That sounds like a good plan. I think the borders alone are a big help.

> What do you think of the yellow highlight when clicking on a search hit?

Not sure I follow - is this implemented at [greenelab.fgtech.com](http://greenelab.fgtech.com/) at the moment?

mhuyck:
Yes. Clicking on one of the search hits (not on the "add" button) will display details about that Experiment or Sample. It always did that, but now when you do there is a yellow highlight on the button.

jaclyn-taroni:
Ah, gotcha. I like that idea. Related note: I like Deb's idea mentioned in #142 about the ability to define groups at that stage, but I will put more info over there on that issue.

mhuyck:
Demo site now has larger buttons and the + glyph. You will need to force-refresh your browser to see the changes.

mhuyck:
Yes. I have an idea how to address Deb's suggestion and might be able to do it before a full refactor. Let's discuss that over at #142.

mhuyck:
Let me know if you think this band-aid is sufficient improvement for now and I'll wrap up this PR and start focusing on larger changes over at #142.

mhuyck:
Oh, I forgot to add in the tooltip idea. I'll figure that out and update here once that's worked in too.

mhuyck:
Tooltips have proven to be an unexpected challenge. The `ui.bootstrap` Angular component has a `tooltip` feature (https://angular-ui.github.io/bootstrap/#!#tooltip ), but it seems to be a limited implementation. What happens when I enable the tip we discussed on the `add` button is shown in the image below:

screen shot 2017-07-20 at 11 17 53 pm

The black callout to the right of the top `add` button is getting clipped by the boundary of the table that encloses the button. It's actually even more limited than that, as shown in the next one:

screen shot 2017-07-20 at 11 18 16 pm

Here we can see that the bottom is clipped as well, indicating that the tooltip is bound within the table cell. This makes it not much of a tooltip, in my opinion. While I'm sure this implementation works fine for elements that are not enclosed within small bounding elements like this, it certainly does not work for us.

I took a stab at trying to work around this limitation, thinking that I could adjust the `z-index` for the tooltip via CSS, but it didn't work. I'm going to file a separate issue to get tooltips working for us (perhaps by upgrading to a newer version of Angular or `ui.bootstrap`) but that won't happen right away.

mhuyck:
Well, duh. If I put the tooltip on the left, it works out okay. I'm still annoyed by what appears to be a severe limitation (I think they'd look better on the right, where there is more room) this will work fine:

screen shot 2017-07-21 at 12 14 05 pm

And after clicking, it seems to still look okay:

screen shot 2017-07-21 at 12 15 41 pm

mhuyck:
Okay, this is now ready for code review. The demo server at http://greenelab.fgtech.com/ is updated to match this code.

interface/src/app/analyze/analyze.js
```diff
@@ -52,6 +53,21 @@ angular.module('adage.analyze', [
         return classList;
       },
 
+      item_tooltip: function(search_item) {
```
rzelayafavila: Hmmm, I think I might be missing something - any reason for using underscores instead of camelCase (here and also above in the `item_style` function and the `search_item` argument, and below in the `scroll_to_id` function)?

interface/src/app/analyze/analyze.js
```diff
@@ -52,6 +53,21 @@ angular.module('adage.analyze', [
...
+        // Determine what text to show as a tooltip for this search_item.
+        if (SampleBin.hasItem(search_item)) {
+          return 'already added to analysis';
```
rzelayafavila: Just curious - any reason for not capitalizing the first letter of these messages?

interface/src/app/analyze/analyze.js
```diff
@@ -52,6 +53,21 @@ angular.module('adage.analyze', [
...
+        }
+        if (search_item.item_type === 'sample') {
+          return 'add this sample to analysis';
+        }
+        if (search_item.item_type === 'experiment') {
+          return 'add these samples to analysis';
```
rzelayafavila: I really like this feature, of returning a custom message in the tooltip - I think it is very helpful!

interface/src/app/analyze/analyze.less
```diff
@@ -18,12 +18,21 @@
       background-color: inherit;
     }
     .btn:hover {
-      background-color: #f8f8f8;
+      background-color: @navbar-default-bg;
```
rzelayafavila: 👍

interface/src/app/analyze/analyze.less
```diff
@@ -18,12 +18,21 @@
...
       opacity: .25;
     }
   }
+  li.active {
+    .btn {
+      background-color: @highlight-general;
+    }
+    .btn.disabled {
+      background-color: inherit;
+      opacity: .25;
```
rzelayafavila: 👍

interface/src/app/analyze/analyze.js
mhuyck: Yes, but not a great reason: this is old code.

I do need to fix *all* of those snake_case labels, but plan to do so when I'm refactoring because this will have a ripple effect into all of the other areas that touch them. I've made a reminder in #63, where I'll be doing that bit of refactoring.

interface/src/app/analyze/analyze.js
mhuyck: Good question. I guess we should have a consistent style for that. I'm thinking that because it's a "tip" (and not a complete sentence) the capitalization is not needed. What do you think?

interface/src/app/analyze/analyze.js
rzelayafavila: Ok, sounds good, if this will all happen with with the refactoring, and it is being tracked.

interface/src/app/analyze/analyze.js
rzelayafavila: Hmmm, I'm still on the fence about this - I don't think I feel too strongly either way.

rzelayafavila: Ah, so I just talked to @jaclyn-taroni, and it does sound like capitalizing the first letter would be a good idea. Additionally, we were discussing putting the word `ADD` in all caps as well.

mhuyck: I strongly object to `ADD` in all caps, but if you wanted initial caps to be our style on button labels I'd be okay with that. @jaclyn-taroni, can you explain why you'd want the add button to YELL?

I do not have a strong opinion about tool tips, so I'll make those sentence case.

- [ ] change tool tips to sentence case

jaclyn-taroni: They would probably be more noticeable. If you think lowercase is easier to read, that's fine.

rzelayafavila: I don't have a strong opinion about the all caps, but I agree that making the tool tips sentence case is a good idea :+1:

mhuyck: Okay, I'm making the button and the tool tips capitalized (initial caps only, or sentence case). I'll go more in depth on my thinking about the ALL CAPS issue in the follow-up over at #142, but I don't have time to write that up at the moment.

mhuyck:
Okay, I think I've got everything now. Looks like I'm going to have to resolve a couple of merge conflicts, but I can handle that later tonight. Please let me know if you spot any feedback I missed @rzelayafavila and @jaclyn-taroni.

mhuyck:
Okay, rebase succeeded... now merging.

PR #212, opened 2017-07-19
[FIX] upgrade InstanceType with more memory

mhuyck:
This is a very small change to handle larger data loads without trouble and improve overall web
server performance.


mhuyck:
We've been running on this InstanceType since the deployment in April, but this change didn't make it into the `config.py.template` file. My bad.

mhuyck:
Thanks @rzelayafavila! I'll let CircleCI finish up before merging.

PR #213, opened 2017-07-24
Samplebin mlmodeltracker

### LINKED ISSUE #169: Update SampleBin to defer to the MlModelTracker for mlmodel state
In reworking the page flow for PR #161, the SampleBin (rightly) is no longer the source of state tracking for the selected `MLModel`. This design change needs to be fully reflected in the source for `sampleBin.js`.

mhuyck:
This PR fixes #169.


interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -510,7 +510,7 @@ MathFuncts, errGen) {
   function SampleBinCtrl($scope, SampleBin, MlModelTracker) {
     // give our templates a way to access the SampleBin service
     $scope.sb = SampleBin;
-    $scope.modelInfo = MlModelTracker;
+    SampleBin.mlModelInfo = MlModelTracker;
```
rzelayafavila: I know that this works and does not really make a difference in terms of functionality, but would it make sense to change the order of these two statements like this:
```
SampleBin.mlModelInfo = MlModelTracker;
$scope.sb = SampleBin;
```
?
I feel like this makes more sense intuitively since we are setting the `mlModelInfo` property to `MlModelTracker` first and *then* assigning $scope.sb to this, which will let the page access this information.

interface/src/app/analyze/analysis/sampleBin.tpl.html
```diff
@@ -1,5 +1,5 @@
-<li ui-sref-active="active" class="sample-bin" ng-if="modelInfo.title">
-  <a ui-sref="analysis-detail({mlmodel: modelInfo.id})">
+<li ui-sref-active="active" class="sample-bin" ng-if="sb.mlModelInfo.title">
+  <a ui-sref="analysis-detail({mlmodel: sb.mlModelInfo.id})">
```
rzelayafavila: 👍

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -30,7 +30,7 @@ angular.module('adage.analyze.sampleBin', [
 function($log, $cacheFactory, $q, Sample, Activity, Signature, SignatureSet,
 MathFuncts, errGen) {
   var SampleBin = {
-    selectedMlModel: {
+    mlModelInfo: {
```
rzelayafavila: Does it make more sense to keep track of this information here or to just inject the `MlModelTracker` factory as a dependency to this factory and grab the information from there?

mhuyck: Great question. My goal with this PR was to keep to the original design of `sampleBin` for now but make sure the retro-fit that @dongbohu did when he created `MlModelTracker` was fully integrated.

The entire `sampleBin.js` is due for refactoring... it is quite overgrown. I know I'm punting a lot here, but I'd like to put this change into the scope of the #63-related refactoring work, which I am working to organize next (and will maybe be broken out as a separate issue there too, but put into a logical sequence that helps create a better structure for this part of the code).

interface/src/app/analyze/analysis/sampleBin.js
mhuyck: I think I see what you're saying, and I agree. The `MlModelTracker` line got added during @dongbohu's big retro-fit and he had attached `MlModelTracker` directly to the `$scope` in the same way he did for other pages.

To get that retro-fit to work properly with the original design for this module, `SampleBin` really needs to have a handle to `MlModelTracker`, which is why I changed that. If we think of setting `SampleBin.mlModelInfo` as an "initialization" step, then swapping the lines the way you suggest does make better sense.

I'll swap it around. This might change with `sampleBin` refactoring, but it's good to be thinking about it this way.

rzelayafavila: Ok sounds good!

interface/src/app/analyze/analysis/sampleBin.js
```diff
@@ -508,9 +508,10 @@ MathFuncts, errGen) {
 
 .controller('SampleBinCtrl', ['$scope', 'SampleBin', 'MlModelTracker',
   function SampleBinCtrl($scope, SampleBin, MlModelTracker) {
+    // initialize mlModelInfo property on SampleBin
```
rzelayafavila: :+1: , also for the documentation line!

interface/src/app/analyze/analysis/sampleBin.js
rzelayafavila: Gotcha, that makes sense, if we are keeping track of it in that issue or another one that gets broken out like you say.

mhuyck:
Awesome. Thanks, @rzelayafavila!

Issue #207, opened 2017-07-17
Figure out a system to deploy the interface using Docker


rzelayafavila:
Sub-issue of issue #193

rzelayafavila:
Resolved with PR #214

PR #214, opened 2017-07-26
Dockerizing interface for adage-server. Also removing unused 'placeholders' dependency, which was causing problems when trying to Dockerize.

rzelayafavila:
This addresses issue #207


interface/.dockerignore
```diff
@@ -0,0 +1,4 @@
+bin
+build
+node_modules
```
kurtwheeler: Are these `node_modules` something that were created by `npm` and now can be cleaned up since `npm` isn't installing packages on the host machine anymore? (I don't see the file/directory within the repo.)

kurtwheeler:
I think it'd be good to update the [build section of the README](https://github.com/greenelab/adage-server#build-adage-server-web-interface) to reflect the new build process. I'm not sure how complicated your `docker build/run` commands are but if they're fairly complex it might be nice to include a script and just make the instructions for building the interface be something like "Run `build_interface.sh`".

interface/.dockerignore
mhuyck: `node_modules` is a cache (enormous dumping ground, really) where `npm` downloads its packages. The vast majority of what's in there is used for the toolchain that builds the interface. A small number of packages are plucked out and bundled into our minified JS payload as part of the final steps in the build.

Since it is downloaded for the build, nothing from the `node_modules` directory should appear in our repo. In fact, you'll see that our `.gitignore` lists `interface/node_modules/` to be sure we don't accidentally add any of it.

mhuyck: Which is to say yes, I think you may be right. If we are at the stage where Docker is handling the whole interface build process then `node_modules` could get cleaned out. Also: HOORAY for Docker builds!!! The primary reason our builds are dodgy is because of `npm`, so if we've got a Docker image containerizing that headache now I am very very very happy.

😄 😸 😃 😺 😄 😸 😃 😺 😄 😸 😃 😺 😄 😸 😃 😺 😄 😸 😃 😺 😄 😸 😃 😺

kurtwheeler: Hmm interesting. Was the dodginess of our builds at all related to the `node_modules` cache? If not then it might decrease the build time if that directory is mounted as a Docker `volume` so that `npm` can use its cache.

mhuyck: I'd say that the dodgyness derived from two things:

1. The process of downloading and installing the large volume of packages into `node_modules` is fragile, which seems to be attributable to `npm` itself — builds on my dev instance rarely have trouble, but the AWS deployment, which must re-create that cache every time, will most commonly fail with an error downloading or installing something into `node_modules`.
2. Several `npm` packages from our toolchain are old and crufty. We need to go through and bring our toolchain into 2017. That process will be easier to handle once we have a robustly Dockerized interface build.

kurtwheeler: > but the AWS deployment, which must re-create that cache every time, will most commonly fail with an error downloading or installing something into node_modules.

Hmm well now this exact process is happening inside the Docker container. @rzelayafavila did you experience any issues like that while developing this?

mhuyck: @kurtwheeler: to address your question more directly, though: I think you're suggesting that we keep the `node_modules` on the host and map that into the Docker container. That would more or less leave things as they are.

I think if we have a Docker image that houses `node_modules` in one of its layers and that becomes a base image that our CI process pulls to do the build then we will only have `npm`-related headaches when we have to `docker build` that "build image". Am I making sense?

rzelayafavila: @kurtwheeler and @mhuyck : I did experience some issues when downloading things into node_modules npm (this is what happened yesterday), since sometimes those servers may be out. However, I agree with @mhuyck that we should keep the `node_modules` folder inside the container itself, as one of its layers, so that this install & build process is reproducible by anyone (both when we deploy the interface files and when somebody else wants to use the container to build the interface reproducibly.)

rzelayafavila: And I also agree that we should update all of our frontend toolchain, but that it would make sense to do it once we have a sturdy Docker environment making builds consistently.

kurtwheeler: Yes, yes you are :)

That sounds like a really good plan. It's a similar strategy to the one I'm using in the Data Refinery for its dependencies. They take a long time to run so I've hosted a base image [here](https://hub.docker.com/r/ccdl/data_refinery_worker_base/) so that if someone wants to develop on the Data Refinery they don't have to first wait for 2 hours while the image is built on their machine and so that CI can run in a reasonable amount of time.

kurtwheeler: Oh sorry @rzelayafavila I didn't see your comments when replying. To summarize this thread:
* We don't need the `node_modules` directory anymore. These should be `rm -r`'d from your machines and dropped from the `.dockerignore` file.
* This `Dockerfile` should be split into two, adding a `Dockerfile.base` which contains all dependencies and is hosted in Dockerhub. This `Dockerfile` should then be based off of that one and probably not do too much more than copy the code in and run `grunt`.

Does that seem accurate?

rzelayafavila: Sure, I think the splitting the Dockerfiles into two and making a `Dockerfile.base` makes sense. I guess we can totally rm -r the node_modules folder from our machines and drop it from the .dockerignore file here, as long as that does not interfere with people's development environments while we switch over to always having the Docker container build these files (even on our development machines). Do you think this makes sense @mhuyck?

rzelayafavila:
@kurtwheeler - sure, I'll add that to the README (without overwriting what is already there, so that people can still use that other approach while we make the total switch over to Docker), and I'll add the shell script that builds and runs the images.

interface/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM phusion/baseimage:0.9.22
```
mhuyck: Very happy to know about `phusion/baseimage`!

interface/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN groupadd user && useradd --create-home --home-dir /home/user -g user user
+WORKDIR /home/user
+
+# libfontconfig is needed for phantomjs, which is apparently installed in
+# our ubuntu machines, but required with this phusion/baseimage:0.9.22
```
mhuyck: Thank you for the clear documentation!

interface/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+  nodejs-legacy \
+  npm \
+  git \
+  libfontconfig
+
+# Install global node packages.
+RUN npm -g install grunt-cli karma-cli bower
+
+# Clean up APT when done.
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
mhuyck: Should we be combining these three `RUN` statements with `&&` (like you have on line 8) to minimize creation of extra layers?

interface/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Copy code to server directory
+COPY . .
+
+USER user
+
+RUN npm install
+
+RUN bower install
```
mhuyck: Similar question here: combine these two `RUN`s with `&&`?

interface/build.config.js
```diff
@@ -86,9 +86,7 @@ module.exports = {
       // greenelab.stats.ttest.js from ES6 to ES5 to fit with the rest of
       // our tools.
       'node_modules/ttest/greenelab.stats.ttest.js',
-      'node_modules/multtest/greenelab.stats.multtest.js',
-
-      'vendor/placeholders/angular-placeholders-0.0.1-SNAPSHOT.min.js'
```
mhuyck: buh-bye crufty stuff. :tada:

interface/src/app/about/about.js
```diff
@@ -18,12 +17,7 @@ angular.module('adage.about', [
 })
 
 .controller('AboutCtrl', function AboutCtrl($scope) {
-  // This is simple a demo for UI Boostrap.
-  $scope.dropdownDemoItems = [
-    'The first choice!',
-    'And another choice for you.',
-    'but wait! A third!'
-  ];
```
mhuyck: :tada:

interface/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+  libfontconfig
+
+# Install global node packages.
+RUN npm -g install grunt-cli karma-cli bower
+
+# Clean up APT when done.
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Copy code to server directory
+COPY . .
```
mhuyck: Just realized something: do we need to be careful that we're not copying `config.py` (which contains secrets) into our Docker image here?

kurtwheeler: Hmm, well since we're going to be splitting the dockerfile into two and making one of them public via dockerhub, we should be careful not to put them in the public image. However we can have our secretes on the private image as long as we're careful about it.

rzelayafavila: Actually we don't need to worry about it at all for this Dockerfile, since it only includes the files in this directory (so it will not include any of the files in the parent directory, or its subdirectories, like the config.py file). In fact, I don't think we will need any secrets for the interface portion. However, I will be careful to make sure and ignore it in the Dockerfile for the backend.

interface/build.config.js
rzelayafavila: :smile: !

interface/Dockerfile
rzelayafavila: Yeah, good call - I will join them so that they are in the same layer.

interface/Dockerfile
rzelayafavila: Sure, I can do that

interface/Dockerfile
mhuyck: Aha. I missed that context. If we're doing this from `interface/`, not from the repo root then we should be good with secrets. Thanks, and sorry for the false alarm!

rzelayafavila:
Ok @mhuyck and @kurtwheeler - I think I have addressed everything, including adding the documentation for running the shell script that builds the container to the README.md file. Please let me know if anything else should be changed! Thanks!

interface/docker_build_interface.sh
```diff
@@ -0,0 +1,22 @@
...
+    mkdir $compile_folder
+    chmod 775 $compile_folder
+fi
+
+docker run \
+      --volume $compile_folder:/home/user/compile_folder \
+      adage-server/docker-interface-build
+
+mv $compile_folder/bin ./
+mv $compile_folder/build ./
```
kurtwheeler: <nit-picky ignorable="true">
  Couldn't `./` just be `.`?
</nit-picky>

kurtwheeler:
LGTM!

interface/docker_build_interface.sh
rzelayafavila: True, I'll change it since it is cleaner

.dockerignore
```diff
@@ -0,0 +1,3 @@
+data
+interface
+adage/adage/config.py
```
mhuyck: 👍

interface/.dockerignore
```diff
@@ -1,4 +1,3 @@
 bin
 build
-node_modules
-vendor
+adage/adage/config.py
```
mhuyck: This one isn't needed, right? It's the one in the root-level `.dockerignore` that we want I think.

interface/Dockerfile
```diff
@@ -1,29 +1,8 @@
...
-  npm \
-  git \
-  libfontconfig
-
-# Install global node packages.
-RUN npm -g install grunt-cli karma-cli bower
-
-# Clean up APT when done.
-RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+FROM adage-server/docker-interface-base
```
mhuyck: 👍 

Woot!

interface/docker_build_interface.sh
```diff
@@ -0,0 +1,22 @@
+docker build -t adage-server/docker-interface-base -f Dockerfile.base .
```
mhuyck: What happens if `adage-server/docker-interface-base` is already up to date? Does this step needlessly re-build it?

interface/Dockerfile
```diff
@@ -1,29 +1,8 @@
...
 
-USER user
-
-RUN npm install
-
-RUN bower install
+RUN grunt
 
-CMD ["grunt"]
+CMD mv bin compile_folder && mv build compile_folder
```
mhuyck: Nice! I like how this neatly collects the output. I think I might have used `artifacts` instead of `compile_folder`, to use terminology consistent with CircleCI (and Travis, I think), but that is a minor quibble.

interface/docker_build_interface.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+docker run \
+      --volume $compile_folder:/home/user/compile_folder \
+      adage-server/docker-interface-build
+
+mv $compile_folder/bin .
+mv $compile_folder/build .
+
+rmdir $compile_folder
```
mhuyck: Ahh... I see. If the `compile_folder` directory is wiped out in the end then its name becomes less important. Why aren't we just mapping the Docker container straight to `bin/` and `build/`, then? Is there some benefit to the copying step?

interface/notes_for_s3_interface_deployment.md
```diff
@@ -0,0 +1,18 @@
...
+Initialize S3 bucket using AWS console or something like terraform
+(https://www.terraform.io/)
+
+Upload bin folder to S3 bucket using aws-cli 'sync' command
+
+```shell
+# Inside the current folder (adage-server/interface), run
+aws s3 sync ./bin s3://<your s3 bucket name>
+
+```
```
mhuyck: I still don't see much benefit to using an S3 bucket for this (some negatives, in fact: cost, plus extra administrative overhead), but maybe I'm missing something. We don't need to nail down this deployment detail just yet anyway.

interface/.dockerignore
mhuyck: I think this one will just uselessly filter out a file called `interface/adage/adage/config.py`, which does not exist.

rzelayafavila: Yep, it's true, I added that one to the root-level .dockerignore file, so I'll remove this. I guess I was thinking of the use when somebody was trying to run the Docker container to build the interface from a folder other than `interface/` but I guess that wouldn't matter because the paths in this file are relative to where this file is.

interface/docker_build_interface.sh
rzelayafavila: Well, it "rebuilds" it, but the way Docker works is that it if everything is already up to date and there are no changes, then it just uses the cache of the previously built layers, so it takes very little time (a couple of seconds), and does not actually build and store any new images of the container. However, this is a good way to check if anything has changed.

interface/docker_build_interface.sh
rzelayafavila: Yes - I tried that approach before, and it caused me to bang my head against my desk for quite a while, because it seems that you cannot delete folders that are the direct volume mounting points for Docker. However, this is exactly what grunt does, so we are not able to call these two folders as the volume mounting points directly.

interface/docker_build_interface.sh
mhuyck: Awesome! I was hoping that was the answer here but not sure how automatic Docker made that process.

interface/docker_build_interface.sh
mhuyck: Gotcha, that makes sense then. Could you add a brief comment here (maybe above line 19?) and in `interface/Dockerfile` to explain that, since it's not obvious?

mhuyck: Oh, and: sorry about your head! 🤕

interface/notes_for_s3_interface_deployment.md
rzelayafavila: Yes I can see that - I talked to @dcgoss and @cgreene, and since this is the way that things are deployed in cognoma, I decided to go ahead and try it out. It worked, and I put some notes for that process in this markdown file to keep track of it, in case we decide to deploy this way. We definitely don't need to nail this down right now (at least for this pull request) - we can all keep talking about it and decide what the best strategy for deployment is.

mhuyck: Well, of course it will work... just seems unnecessary to me. We're talking about assets that total less than 1.4MB. Small enough to fit on a floppy, for those who even remember what those are!

cgreene:
I liked @dcgoss's rationale. I think I can summarize it as: "Why spend time/effort configuring and deploying an interface for static files? S3 will do that for you, and you're going to have to pay storage costs one way or another anyway."

interface/notes_for_s3_interface_deployment.md
cgreene: Replied in the wrong place:
I liked @dcgoss's rationale. I think I can summarize it as: "Why spend time/effort configuring and deploying an interface for static files? S3 will do that for you, and you're going to have to pay storage costs one way or another anyway."

rzelayafavila: @mhuyck , yep, the question was not if it would work, but just getting it done with this repo, and tracking the steps for it. Haha, yep, I actually not only remember those floppy disks, but also the original 5-1/4" actual floppy disks!

interface/docker_build_interface.sh
rzelayafavila: Haha, thanks! Yep, I'll put in a comment about it!

interface/Dockerfile
rzelayafavila: Hmmm - now that you know that this `compile_folder` will be wiped out, do you still think it's name should be changed to `artifacts`?

interface/notes_for_s3_interface_deployment.md
dcgoss: @dcgoss is back! Honestly the process of deploying to S3 is so simple that in the time that it takes to discuss whether you want to use S3 it can already be done. If we decide on another option, migrating from S3 is about the same effort as starting from square one anyways.

dcgoss: And for the record, it costs about 2 cents per month to store 2 MB.

mhuyck:
If it really is simple and cheap to go the S3 route, that's great! I just don't see how all of the pieces fit together. As I said... maybe I'm missing something. At any rate, let's not try to sort out that detail now since it's not really relevant to this PR. We'll nail it down when we get to the actual Docker deployment changes next month!

interface/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM adage-server/docker-interface-base
+
+# Copy code to server directory
+COPY . .
```
dcgoss: This is super minor, but it would be nice if you could substitute the full path in for the second period

rzelayafavila: Sure

interface/Dockerfile
mhuyck: Not necessary now that I understand that... we only need to be sure this directory name won't clobber anything that already exists.

rzelayafavila:
Ok @mhuyck, @dcgoss, I think I got everything - please let me know if I missed anything! Thanks!

interface/docker_build_interface.sh
```diff
@@ -12,6 +12,11 @@ if [ ! -d "$compile_folder" ]; then
     chmod 775 $compile_folder
 fi
 
+# The reason we mount the parent "compile_folder" and then move the
+# "bin" and "build" folders to the current directory is that Docker will
+# not allow the mounting points or folders themselves to be deleted.
+# However, deleting these folders is exactly what the "grunt" command needs
+# to do right before it builds and compiles the interface files.
```
mhuyck: 👍

mhuyck:
Think you've got it all. Let's merge and deploy!

Issue #216, opened 2017-08-03
Break off Elasticsearch into its own container/service.


rzelayafavila:
This is a sub-issue of issue #193

rzelayafavila:
Resolved with PR #217

PR #217, opened 2017-08-07
Adding changes needed to break off Elasticsearch and run it in its own container.

rzelayafavila:
Resolves Issue #216 

Upgrading django-haystack version to 2.6.0 in requirements.txt (as it is needed for Elasticsearch v2.x), and adding a stub for the steps needed to run a local deployment of this webserver. This stub right now contains the commands to pull the appropriate Elasticsearch Docker image and run it so that the django project can access it.

### DIFF
--- a/adage/requirements.txt
+++ b/adage/requirements.txt
@@ -1,7 +1,7 @@
 argparse==1.2.1
 Django==1.8.11
 django-fixtureless==1.4.3.3
-django-haystack==2.4.1
+django-haystack==2.6.0
 django-tastypie==0.13.3
 elasticsearch==1.9.0
 gunicorn==19.4.5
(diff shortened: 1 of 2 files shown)


adage/requirements.txt
```diff
@@ -1,7 +1,7 @@
 argparse==1.2.1
 Django==1.8.11
 django-fixtureless==1.4.3.3
-django-haystack==2.4.1
+django-haystack==2.6.0
```
rzelayafavila: Note - this newer django-haystack version is needed in order to work with Elasticsearch v2.x

rzelayafavila:
Thanks @kurtwheeler and @dcgoss!

Issue #219, opened 2017-08-08
Update the Dockerfile for the 'adage' Django project container


rzelayafavila:
Sub-issue of issue #193

rzelayafavila:
Resolved with https://github.com/greenelab/adage-server/pull/221

Issue #193, opened 2017-07-05
Switch to Docker-based (instead of fabric-based) deployment for this webserver.


rzelayafavila:

cgreene:
Some work on this in #12 but that is woefully out of date. Might be easiest to lift the bits that are still usable but dump the PR itself.

cgreene:
BTW that one is stuck where secrets are needed to deploy. My impression is that CircleCI 2.0 makes this easier + there are some solutions available that others in the lab now have more familiarity with. :smile:

rzelayafavila:
@cgreene : Ah gotcha, ok, I'll take a look at it and lift the parts that we can use. Thanks for pointing to it!

mhuyck:
Aha, thanks for the reminder about the specific issue with secrets, @cgreene. @rzelayafavila, when we were talking my intuition was to recommend prototyping with a simplified example but I couldn't remember why. I think a prototype deployment process would still be a good idea so we can flush out and resolve any issues like the secrets without wrestling with the full complexity of a deployment all at once. Having a stable overall process will allow us to build on it incrementally.

mhuyck:
I am noting here that I need to:
- [ ] post my Djangocon brain dump notes about Docker migration

mhuyck:
Okay, @rzelayafavila, I have captured a brain dump of the thoughts I've had about this since I started mulling over how to do this around a year ago. Some of this might be misguided given my limited experience with Docker, but I hope you will find these notes helpful.

# Docker deployment ideas

## Overall goals

1.  deployment completes reliably and efficiently with a single command
2.  build system is separate from deployable "target" objects
    (for example, we should not need `npm` or `grunt` commands in the 
    production container)
3.  a developer can pull our Docker image and get up and running with minimal
    configuration (maybe we could provide a "starter database" in a separate
    Docker image to simplify setup?)
4.  an independent lab should be able to set up their own instance of ADAGE
    with minimal configuration

## Gotchas and other notes

1.  need to avoid putting AWS keys into a repo or other non-secure service
2.  if using AWS, need to figure out boto methods for reallocating elastic IP
    address if we want to automate that last step
3.  upon successful deployment, need to terminate old EC2 instance
4.  if using existing fabfile, need to skip database bootstrapping and re-use
    existing database
5.  current deployment steps (using fabfile) throw lots of warnings about
    old `npm` tools we're using; updates to deployment should review these
    warnings and modernize the build process

## Options for getting CI working NOW

[...]

PR #221, opened 2017-08-09
Updating Dockerfile for 'adage' Django project container. Also adding shell script to get this container to be able to talk to the local postgres host.

rzelayafavila:
Resolves issue #219

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,10 @@
-data
 interface
 adage/adage/config.py
+
+**/*~
+**/*.pyc
+
+data/gene_history
+data/*.gene_info
+data/bootstrap_cache_*
+data/*_pickled_genesets
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,9 @@ COPY data data
 COPY adage adage
 WORKDIR adage
 
+# The next couple of files are part of the 'get_pseudomonas' repository
+# in bitbucket. That code does not really belong in this repository, but
+# is needed for deployment of this server.
 RUN wget https://bitbucket.org/greenelab/get_pseudomonas/raw/tip/get_pseudo_sdrf.py && \
     wget https://bitbucket.org/greenelab/get_pseudomonas/raw/tip/gen_spreadsheets.py
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,17 +4,12 @@ FROM phusion/baseimage:0.9.22
 # To make it public, use -p flag
 EXPOSE 8000
 
-# Directory containing ADAGE server source code
-ENV ADAGE_SRC=adage
-
-# Directory used in container
-ENV ADAGE_SRV=/srv
-
 # Create required directories
-WORKDIR $ADAGE_SRV
+WORKDIR /srv
 RUN mkdir static logs
 
 RUN apt-get update && apt-get install -y \
+  wget \
   python \
   python-pip \
   python-psycopg2 # Install here so that postgres lib dependency is met.
@@ -25,15 +20,22 @@ RUN pip install --upgrade pip
 
 # Copy requirements.txt file and install requirements here to save time
 # when building this docker image again if no requirements have changed
-COPY $ADAGE_SRC/requirements.txt requirements.txt
+COPY adage/requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 
 # Clean up APT when done.
 RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Code to server directory
-COPY $ADAGE_SRC .
+# Copy necessary data files to bootstrap database
+COPY data data
+
+# Copy code for Django project
+COPY adage adage
+WORKDIR adage
+
+RUN wget https://bitbucket.org/greenelab/get_pseudomonas/raw/tip/get_pseudo_sdrf.py && \
+    wget https://bitbucket.org/greenelab/get_pseudomonas/raw/tip/gen_spreadsheets.py
 
 # Copy entrypoint script into the container
-COPY ./docker-entrypoint.sh /
-ENTRYPOINT ["/docker-entrypoint.sh"]
+COPY docker-entrypoint.sh .
+ENTRYPOINT ["./docker-entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,8 @@ RUN apt-get update && apt-get install -y \
 # be out of date with an older version of pip
 RUN pip install --upgrade pip
 
-# Copy requirements.txt file here and install requirements to save time
+# Copy requirements.txt file and install requirements here to save time
+# when building this docker image again if no requirements have changed
 COPY $ADAGE_SRC/requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,37 +1,38 @@
-FROM phusion/baseimage:0.9.19
+FROM phusion/baseimage:0.9.22
 
-# Use baseimage-docker's init system.
-CMD ["/sbin/my_init"]
+# Make available on port 8000 - however, it is not public yet
+# To make it public, use -p flag
+EXPOSE 8000
 
 # Directory containing ADAGE server source code
 ENV ADAGE_SRC=adage
+
 # Directory used in container
 ENV ADAGE_SRV=/srv
-# Directory where ADAGE code lives in the container
-ENV ADAGE_SRVSRC=$ADAGE_SRV/$ADAGE_SRC
+
+# Create required directories
+WORKDIR $ADAGE_SRV
+RUN mkdir static logs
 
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,37 +1,39 @@
...
 # Directory containing ADAGE server source code
 ENV ADAGE_SRC=adage
+
 # Directory used in container
 ENV ADAGE_SRV=/srv
-# Directory where ADAGE code lives in the container
-ENV ADAGE_SRVSRC=$ADAGE_SRV/$ADAGE_SRC
+
+# Create required directories
+WORKDIR $ADAGE_SRV
```
dcgoss: To me, setting these things with environment variables is a little confusing and adds extra steps. It seems more direct and clear to just specify the directory

Dockerfile
```diff
@@ -1,37 +1,39 @@
...
-# Code to server directory
-COPY $ADAGE_SRC $ADAGE_SRV
+# Upgrade pip to avoid issues with some of the installation tools that may
+# be out of date with an older version of pip
+RUN pip install --upgrade pip
 
-# Install ADAGE deps
+# Copy requirements.txt file and install requirements here to save time
+# when building this docker image again if no requirements have changed
+COPY $ADAGE_SRC/requirements.txt requirements.txt
```
dcgoss: Awesome copying requirements.txt first!

docker_local_deploy.sh
```diff
@@ -20,3 +20,11 @@
 # https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-gsg.html
 docker pull library/elasticsearch:2.3
 docker run -d -p 9200:9200 elasticsearch:2.3
+
+HOST_IP=$(ip route get 8.8.8.8 | awk '{print $NF; exit}')
```
dcgoss: See comment below

Dockerfile
```diff
@@ -1,37 +1,39 @@
-FROM phusion/baseimage:0.9.19
+FROM phusion/baseimage:0.9.22
```
dcgoss: Is there a reason why you use phusion as the base image?

local_postgres_docker_setup.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
dcgoss: Couldn't you just run a [postgres container](https://hub.docker.com/_/postgres/) instead of dealing with all the networking to localhost?

Dockerfile
kurtwheeler: I agree with this, I think I echoed a similar thought when Rene and I originally paired on this Dockerfile.

local_postgres_docker_setup.sh
rzelayafavila: Hmm, I guess that's true. Good point @dcgoss - that would probably be even better!

Dockerfile
rzelayafavila: Mmm gotcha, I can see that. I think @kurtwheeler had said something along the same lines, so I will change these to not use env variables.

Dockerfile
rzelayafavila: Good question - I stuck with phusion because it was the one that was already being used by this Dockerfile (which I did not write), but I think I have talked to other programmers in the lab and the consensus is that there was really no super-good reason for using it. Which other Ubuntu 16.04 images do you recommend?

kurtwheeler: I'm using `python:3.6.1-slim` in one of my docker images, but just raw `ubuntu:16.04` in another. You can see the differences between the two [here](https://github.com/docker-library/python/blob/d3c5f47b788adb96e69477dadfb0baca1d97f764/3.6/jessie/slim/Dockerfile). (Note that the python-slim image is based off of Debian instead of Ubuntu)

rzelayafavila: Yeah ... I would like to stick to using a Ubuntu image instead of Debian or other linux. Is the raw ubuntu one very large?

local_postgres_docker_setup.sh
rzelayafavila: Actually, on second thought, and given my conversation with @kurtwheeler about the possible danger of running production databases on containers (without mounting them first to a volume on the local computer), I think I will create a whole separate issue for this, to decide how it would be implemented - I think it might be outside the scope of this PR.

rzelayafavila:
Ok @kurtwheeler and @dcgoss - I made some changes and opened an issue for the Postgres container, since it doesn't really fall in the scope of this PR. Let me know how this looks, thanks!

.dockerignore
```diff
@@ -1,3 +1,10 @@
-data
 interface
 adage/adage/config.py
+
+**/*~
+**/*.pyc
```
kurtwheeler: What does the double `*` do here?

Dockerfile
```diff
@@ -25,15 +20,22 @@ RUN pip install --upgrade pip
...
-COPY $ADAGE_SRC .
+# Copy necessary data files to bootstrap database
+COPY data data
+
+# Copy code for Django project
+COPY adage adage
+WORKDIR adage
+
+RUN wget https://bitbucket.org/greenelab/get_pseudomonas/raw/tip/get_pseudo_sdrf.py && \
+    wget https://bitbucket.org/greenelab/get_pseudomonas/raw/tip/gen_spreadsheets.py
```
kurtwheeler: I'm curious as to why this is how we retrieve these files instead of including them in this repo or installing them via pip?

.dockerignore
rzelayafavila: So it is a wildcard string "that matches any number of directories (including zero)" (https://docs.docker.com/engine/reference/builder/#dockerignore-file). It basically ignores any files ending in `~` and `.pyc` in the current folder or any of the subdirectories.

Dockerfile
rzelayafavila: Great question - I think ideally these are ones that would be pip-installable, since (as I understand it) they are part of another project, which is maintained separately from this webserver repository (is this right @cgreene?). So far, we have been downloading them for the deployment process. I agree - ultimately I think we want to have all the modules in that project be pip-installable, so I can create an issue for that.

cgreene: I copied @mhuyck's deployment process. These are ultimately things that would come from the data refinery or other projects. The code to build a data compendium and such doesn't belong in the ADAGE server. Probably should be commented though.

local_postgres_docker_setup.sh
cgreene: Is this for a production situation? I thought this was mostly for dev, with RDS or equivalent used n production.

Dockerfile
dcgoss: If a slimmer image works, I would go with that as it's always easier to work with lightweight containers. Seems like they've slimmed down the official ubuntu image to around ~125MB so should be okay

local_postgres_docker_setup.sh
rzelayafavila: Yes, this is mostly for development (since we would use RDS or equivalent in our lab production instance), but we were also talking about the case where collaborators and other researchers would want to spin up their own small instances of the adage-server, maybe on their machines. This case would not be a development instance, but also we cannot assume that RDS/equivalent would be available.

Dockerfile
rzelayafavila: Yep - actually these lines were not in the original Dockerfile before this PR, I was putting them in because I was also replicating the current deployment in the fabfile (and these files are needed for deployment). For sure, I will  add a comment for these lines.

local_postgres_docker_setup.sh
cgreene: I think that you can say: "we recommend RDS/etc for production, but include a postgres container for development." That seems entirely reasonable.

rzelayafavila:
Ok @kurtwheeler , I think I have made all necessary changes and added the comments suggested by @cgreene, let me know how it looks!

kurtwheeler:
LGTM! Approved! (I think since I already did it before these changes I can't approve it via github again.)

rzelayafavila:
Great, thank you @kurtwheeler!

local_postgres_docker_setup.sh
mhuyck: What @cgreene says is precisely what I was thinking. We definitely want to use a properly tuned, optimized and backed up server for our production DB, and we get that with RDS on AWS. At the same time, a Docker-based DB would fulfill two of the goals I think we should have:

> a developer (or independent lab) can pull our Docker image and get up and running with minimal configuration

Noted here: https://github.com/greenelab/adage-server/issues/193#issuecomment-313806195

rzelayafavila: Right, this is what I was thinking as well. We were always thinking of using RDS or equivalent for production, but those are the advantages of using a Docker-based DB (for developers and other researchers).

Dockerfile
mhuyck: There's an issue for this... aha, so old that it's over at Bitbucket:

> https://bitbucket.org/greenelab/adage-server/issues/2/clean-up-code-imported-from

Basically, this messiness is one of the reasons we need the data refinery!

PR #246, opened 2017-08-18
Update django-genes version in requirements.txt

dongbohu:
Yet another trivial PR!


dongbohu:
I have updated the package on PyPI, by the way:
https://pypi.python.org/pypi/django-genes

Issue #218, opened 2017-08-08
Create a Dockerfile/container for nginx


rzelayafavila:
Sub-issue of issue #193 
- [ ] Have this container hook up the `adage` Django project container (which should be serving the Django project using gunicorn) and the interface files, wherever they are located.

Issue #225, opened 2017-08-11
Figure out if we should make Postgres its own Docker container for purposes of development and other small, local (non-AWS or equivalent)- based instances, and implement it if so


rzelayafavila:
This is a sub-issue of issue #193 

This came up in https://github.com/greenelab/adage-server/pull/221

mhuyck:
I added a comment over at #221, but that's already closed so it really belongs here.

We definitely want to use a properly tuned, optimized and backed up server for our production DB, and we get that with RDS on AWS. At the same time, a Docker-based DB would fulfill two of the goals I think we should have:

> a developer (or independent lab) can pull our Docker image and get up and running with minimal configuration

Noted here: https://github.com/greenelab/adage-server/issues/193#issuecomment-313806195

rzelayafavila:
I agree @mhuyck, it is what I was thinking as well. I think we've always wanted to use RDS or equivalent for production, and for the reasons you describe, having a Docker-based DB for developers and other researchers. I think I should've made the title of this issue more specific to address this - I will change it.

rzelayafavila:
Resolved via https://github.com/greenelab/adage-server/pull/247

PR #247, opened 2017-08-25
Adding commands and configuration file to run Docker containers for NGINX and Postgres

rzelayafavila:
This PR aims to resolve https://github.com/greenelab/adage-server/issues/218 and https://github.com/greenelab/adage-server/issues/225


adage/adage/config.py.template
```diff
@@ -217,5 +217,29 @@ CIRCLECI_CONFIG.update({
...
 
+DOCKER_DEV = CIRCLECI_CONFIG.copy()
+DOCKER_DEV.update({
+    'databases': {
+        'default': {
+            'ENGINE': 'django.db.backends.postgresql_psycopg2',
+            # database name and user cannot have upper case letters
+            'NAME': 'docker_adage',
+            'USER': 'docker_adage',
+            'PASSWORD': 'password123',
```
dongbohu: So do we expect the users to customize this database password? If yes, we may want to add some comments here.

interface/docker_build_interface.sh
```diff
@@ -1,17 +1,37 @@
...
-# if it doesn't already exist.
-script_directory=`dirname "${BASH_SOURCE[0]}"  | xargs realpath`
 compile_folder="$script_directory/compile_folder"
+bin_folder="$script_directory/bin"
+build_folder="$script_directory/build"
 
-if [ ! -d "$compile_folder" ]; then
+# Set up mounting volume directory that will house bin and build folders
+# if it doesn't already exist.
+if [ -d "$compile_folder" ]
```
dongbohu: Do you think it makes more sense to use `-e` instead of `-d`? If `$compile_folder` exists but is not a directory, the `mkdir $compile_folder` command on line 19 will fail.

interface/docker_build_interface.sh
```diff
@@ -1,17 +1,37 @@
...
+then
+    rm -r $compile_folder
+else
     mkdir $compile_folder
     chmod 775 $compile_folder
 fi
 
+# Remove previous bin and build folders if they already exist in the
+# interface folder
+if [ -d "$bin_folder" ]
```
dongbohu: Is `-e` better?

interface/docker_build_interface.sh
```diff
@@ -1,17 +1,37 @@
...
+then
+    rm -r $bin_folder
+fi
+
+if [ -d "$build_folder" ]
```
dongbohu: Is `-e` better?

interface/docker_build_interface.sh
rzelayafavila: Hmm, good question. I don't know if `-e` would be better than `-d` here since we really only expect the compile_folder, bin_folder and build_folder to be directories that were previously built by out interface build process. I think if they were something other than directories, that would be kind of unexpected, and maybe we would want this to fail. What scenarios were you thinking of where these would be files that were not directories?

adage/adage/config.py.template
rzelayafavila: I think this is actually more of a dummy, non-secure password for the development docker database. I was thinking it would not be changed by the user. What do you think @dongbohu and @mhuyck?

interface/docker_build_interface.sh
dongbohu: Is it possible that these directories could be soft links? If they must be directories, then it is okay to use `-d` here, because the error generated by `mkdir` command will remind the user of this requirement.

adage/adage/config.py.template
dongbohu: If it is just a development database, I think this dummy password is acceptable.

mhuyck: As long as there is little chance that this configuration would inadvertently be used for deploying anywhere, I think it's fine. I like that this config is explicitly labeled as `DOCKER_DEV`, which helps with that.

interface/docker_build_interface.sh
```diff
@@ -1,17 +1,37 @@
+script_directory=`dirname "${BASH_SOURCE[0]}"  | xargs realpath`
```
mhuyck: I guess we don't have an official style guide for Bash scripts, but the extra space before the `|` bothers me a bit... we don't need that, right?

interface/docker_build_interface.sh
mhuyck: We are setting this up ourselves, so we know for sure if `compile_folder` is a real directory or a symlink (see the `mkdir` at line 19 below). I'd say the `-d` is exactly what we want.

nginx/dev/adage-nginx.conf
```diff
@@ -0,0 +1,63 @@
+server {
+    # default server - drop requests without a Host header
+    listen 80 default_server;
+    server_name "";
+    return 444;
+}
```
mhuyck: This section is useful for security on a production server, but quite inconvenient for development. I think we can remove it and use `server_name "";` at what is currently line 13 below.

Note: I even hacked together a little `sed` command that edits our default `adage-nginx.conf` this way for a development setup. See the `tweak_nginx_for_dev` task in `adage-deploy/fabfile.py`.

nginx/dev/adage-nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+
+    # /static is linked to static files required for use by the adage web application
+    location /static {
+        alias /home/static;
+    }
+
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
```
mhuyck: Cool! This is missing from our current deployment setup.

nginx/dev/adage-nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+
+    # / directs to the index.html in the static folder for the
+    # adage web application
+    # for detail on the P3P header, see http://www.w3.org/P3P/
+    location / {
+        root /home/static;
```
mhuyck: Nice. Now I see how we bypass Django for serving up the interface. This combines with the `location /api` below to serve static interface files or proxy the REST API. I like it. It's good to have a clean separation here.

nginx/dev/adage-nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+        proxy_set_header X-Real-IP $remote_addr;
+        add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"';
+    }
+
+    location /search {
+        proxy_pass http://adage-django:8000/search;
+        proxy_set_header X-Forwarded-Host $server_name;
+        proxy_set_header X-Real-IP $remote_addr;
+        add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"';
+    }
```
mhuyck: Hmm. What is this? All of our searches should be handled via the `/api` above, right? Access to Elasticsearch would happen on its own separate port at `9200`.

nginx/dev/adage-nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+
+    location /tribe_client {
+        proxy_pass http://adage-django:8000/tribe_client;
+        proxy_set_header X-Forwarded-Host $server_name;
+        proxy_set_header X-Real-IP $remote_addr;
+        add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"';
+    }
```
mhuyck: I am less well versed in how `tribe_client` works, so defer to you on setting it up properly. I think it would be architecturally cleaner if there were a way to ensure all of its back-end calls were channeled via `/api`, though. Is there a technical reason we have to set it up this way?

docker_local_deploy.sh
```diff
@@ -21,10 +21,35 @@
 docker pull library/elasticsearch:2.3
```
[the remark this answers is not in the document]
rzelayafavila: Ah good call, you're right, I'll remove this.

docker_local_deploy.sh
```diff
@@ -21,10 +21,35 @@
...
 docker run -d -p 9200:9200 elasticsearch:2.3
 
+# Start up postgres container
+docker run -e POSTGRES_USER=docker_adage -e POSTGRES_PASSWORD=password123 \
+    --name docker-postgres -p 5431:5432 -d postgres
```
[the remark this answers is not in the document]
rzelayafavila: Yes, we are connecting it to port 5431 in case there is already a postgres server running on port 5432 (the traditional postgres port) in whichever machine is running this code. I know this is the case for my machine, so that way I don't have to stop the local postgres server and tie this container to port 5432.

interface/docker_build_interface.sh
rzelayafavila: Ah for sure - that extra space wasn't intentional - nice catch!

adage/adage/urls.py
```diff
@@ -41,7 +40,6 @@
 v0_api.register(ExpressionValueResource())
 
 urlpatterns = [
-    url(r'^$', TemplateView.as_view(template_name="index.html"), name='home'),
```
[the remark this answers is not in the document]
rzelayafavila: Yes, exactly

docker_local_deploy.sh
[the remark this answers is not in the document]
mhuyck: Oh, good call. I'd say that is a very good reason!

interface/docker_build_interface.sh
rzelayafavila: Right, ok, that makes sense - I'll leave it with `-d` then.

nginx/dev/adage-nginx.conf
rzelayafavila: Yep, you're right @mhuyck - I had just copied all of the urls in `urls.py`, but the `/search` endpoint there was just a vestigial remnant from the haystack tutorial, probably, so I went ahead and removed it from there as well.

nginx/dev/adage-nginx.conf
rzelayafavila: Sure - I think that might be outside the scope of this PR, to integrate the `tribe_client` endpoints with all other `/api` endpoints. But yes, that could be a good way of doing it, and possibly other servers would benefit from it down the road, so I'll open an issue for it?

docker-entrypoint.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 python manage.py migrate                  # Apply database migrations
-python manage.py collectstatic --noinput  # Collect static files
+
+python manage.py test
```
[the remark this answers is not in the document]
rzelayafavila: Yes - you're right. The running of the tests should not be here in the first place, so I removed it. I think we probably do not even need to run tests here, since they should have already passed when the CI process runs the tests (or otherwise the code would not be merged into master).

rzelayafavila:
Ok @mhuyck and @dongbohu I think I have gotten everything, let me know how it looks - thanks!

adage/adage/config.py.template
```diff
@@ -226,15 +226,15 @@ DOCKER_DEV.update({
...
+            'HOST': 'db',
+            'PORT': '5432'
         }
     },
     'haystack': {
         'default': {
             'ENGINE':
                 'adage.search_backend.CustomElasticsearchEngine',
-            'URL': 'localhost',
+            'URL': 'elasticsearch',
```
mhuyck: 👍

docker-entrypoint.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
-python manage.py migrate                  # Apply database migrations
-
-python manage.py test
```
mhuyck: 👍

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+    # one of our dependencies, only supports Elasticsearch versions
+    # 1.x and 2.x, not the newer versions (as of 8/7/17). See:
+    # http://django-haystack.readthedocs.io/en/v2.6.0/installing_search_engines.html#elasticsearch
+    #
+    # Also, we are using Elasticsearch v2.3 (even though v2.4.6 is
+    # available) because it is the highest version that AWS supports of the
+    # 2.x major version (as of 8/7/17). The logic here is that we want the
+    # local deployment to mirror our deployment on AWS. See:
+    # https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-gsg.html
+    image: elasticsearch:2.3
```
mhuyck: I think this version is okay if it works for us, but I want to clarify: the AWS support note above does not apply to us since we are rolling our own via Docker here.

I guess if we wanted to we could deploy to the AWS-supplied version of Elasticsearch so it may be wise to keep what we use in sync. I don't quite see a reason to "undo" our current AWS independence, however, unless we happened to run into some issue with search performance or something.

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
+# As of 9-12-17, Amazon ECS CLI supports Docker compose file syntax versions
+# 1 and 2
+# (see: http://docs.aws.amazon.com/AmazonECS/latest/developerguide/cmd-ecs-cli-compose.html)
+version: '2'
```
mhuyck: It is great to have this file to tie everything together for us. I find this much easier to deal with than the previous bash scripts. Thank you for pulling it together, @rzelayafavila!

README.md
```diff
@@ -9,7 +9,49 @@ considered pre-release status. The following instructions detail the steps
 required for getting a development instance up and running manually. For a guide
 to automated deployment, see [Deployment Steps](#deployment-steps) below.
 
-## Get a working instance of the adage-server running
+## Get a working instance of the adage-server running with Docker
+
+We recommend using [Docker](https://www.docker.com/) to get a local working
+instance of the server.
```
dongbohu: Maybe change "server" into "adage-server" to make it more specific?

README.md
```diff
@@ -9,7 +9,49 @@ considered pre-release status. The following instructions detail the steps
...
+1. Install [Docker](https://docs.docker.com/get-started/) on your computer.
+
+   Also, if your computer is not running Mac or Windows OS, you will need to
+   [install Docker Compose](https://docs.docker.com/compose/install/).
+
+2. Fork and clone the adage-server repository
+
+   Fork [the adage-server repository on Github](https://github.com/greenelab/adage-server)
+   (see [Github's documentation](https://help.github.com/articles/fork-a-repo/)
+   for forking repositories) and then clone that fork you made in the directory
```
dongbohu: I would add a comma like this: 
```... repositories), then ...```

File: load_default_pseudomonas_data.sh [hunk 21]
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+# Script to load default Pseudomonas data (in the files included in this
+# repository) into the adage-server database via management commands.
```
mhuyck: There is a `fab` task that does all of this already. Are you running it this way because you're thinking we'll eliminate our dependency on `fabric` once we've got Docker working?

load_default_pseudomonas_data.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Import activity data into the database:
+python manage.py import_activity \
+    ../data/all-pseudomonas-gene-normalized_HWActivity_perGene_with_net300_100models_k=300_seed=123_ClusterByweighted_avgweight_network_ADAGE.txt \
+    "Ensemble ADAGE 300"
+
+# Save Ensembl and PseudoCAP Cross-reference Databases
+python manage.py genes_add_xrdb \
+    --name=Ensembl \
+    --URL=http://www.ensembl.org/Gene/Summary?g=_REPL_
```
mhuyck: Aha. We do not have this step in our `fab` deploy script (just the one for PseudoCAP below). Do we need it?

load_default_pseudomonas_data.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+# Script to load default Pseudomonas data (in the files included in this
+# repository) into the adage-server database via management commands.
+
+# Import Pseudomonas annotation data
+python manage.py import_data ../data/PseudomonasAnnotation.tsv
```
mhuyck: Using relative paths here means that this script must be run from a particular working directory. Can you document that in the comments at the head of the file? I guess an alternative would be to replace all of the relative paths with full paths -- perhaps one coded into an explicit constant.

If we are going in the direction of eliminating `fabric` in this way, we also need to make an issue to remind us to remove the configuration settings from `config.py.template` that hold all of these file names. That would need to be an issue because we'll still need those references to work until the transition to Docker deployment is complete.

load_default_pseudomonas_data.sh
```diff
@@ -0,0 +1,71 @@
...
+    "High-weight genes"
+
+python manage.py import_gene_sample_expr \
+    ../data/all-pseudomonas-gene-normalized.pcl \
+    208964
+
+python manage.py tribe_client_pickle_public_genesets
+
+# Build haystack search indexes:
+python manage.py update_index
```
mhuyck: The way we've been doing this via the `fab` scripts is a full rebuild command, which looks like:

```bash
python manage.py rebuild_index --noinput
```

I think that `update_index` will do mostly the same thing if we're starting with fresh data, as we are here, but have you tested? Also: does `update_index` have a `--noinput` parameter akin to `rebuild_index`? I feel like we needed that but maybe it was just to get past the confirmation about clobbering old indexes, which isn't relevant for `update_index`, I guess.

nginx/dev/adage-nginx.conf
```diff
@@ -1,16 +1,8 @@
...
-    return 444;
-}
-
-# configuration of the server
-server {
     # the port your site will be served on
     listen      80;
     # the domain name it will serve for
-    server_name localhost;
+    server_name "";
```
mhuyck: 👍

nginx/dev/adage-nginx.conf
```diff
@@ -47,13 +39,6 @@ server {
         add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"';
     }
 
-    location /search {
-        proxy_pass http://adage-django:8000/search;
-        proxy_set_header X-Forwarded-Host $server_name;
-        proxy_set_header X-Real-IP $remote_addr;
-        add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"';
-    }
-
```
mhuyck: 👍

load_default_pseudomonas_data.sh
rzelayafavila: Ah, ok, I will look into this and see which one we want.

load_default_pseudomonas_data.sh
rzelayafavila: I think so - I think it is nicer to have these commands here so we do not have to rely on fabric. If I remember correctly, the fabric project was having issues with migrating to Python 3, and as a plus, we do not have to keep maintaining/thinking about our fabric code. Also, as a plus, it is nice to have these commands separated so that somebody who is not using fabric can apply all of these management commands (and also read them in a format that is more reader-friendly than the fabric format). As an example, I was thinking of when we were spinning up an instance with @jaclyn-taroni and wanted to know what exactly all the management commands to populate the database were.

load_default_pseudomonas_data.sh
rzelayafavila: Right - I will make an issue for that but also document the fact that these are relative paths at the beginning of this file, since I think that is the most pragmatic approach.

load_default_pseudomonas_data.sh
mhuyck: I am good with that plan. `fabric` is essentially just a Python-friendly way of scripting remote ssh, and that would be largely redundant for a Docker deployment, I think.

load_default_pseudomonas_data.sh
rzelayafavila: Ah, good catch - I think we don't actually need this here, and I put this in by accident. I will remove it.

load_default_pseudomonas_data.sh
mhuyck: I think the long-term strategy is to integrate with a `data-refinery` to help harvest new data, so, ultimately, we want an `update_index` to run automatically after new data are imported.

But right now we only import data to initialize the database. For now, I just want to make sure what we have works.

rzelayafavila: That's right - I think the `update_index` works, and as you said, it does not need (or even have) the `--noinput` parameter since no indexes would be clobbered. I don't have strong feelings either way. Do you?

rzelayafavila:
Ok @mhuyck and @dongbohu for reviewing and approving this PR! I did want to say that I added one more bit, though, where the docker-entrypoint.sh waits until the postgres container is up and running to run the django app. Let me know if it looks ok! Thanks!

docker-entrypoint.sh
```diff
@@ -1,5 +1,16 @@
...
+# and many times the django container finishes starting up before the
+# postgres container does. If this happens, the django container will
+# error out the first time it tries to access the postgres container.
+# To avoid this, we will wait until the "python manage.py migrate",
+# command to apply database migrations, runs.
+# For more information, see: https://docs.docker.com/compose/startup-order/
+until python manage.py migrate; do
+  >&2 echo "Postgres is unavailable - sleeping"
+  sleep 1
+done
```
mhuyck: Cool! It's nice to have a way to do this. I'm a little surprised that Docker doesn't do this more gracefully in `docker-compose`, actually, given that we have a `depends_on` configuration in the `core` section of our `docker-compose.yml` file. Maybe it's a "not implemented yet" kind of feature.

mhuyck:
LGTM. Thanks for all the hard work, @rzelayafavila!

docker-entrypoint.sh
rzelayafavila: Right, I agree - I thought docker-compose would handle that more gracefully, especially, as you point out, as the `core` `depends_on` the database container.

rzelayafavila:
Ok, thank you @mhuyck and @dongbohu for your great help and feedback!!

Issue #183, opened 2017-06-29
Load (more) complete gene-gene network into the database


dongbohu:
Currently the gene-gene network file that has been loaded into the database has a cutoff value of 0.4. This file includes *139,111* gene pairs. 

Ideally we should load all gene-gene pairs into the database without any cutoff value. According to @tj8901nm, the complete gene-gene network file includes 5,549 * 5,548 / 2 = 15,392,926 gene pairs. Although such a big file is likely to hurt the performance of both back end and front end, we should load it to get a better idea of how much it will affect the server, and possibly set a lower cutoff value with acceptable performance.

dongbohu:
Location of the complete gene-gene network file:

```
Dropbox/GreeneLab/ADAGE_server/gene-gene_network_complete.txt
```

File size: 507 MB

dongbohu:
Two genes `PA2245` and `PA4638` are not found in gene info, so the number of genes that are recognizable is 5,549 - 2 = **5,547**, and the number of gene-gene network records loaded into the database is: 5547 * 5546 / 2 = **15,381,831**.

dongbohu:
Here is a brief summary of edges weights in the complete gene-gene network:
```
0.0 to 0.1:  8,683,750
0.1 to 0.2:  4,560,391
0.2 to 0.3:  1,579,674
0.3 to 0.4:    430,000
0.4 to 0.5:    101,042
0.5 to 0.6:     24,865
0.6 to 0.7:      7,728
0.7 to 0.8:      3,475
0.8 to 0.9:      1,675
0.9 to 1.0:        326
Total:      15,392,926
```
It shows that ~86% edges' weight are less than 0.2, which most users may not be interested in. After talking with @tj8901nm and @cgreene, we decided to upload only the edges whose weight is >= 0.2, instead of loading the complete gene-gene network.

dongbohu:
Fixed in PR #255.

PR #255, opened 2017-09-15
Load a more complete gene-gene network into database

dongbohu:
This PR is the fix of issue #183. It also removes a redundant activity data file (the one that has both positive and negative activity values).  

The management command `import_gene_network.py` has been optimized to speed up the importing process. Right now the size of `bulk create` is set to 2000. I am not sure whether the current AWS instance has enough memory to handle it.


dongbohu:
Once this PR is done, I will also modify the deployment code that imports the gene-gene network.

dongbohu:
You can ignore the file `gene-gene_network_cutoff_0.2.txt.gz`. I created it because the unzipped version is too large to be accepted by github.

adage/analyze/management/commands/import_gene_network.py
```diff
@@ -121,6 +128,10 @@ def check_and_import(file_handle, ml_model):
                             "value in [-1.0, 1.0]" %
                             (line_index + 1, tokens[3] + tokens[2]))
 
+        # Skip the line if the weight is less than cutoff.
+        if abs(weight) < ml_model.g2g_edge_cutoff:
+            continue
```
mhuyck: 👍

adage/analyze/management/commands/import_gene_network.py
```diff
@@ -153,31 +164,42 @@ def check_and_import(file_handle, ml_model):
...
 
-def find_gene(systematic_name):
+def find_gene(systematic_name, cached_genes):
     """
     Return the gene in database whose systematic_name matches input
     "systematic_name".  An exception will be raised if no gene is found
     or multiple genes exist in the database.
     """
+    if systematic_name in cached_genes:
+        return cached_genes[systematic_name]
```
mhuyck: Interesting. So does this help speed things up as well? I would have thought that Django would do some QuerySet caching that would make the `Gene.objects.get` at line 204 work faster on subsequent calls with the same `systematic_name`. Maybe that's not how it works, though.

mhuyck: Aha... okay, answered my own question. The `get()` method does not use a cache:

> https://docs.djangoproject.com/en/1.11/ref/models/querysets/#methods-that-do-not-return-querysets

so it would make sense that you would get a speedup from implementing your own cache.

adage/analyze/management/commands/import_gene_network.py
```diff
@@ -153,31 +164,42 @@ def check_and_import(file_handle, ml_model):
...
+        records.append(
+            Edge(mlmodel=ml_model, gene1=gene1, gene2=gene2, weight=weight)
+        )
+        if len(records) == BULK_SIZE:  # dump bulk records into database
+            Edge.objects.bulk_create(records)
+            records = []
+    # Don't forget the last bulk:
+    if len(records) > 0:
+        Edge.objects.bulk_create(records)
+        records = []
```
mhuyck: Nice. I am sure that `bulk_create` helps a lot.

The only question I have here is what happens if an error is thrown by `bulk_create`? I guess it's no worse than before, since we don't do any exception handling here. Should we consider wrapping things in a transaction so it can be rolled back if there's a failure? I thought we had done that already somewhere but I can't remember where.

dongbohu: @mhuyck: If `bulk_create()` throws an exception, it will be caught by the `except` block on line 57, and the transaction will be rolled back too.

dongbohu: Yes, it helped a lot!

mhuyck: Aha! I *knew* I had seen a transaction wrapper. I just didn't expand enough of the unchanged code in the diff view to see that. 👍

adage/analyze/management/commands/import_gene_network.py
mhuyck: Yeah, the way I read the docs above, there is *no* caching on a Django `.objects.get()` call. This makes sense because you sometimes need to ensure you have the most up to date record.

mhuyck:
This looks great to me. Thanks for answering my questions, @dongbohu!

adage/analyze/management/commands/import_gene_network.py
```diff
@@ -153,31 +164,42 @@ def check_and_import(file_handle, ml_model):
...
+            gene2 = find_gene(tokens[1], cached_genes)
         except Exception as e:
             logger.warning("Input file line #%d will be skipped: %s",
                            line_index + 1, e)
             continue
 
         # Check whether the triplet (ml_model, gene1, gene2) is unique.
-        if not unique_together(ml_model, gene1, gene2):
+        if check_unique and not unique_together(ml_model, gene1, gene2):
             raise Exception("Input file line #%d: (%s, %s, %s) is not unique "
```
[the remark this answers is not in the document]
dongbohu: No. If this exception is raised, it will terminate the program and roll back the database, because we would like the users to go back and check the input file immediately.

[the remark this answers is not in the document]
rzelayafavila: Ok, nice!

rzelayafavila:
BTW @dongbohu, I'm just curious, how long does it take to run this command and add all gene-gene edges > 0.2 on your computer?

dongbohu:
@rzelayafavila: about 3 minutes.

rzelayafavila:
Ok nice, that sounds good - merge at will!

Issue #188, opened 2017-06-30
Sub-issue of #187: Specify which commands are and are not necessary if running fab command.


rzelayafavila:
Sub-issue of #187:
For example, `python manage.py migrate` does not need to be run if using fab command to deploy an instance in localhost.

Issue #187, opened 2017-06-30
Need to update README to include more clear instructions on setting up a deployment instance


rzelayafavila:
These items were discovered while setting up a deployment instance with @jaclyn-taroni. I will make smaller sub-issues to track the items that came up, and link them to this issue. 

- [x] https://github.com/greenelab/adage-server/issues/188
- [x] https://github.com/greenelab/adage-server/issues/189
- [x] https://github.com/greenelab/adage-server/issues/190
- [x] https://github.com/greenelab/adage-server/issues/191

cgreene:
Manuscript reviewer comment:
"Please provide software stack along with instructions to initialize instance of the ADAGE web server so that users can load models of interest as mentioned in the manuscript"

Manuscript is due back Sept 28. Means that we probably need to have these things done ~2 weeks before. With the changes in the software deployment, this puts a very hard deadline on getting the containerization changes done. @rzelayafavila / @dongbohu

rzelayafavila:
Ok, sounds good, so I'll aim to have the containerization all wrapped up by Sept. 15

cgreene:
+ Instructions on how to initialize an instance! :smiley_cat:

rzelayafavila:
Closed final sub-issue of this with the merging of https://github.com/greenelab/adage-server/pull/267

PR #267, opened 2017-09-26
README.md modifications

rzelayafavila:
Modifying README.md so that it contains an example of how to copy and load a data file using docker, and also specifies that the 'python manage.py migrate' management command is not necessary if running fabric init_instance command, along with a couple of other tweaks.

This resolves https://github.com/greenelab/adage-server/issues/188 and https://github.com/greenelab/adage-server/issues/187

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ instance of the adage-server.
 
 4. (Optional) Loading ADAGE models into your new instance.
 
-   If you want to load the default Pseudomonas data into the server database,
-   enter the following command:
+   If you want to load the default _Pseudomonas_ data into the server
+   database, enter the following command:
 
    ```
    docker-compose exec core ./load_default_pseudomonas_data.sh
@@ -111,7 +111,7 @@ instance of the adage-server.
    * `delete_participation_type`,
    * `import_activity`,
    * `import_data`,
-   * `import_gene_network.py`,
+   * `import_gene_network`,
    * `import_gene_sample_expr`, and
    * `import_signature_gene_network`.
 
@@ -259,7 +259,7 @@ and put them in the `adage-server/adage/` folder.
 
 * **Note:** This step is not necessary if you are running the `fabric`
 commands in the first part of the *Populate the database* section below,
-as these commands will automatically run it for you.
+because the `fabric` commands will automatically run it for you.
 
 ```shell
 python manage.py migrate
--- a/README.md
+++ b/README.md
@@ -106,14 +106,14 @@ instance of the adage-server.
    ```
 
    These are the management commands currently available to load data files:
-   * `add_ml_model`,
-   * `create_or_update_participation_type`,
-   * `delete_participation_type`,
-   * `import_activity`,
-   * `import_data`,
-   * `import_gene_network`,
-   * `import_gene_sample_expr`, and
-   * `import_signature_gene_network`.
+   * `add_ml_model`
+   * `create_or_update_participation_type`
+   * `delete_participation_type`
+   * `import_activity`
+   * `import_data`
+   * `import_gene_network`
+   * `import_gene_sample_expr`
+   * `import_signature_gene_network`
 
    To see an example of how these management commands are used, see
    [the load_default_pseudomonas_data.sh script](https://github.com/greenelab/adage-server/blob/master/load_default_pseudomonas_data.sh).
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ instance of the adage-server.
 
    If you would like to load different data files from your own ADAGE model,
    you can do so, provided that they are in the same format as the
-   corresponding files in the `data/` folder.
+   corresponding files in the `data/` folder described above.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,7 @@ instance of the adage-server.
 
    If you would like to load different data files from your own ADAGE model,
    you can do so, provided that they are in the same format as the
-   corresponding files in the `data/` folder.
+   corresponding files in the `data/` folder described above.
```
mhuyck: Minor quibble: in the above-referenced text (at line 56, which I can't comment on directly because it's too far up from this diff) the Latin name _Pseudomonas_ should be italicized. So the line should read:

```
    If you want to load the default _Pseudomonas_ data into the server database,
```

README.md
```diff
@@ -87,15 +87,33 @@ instance of the adage-server.
...
+       /srv/data/SomeOrganismAnnotation.tsv
+   ```
+
    These are the management commands currently available to load data files:
    `add_ml_model`,
    `create_or_update_participation_type`,
    `delete_participation_type`,
    `import_activity`,
    `import_data`,
    `import_gene_network.py`,
```
mhuyck: We should remove the `.py` here, for consistency.

README.md
```diff
@@ -87,15 +87,33 @@ instance of the adage-server.
...
-   `import_gene_sample_expr`,
-   `import_signature_gene_network`
+   `import_gene_sample_expr`, and
+   `import_signature_gene_network`.
```
dongbohu: Do you think it will be more readable if we list these commands with bullets?

README.md
```diff
@@ -231,24 +249,29 @@ cd adage/
...
+
+Download the `get_pseudo_sdrf.py` and `gen_spreadsheets.py` files from
+[this repository](https://bitbucket.org/greenelab/get_pseudomonas/src),
+and put them in the `adage-server/adage/` folder.
+
 ### Run Django migration of database tables
 
+* **Note:** This step is not necessary if you are running the `fabric`
+commands in the first part of the *Populate the database* section below,
+as these commands will automatically run it for you.
```
mhuyck: I think this line should read:

> because the `fabric` commands will automatically run it for you.

to avoid ambiguity.

README.md
```diff
@@ -231,24 +249,29 @@ cd adage/
...
 ```
 
-### Download other necessary files
-
-Download the `get_pseudo_sdrf.py` and `gen_spreadsheets.py` files from
-[this repository](https://bitbucket.org/greenelab/get_pseudomonas/src),
-and put them in the `adage-server/adage/` folder.
+Django will use the database you configured earlier to build tables to support
+its models with this command, but the tables will remain empty
+until you run the management commands to import data into the models.
```
mhuyck: 👍

README.md
rzelayafavila: Yeah - I was wondering that as well. That makes me think that bullet points might be the right approach. I'll give put them in and see how it looks.

mhuyck:
Please let me know if I should run through the steps as an additional test.

README.md
rzelayafavila: Ah, yes, good catch! I hadn't noticed this.

rzelayafavila:
Ok @dongbohu and @mhuyck - I have made changes for all the points that you mentioned, let me know how this looks. @mhuyck - we have tested it, but an additional test never hurts!

README.md
```diff
@@ -87,15 +87,33 @@ instance of the adage-server.
...
    These are the management commands currently available to load data files:
-   `add_ml_model`,
-   `create_or_update_participation_type`,
-   `delete_participation_type`,
-   `import_activity`,
-   `import_data`,
-   `import_gene_network.py`,
-   `import_gene_sample_expr`,
-   `import_signature_gene_network`
+   * `add_ml_model`,
```
dongbohu: The bullets do look better, thanks. I think the trailing comma (lines 109-115), `and` (line 115) and dot (line 116) probably should be removed.

rzelayafavila: Sure, I'll remove them

README.md
```diff
@@ -106,14 +106,14 @@ instance of the adage-server.
...
-   * `import_gene_sample_expr`, and
-   * `import_signature_gene_network`.
+   * `add_ml_model`
+   * `create_or_update_participation_type`
+   * `delete_participation_type`
+   * `import_activity`
+   * `import_data`
+   * `import_gene_network`
+   * `import_gene_sample_expr`
+   * `import_signature_gene_network`
```
dongbohu: Looks good. I don't have any other concerns.

rzelayafavila: Ok, thanks @dongbohu!

mhuyck:
This looks great. I will test it, but that shouldn't hold up this PR (and I don't really expect to find any issues) so we should merge. If I find anything I will make a new issue.

rzelayafavila:
Ok, sounds good, thanks @mhuyck!

Issue #268, opened 2017-09-26
Workflow Refactoring


dongbohu:
Attached is the workflow diagrams that @mhuyck created.

[adage-server-analyze-flow-20170926.pdf](https://github.com/greenelab/adage-server/files/1334337/adage-server-analyze-flow-20170926.pdf)

mhuyck:
These diagrams describe the UI design changes and internal refactoring required to support https://github.com/greenelab/adage-server/issues/142

PR #269, opened 2017-09-28
Refactor experiment into a reusable component & update experiment analysis workflow

mhuyck:


mhuyck:
Hah! Nice to have a pat on the back from codeclimate for once rather than a complaint.

interface/src/app/app.js
```diff
@@ -36,20 +36,15 @@ angular.module('adage', [
...
-    // 'signature_search'. (Used by index.html to highlight the 'signature' tab
-    // on web UI in either state.)
-    $scope.inSignatureStates = function() {
-      var currState = $state.current.name;
-      return currState === 'signature' || currState === 'signature_search';
+      if (stateArr.indexOf(currState) === -1) {
+        return false;
+      } else {
+        return true;
+      }
```
dongbohu: Great! So `inGeneStates()` and `inSignatureStates()` are now merged into one.

interface/src/app/analyze/experiment/experiment.js
```diff
@@ -4,87 +4,89 @@ angular.module('adage.analyze.experiment', [
...
-        .replace(/{mlDataSource}/g, mlDataSource);
-    },
-    arrExpExperiment: function(experimentId) {
-      return ('http://www.ebi.ac.uk/arrayexpress/experiments/{id}/')
-        .replace(/{id}/g, experimentId);
-    }
-  };
-  return Experiment;
-}])
+// // moved to app/experiment.service.js
```
dongbohu: It is just a copy/paste without any code change, right?

interface/src/app/experiment/experiment.component.js
```diff
@@ -0,0 +1,69 @@
...
+      };
+
+      var queryError = function(responseObject, responseHeaders) {
+        $log.warn('Query errored with: ' + responseObject);
+        ctrl.experiment.status = 'Query failed.';
+      };
+
+      ctrl.show = function(id) {
+        if (!id) {
+          $log.warn('experiment.show called with id', id);
```
dongbohu: Do you think the warning message can be more specific? The word `experiment` in `experiment.show` sounds very generic.

interface/src/app/experiment/experiment.component.js
```diff
@@ -0,0 +1,69 @@
...
+      ctrl.$onChanges = function(changesObj) {
+        ctrl.show(changesObj.id.currentValue);
+      };
+      ctrl.makeHref = Experiment.makeHref;
+      ctrl.experiment = {
+        status: '',
+        relatedSamples: []
+      };
+
+      var queryError = function(responseObject, responseHeaders) {
```
dongbohu: The second argument `responseHeaders` is not needed.

interface/src/app/experiment/experiment.component.js
```diff
@@ -0,0 +1,69 @@
...
+                ctrl.experiment.status = '';
+                ctrl.experiment.relatedSamples.push(responseObject.data);
+              }
+            },
+            queryError
+          );
+        };
+        Experiment.get(
+          {accession: id},
+          function(responseObject, responseHeaders) {
```
dongbohu: `responseHeaders` is not needed here.

interface/src/app/experiment/experiment.component.js
```diff
@@ -0,0 +1,69 @@
...
+            if (responseObject) {
+              ctrl.experiment.results = responseObject;
+              ctrl.experiment.status = 'Retrieving sample details...';
+              for (var i = 0; i < responseObject.sample_set.length; i++) {
+                getSampleDetails(responseObject.sample_set[i]);
+              }
```
dongbohu: If you want, you can use `forEach()` to replace this `for` loop. (I found that most `for` loops in JS can be replaced by `forEach`, which is more compact.

interface/src/app/analyze/experiment/experiment.js
mhuyck: Right. The only thing different there is that the `angular.module` statement at top can be a little simpler since this service only needs `ngResource` and `adage.utils`.

Obviously this whole `experiment.js` file will be removed when I'm done. I found it helpful to remind myself where things were going as I refactored.

interface/src/app/experiment/experiment.component.js
mhuyck: True. Thanks!

interface/src/app/experiment/experiment.component.js
mhuyck: I thought about this a little. I want the message to be somewhat generic since this is a reusable component now. On the other hand, I think we can do better by using the name of this module instead, so it could say this instead:

```
'adage.experiment: show() called with id'
```

interface/src/app/experiment/experiment.component.js
mhuyck: I think I wrote this bit of code before I knew about `forEach()`. Progress!

interface/src/app/experiment/experiment.component.js
dongbohu: Yes, I think the new message is much better.

interface/src/app/analyze/search/searchDetail.js
```diff
@@ -18,14 +16,19 @@ angular.module('adage.analyze.detail', [
...
         $scope.detail.search_item = search_item;
-        $scope.detail.showing = true;
-        $timeout(function() {
-          // if we inititate the scroll too early, the search results list
-          // hasn't finished compressing width and it doesn't work right,
-          // so this timeout waits until the transition has finished
-          // before scrolling
-          $scope.analyze.scroll_to_id(search_item.pk);
-        }, 500);
+        if (search_item.item_type === 'experiment') {
```
dongbohu: So the idea is if the matched item of the search is an experiment, the experiment page will be loaded? Is this a new logic? I don't see it in original code.

dongbohu:
@mhuyck: About the convertion of `snake_case` to `camelCase`, please add an issue and put it in `Optional for Next Release`. I don't think you will have time to handle it for this month's release.

interface/src/app/analyze/analyze.tpl.html
```diff
@@ -12,9 +12,11 @@
...
     <div id="search" ng-class="detail.showing ? 'with-detail':''">
       <p>
-        In this area, users will be able to search for an experiment or set of
-        samples to analyze, define analysis groups and generate a DA heatmap to
-        review ADAGE signature intensities for each sample.
+        Search for an experiment or set of samples. Click on a search result
+        to view more details, add them to an analysis with the
+        "<i class="fa fa-plus" aria-hidden="true"></i> Add" button,
+        or use the "Analyze" button to go directly into an analysis with
+        that experiment or sample added.
```
rzelayafavila: :+1: Nice instructions for the user!

interface/src/index.html
```diff
@@ -74,19 +74,22 @@
                 Download
               </a>
             </li>
-            <li ui-sref-active="active">
+            <li
+            ng-class="{active: inStateArr(['analyze', 'experiment'])}">
```
rzelayafavila: This is super minor, but I notice that the `ng-class` here (and in all of the changes in this file) is lined up with the `<` in the line above, where as in `analysis/sampleBin.tpl.html`, it is offset by one space, so it is lined up with the `li` (which I think makes slightly more sense). What do you think? As I said, it is super minor, so I don't have really strong feelings about it.

interface/src/app/sampleBin/addItemAnalyze.component.js
```diff
@@ -0,0 +1,43 @@
...
+      ctrl.sb = SampleBin;
+
+      ctrl.addItemAnalyze = function() {
+        SampleBin.addItem(ctrl.item);
+        $state.go('analysis-detail', {'mlmodel': MlModelTracker.id});
+      };
+
+      ctrl.itemTooltip = function() {
+        if (!ctrl.item) {
+          return 'ititializing...';
```
rzelayafavila: Typo: `initializing` instead of `ititializing`

interface/src/app/sampleBin/addItem.component.js
```diff
@@ -0,0 +1,37 @@
...
+    item: '<'
+  },
+  controller: ['SampleBin', '$log',
+    function(SampleBin, $log) {
+      var ctrl = this;
+      ctrl.sb = SampleBin;
+
+      ctrl.itemTooltip = function() {
+        if (!ctrl.item) {
+          return 'ititializing...';
```
rzelayafavila: Typo: `initializing` here as well

interface/src/app/sampleBin/addItem.component.js
```diff
@@ -0,0 +1,37 @@
...
+  templateUrl: 'sampleBin/addItem.component.tpl.html',
+  bindings: {
+    item: '<'
+  },
+  controller: ['SampleBin', '$log',
+    function(SampleBin, $log) {
+      var ctrl = this;
+      ctrl.sb = SampleBin;
+
+      ctrl.itemTooltip = function() {
```
rzelayafavila: This is a cool way to return tooltip messages!

interface/src/app/experiment/experimentDetail.view.js
```diff
@@ -0,0 +1,48 @@
...
+        controller: 'ExperimentDetailViewCtrl as ctrl'
+      }
+    },
+    data: {pageTitle: 'Experiment Detail'}
+  });
+}])
+
+.controller('ExperimentDetailViewCtrl', ['$stateParams', 'ApiBasePath',
+  function ExperimentDetailViewCtrl($stateParams, ApiBasePath) {
+    this.id = $stateParams.id;
```
rzelayafavila: Does it make sense to use the same `var ctrl = this;` syntax that we use below? I think it is nice to a) keep it consistent with the other code, and b) keep good track of what we are tacking on to the `ctrl` object (since, as we have discussed before, `this` can point to different things within a segment of code in JS).

interface/src/app/analyze/sample/sample.js
```diff
@@ -1,7 +1,7 @@
 angular.module('adage.analyze.sample', [
   'statusBar',
-  'adage.sample.services',
-  'adage.analyze.experiment'
+  'adage.sample.services',  // TODO: use adage.sample.service when refactored
```
rzelayafavila: :+1:

interface/src/app/analyze/search/searchDetail.js
```diff
@@ -1,12 +1,10 @@
 angular.module('adage.analyze.detail', [
-  'adage.analyze.sample',
-  'adage.analyze.experiment'
+  'adage.analyze.sample', // TODO: use adage.sample when refactored
```
rzelayafavila: :+1:

interface/src/index.html
mhuyck: Nice catch. I will make it consistent.

- [ ] fix indentation at `ng-class` in index.html

interface/src/app/sampleBin/addItemAnalyze.component.js
mhuyck: LOL. Got it... thanks!

interface/src/app/experiment/experimentDetail.view.js
mhuyck: SGTM. This might even be a bug because the stuff inside the `loaded` function is a callback so `this` is not well defined. Nice catch!

- [ ] need to use a closure variable to ensure we use the right `this` in `experimentDetail`

mhuyck:
Test instance is now up and running at http://greenelab.fgtech.com/

interface/src/app/analyze/search/searchDetail.js
mhuyck: This should be more clear from the altered workflow on the test server: http://greenelab.fgtech.com/ (also shown in the workflow design sketch at https://github.com/greenelab/adage-server/issues/268).

Previously, when a user clicked on a search result, the experiment details would slide in from the right to appear within the `analyze` view. This made it complicated to parameterize the state of that page and my observations of users on the system showed the split-screen view to be of dubious value anyway.

Now, when a user clicks on an experiment in the search results, the `ui-router` is engaged to switch state so the experiment details load in their own page. (For now, clicking a sample in the search results retains the original behavior, but this will also change when that component is refactored.)

mhuyck:
I think I have addressed all feedback now. @dongbohu & @rzelayafavila: please let me know if you catch anything else!

mhuyck:
Thank you!

Issue #282, opened 2017-10-04
Signatures of other models shown in "Edge Tips"  window


dongbohu:
When user clicks a certain edge, a list of signatures that are related to the two genes connected by that edge will be displayed. Right now signatures of **ALL** machine learning models will be displayed. See the attached screenshot:
![image](https://user-images.githubusercontent.com/15133253/31184176-aea5cdaa-a8f6-11e7-867b-943ba5ba6263.png)
There were two `Sig53pos` and `Sig76pos` signatures in `Edge Tips` window, because we had two machine learning models that happen to have identical signature names. 

It makes much more sense to show only the signatures of the same machine learning model. On the back end, an extra filter should be added to fix it.

dongbohu:
Fixed in #286.

PR #286, opened 2017-10-05
Fix of #282

dongbohu:
I found the issue #282 is important but easy to fix. Here it is.

I also make the timeout value of the slider rendering longer. The original value (`250 ms`) seems to work fine on my desktop, but not on the production server. `1000ms` should make it better.

### DIFF
--- a/interface/src/app/gene/network/network.js
+++ b/interface/src/app/gene/network/network.js
@@ -157,9 +157,9 @@ angular.module('adage.gene.network', [
       var refreshSlider = function() {
         $timeout(function() {
           $scope.$broadcast('rzSliderForceRender');
-        }, 250);
-        // Note by dhu: 250ms is an arbitrary timeout value. The minimum timeout
-        // value that worked on my desktop is 150ms.
+        }, 1000);
+        // Note by dhu: 1000ms is an arbitrary timeout value. The minimum
+        // timeout value that worked on my desktop is 150ms.
       };
       // Monitor "self.isValidModel". Once it is true, reconfigure a few slider
       // parameters and call refreshSlider().
@@ -327,7 +327,10 @@ angular.module('adage.gene.network', [
           var heavyGenes = [data.gene1.id, data.gene2.id].join(',');
           var target = d3.event.target;
           Signature.get(
-            {'heavy_genes': heavyGenes, 'limit': 0},
+            {'heavy_genes': heavyGenes,
+              'mlmodel': MlModelTracker.id,
+              'limit': 0
+            },
             function success(response) {
               var i = 0, n = response.objects.length;
               var anchorTag;


dongbohu:
Here is the demo page:
The signature names have also been padded in this demo server.

rzelayafavila:
@dongbohu do you have this running in a test instance to check out?

interface/src/app/gene/network/network.js
```diff
@@ -157,9 +157,9 @@ angular.module('adage.gene.network', [
       var refreshSlider = function() {
         $timeout(function() {
           $scope.$broadcast('rzSliderForceRender');
-        }, 250);
-        // Note by dhu: 250ms is an arbitrary timeout value. The minimum timeout
-        // value that worked on my desktop is 150ms.
+        }, 1000);
+        // Note by dhu: 1000ms is an arbitrary timeout value. The minimum
+        // timeout value that worked on my desktop is 150ms.
```
mhuyck: This fixes the slider glitch for me, too. Nice!

The fact that it works for a longer timeout but not for a shorter one is an indication that there is a race condition here. Some event must happen before issuing the `rzSliderForceRender` and it happens sometimes within 250ms and sometimes not. If we knew what that event was we would be able to make a 100% reliable fix. If we don't know what the event is, then this solution is a good enough workaround for me.

interface/src/app/gene/network/network.js
```diff
@@ -327,7 +327,10 @@ angular.module('adage.gene.network', [
           var heavyGenes = [data.gene1.id, data.gene2.id].join(',');
           var target = d3.event.target;
           Signature.get(
-            {'heavy_genes': heavyGenes, 'limit': 0},
+            {'heavy_genes': heavyGenes,
+              'mlmodel': MlModelTracker.id,
```
mhuyck: Nice and simple to fix. Great!

interface/src/app/gene/network/network.js
rzelayafavila: :+1:

interface/src/app/gene/network/network.js
rzelayafavila: :+1:

mhuyck:
Just had one other thought, but it should not hold up merging this fix: can you make a test case that would have caught this problem?

dongbohu:
I took a look at the back end unit test before submitting this PR. There is actually one test case for this `heavy_genes` filter. But to catch this bug, we really need a front end E2E test to simulate the model selecting and edge clicking behavior.

mhuyck:
That's true. Should we make an issue to follow up after we get E2E testing working? I think we will have a much easier time with E2E tests after we get Docker deployment finalized. (The issue I recall having with the E2E setup was specific to our AWS EC2 deployment.)

interface/src/app/gene/network/network.js
dongbohu: @mhuyck What you said about race condition is a good point. I took a look at the source code of `rzslider`. Unsurprisingly the event is related to a lot of DOM manipulations. I tried a few quick fixes but they didn't work. The fact that its README.md suggests using `$timeout()` probably indicates that that is the best solution.

I will add a new issue to keep track of it though.

dongbohu:
@mhuyck: I just talked to @casey about the importance of having E2E test on the front end too. I will add an issue for now.

PR #290, opened 2017-10-06
[FIX] update psycopg2 because old version won't install

mhuyck:
This fix allowed us to get v0.9.1b into production. It's not clear what changed between our release on Monday and last night, but I suspect something in a `psycopg2` dependency got updated and broke its install process.


mhuyck:
@dongbohu: I agree it's weird. The change that broke our deployment is not likely within `psycopg2` itself, but, I suspect, within one of its dependencies. Everything in our `requirements.txt` is fixed to a particular version, but that does not necessarily fix the dependencies.

The function of `psycopg2` is to act as a bridge between Django and PostgreSQL. Aha... check this out: https://www.postgresql.org/about/news/1786/

I'll bet the release of PostgreSQL 10.0 broke the older version of `psycopg2` we were using. This makes sense because our installation script (in `adage-deploy`'s `fabfile.py`, line 193) says:

```python
sudo('apt-get -y -q install postgresql-common libpq-dev postgresql-client')
```

That picks the latest `postgresql` stuff. I think this explains it.

rzelayafavila:
I think this makes sense, thanks for looking into it @mhuyck!

Issue #257, opened 2017-09-15
One slider that specifies minimum edge weight (instead of two sliders for min and max) 


dongbohu:
Slack discussion:

casey: 
don't think we need two sliders anymore - one should be fine?
since we have the pos/neg checkboxes

dhu:
Is it possible that a user may want to see the edges whose weights are in the range such as [0.5, 0.7]?

casey:
i don't know that this would happen unless a user really wants to figure out one very specific edge,
and for that, we might need to make their life easier in some other way.
i think a max genes slider might be the starting point,
plus that zoom/etc thing, so maybe if there are lots of genes our default zoom is further out so that the network can move around more freely, and/or a higher default edge threshold (while letting people scroll down to 0.2)

dongbohu:
Fixed in PR #291.

PR #291, opened 2017-10-09
Gene-Gene Network: Remove max slider from slider bar

dongbohu:
This is a fix for issue #257. I also tweaked the back end so that the signature names on edge tips window will be displayed in the alphabetic order of the signature names (which is consistent to the order on main page of signatures). 

Here is a demo page for this PR:

### DIFF
--- a/adage/analyze/api.py
+++ b/adage/analyze/api.py
@@ -281,6 +281,7 @@ class NodeResource(ModelResource):
             'heavy_genes': ('exact', ),  # New filter, see apply_filters().
             'mlmodel': ('exact', ),
         }
+        ordering = ['name']
 
     def prepend_urls(self):
         return [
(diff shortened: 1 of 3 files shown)


adage/analyze/api.py
```diff
@@ -281,6 +281,7 @@ class Meta:
             'heavy_genes': ('exact', ),  # New filter, see apply_filters().
             'mlmodel': ('exact', ),
         }
+        ordering = ['name']
```
mhuyck: 👍

interface/src/app/gene/network/network.js
```diff
@@ -329,6 +328,7 @@ angular.module('adage.gene.network', [
           Signature.get(
             {'heavy_genes': heavyGenes,
               'mlmodel': MlModelTracker.id,
+              'order_by': 'name',
```

interface/src/app/gene/network/network.js
```diff
@@ -130,21 +130,20 @@ angular.module('adage.gene.network', [
         } else {
           correlationSign = 1;
         }
-        network.filterEdgeWeight(self.slider.min, self.slider.max,
+        network.filterEdgeWeight(self.slider.min, maxCorrelation,
```

Issue #253, opened 2017-09-10
Bug in analysis-detail page when loading E-MEXP-1051 samples and trying to see the heatmap of them


rzelayafavila:
I got this error when playing around with the UI on the `greenelab.fgtech.com` instance. When I add the 4 samples that appear for experiment E-MEXP-1051(Transcription profiling of Pseudomonas aeruginosa PA14 grown in aerobic or anaerobic conditions in cystic fibrosis sputum with added nitrate ), and then try to see the heatmap, the heatmap does not load, and I see some errors in the JavaScript console.

rzelayafavila:
![analysis-detail-screenshot](https://user-images.githubusercontent.com/4641983/30252338-ad2e23a2-963e-11e7-935c-762ae993ca27.png)

mhuyck:
This looks very similar to #125, but the cause noted for that issue is not exactly the same here. In this case, we have an "ML Data Source" listed for each sample. This might still be a dup, but I'll have to dig deeper before I know for sure.

mhuyck:
More information: if I remove the two aerobic samples (aerobic_NO3_1, aerobic_NO3_2), the heatmap can be drawn again.

mhuyck:
There are similar issues with experiment `E-GEOD-26931`. Neither the heatmap nor a volcano plot will draw. In that case, if I remove the four samples with "strain" listed as "E.coli ZK126" then the heatmap recovers.

The pattern seems to be that activity data are missing for some samples. When those samples with missing data are removed then the heatmap works again.

mhuyck:
The pattern seems to be that activity data are missing for some samples. When those samples with missing data are removed then the heatmap works again.

Some of these cases seem to be data errors and some may be intentional missing data. In either case, the `adage-server` should be able to handle the missing data gracefully. Perhaps it should even include a message about what samples are affected.

mhuyck:
Fixed by #292.

Issue #285, opened 2017-10-04
Improve sample group selection process


mhuyck:
The sample group selection process in `analyze-detail` is currently very clumsy and requires many clicks to assign each sample individually to a group. We could alter this to use a more common "select noun (table rows)" then "perform verb (button)" interaction pattern.

Observed during user observation with the Hogan lab. Also: @jaclyn-taroni points out the group selection used by GEO2R:

> Have you checked out the GEO2R sample selection interface? See (this example)[https://www.ncbi.nlm.nih.gov/geo/geo2r/?acc=GSE12345]. It sounds similar to what they are looking for and I have always found it to be pretty intuitive and easy to use.

cgreene:
I think we need to know how much work this is to understand if this is a 1.0 or 1.1 issue. It may be related to #323

PR #292, opened 2017-10-12
[FIX] samples missing activity no longer break the heatmap

### LINKED ISSUE #125: heatmap bug
Heatmap disappears in some situation E-GEOD-4026

mhuyck:
These changes keep samples that are missing activity from being included in the
heatmap. Fixes #125, #253.


mhuyck:
I had initially planned to modify the Vega component to handle missing entries, but that would have made the Vega spec unnecessarily complicated. This solution, which prevents samples with missing activity data from getting added to the heatmap in the first place, is cleaner and allows us to more clearly explain to the user what is going on.

Will get a test server up and running shortly.

mhuyck:
Test server now running at http://greenelab.fgtech.com/ .

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -39,7 +39,7 @@
         <th>Sample Name</th>
         <th ng-repeat="at in analysis.exampleCols">{{
           at.typename
-          }}</th>
+        }}</th>
```
rzelayafavila: Hmmm, just curious - why did you decide to format it like this instead of:
```
        <th ng-repeat="at in analysis.exampleCols">
          {{at.typename}}
        </th>
```
It is a minor quibble, but I wonder if it would help with readability, as interpolation markup in angular is usually shown as `{{variable}}`.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -73,10 +73,62 @@
         <td>{{sb.sampleData[id].name}}</td>
         <td ng-repeat="at in analysis.exampleCols">{{
           sb.sampleData[id].annotations[at.typename]
-          }}</td>
+        }}</td>
```
rzelayafavila: Same question as in line 42 here.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -73,10 +73,62 @@
...
       </tr>
     </tbody>
   </table>
+  <div ng-if="sb.samples.length">
+    <h4>Samples Missing Activity</h4>
+    <p>The samples below are missing activity data, either because raw data
```
rzelayafavila: :+1:  I found this information to the user and the way it was laid out really helpful!

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -73,10 +73,62 @@
...
+        </tr>
+      </thead>
+      <tbody ng-model="sb.samples">
+        <tr ng-repeat="id in sb.samples"
+            ng-mouseover="showControls = true"
+            ng-mouseleave="showControls = false">
+          <td class="row-control">
+            <button type="button" class="btn row-control"
+                    ng-click="sb.removeSample(id)" aria-label="Delete">
+              <i class="fa fa-times" aria-hidden="true"></i>
```
rzelayafavila: Hmm, this `x` button seems too muted to me. I totally missed it the first time I looked at the table (even when the row is highlighted). The user needs to hover directly over it for it to be a bit more clear. Is this the same issue as with the table above, where it would be very CSS-involved to fix it?

interface/src/app/analyze/analysis/analysisDetail.tpl.html
```diff
@@ -73,10 +73,62 @@
...
+          <th>&nbsp;</th>
+          <th>
+            <span class="row-control">
+              <i class="fa fa-bars" aria-hidden="true"></i>
+            </span>
+            ML Data Source
+          </th>
+          <th>id</th>
+          <th>Sample Name</th>
+          <th ng-repeat="at in analysis.exampleCols">{{
```
rzelayafavila: Same question about formatting here as in `analysisDetail.tpl.html` file.

rzelayafavila:
Also, I was wondering about the `ML Data Source` cell appearing blank in the case below. Is this expected? Do we want to fill that cell with something like N/A or something else?
missing_activity_heatmap_screenshot

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: I also prefer `{{...}}`.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: Agree. Can we use a more obvious button here?

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: I am doing it this way to keep the resulting HTML markup as clean as possible. Let's say that `at.typename` resolves to "genotype." Then the way you suggest (which, I agree, makes the code more readable) will result in the following HTML:

```html
<th ng-repeat="...">
  genotype
<th>
```

This puts space between the markup tag and the contents, so the browser effectively reads this as:

```html
<th> genotype </th>
```

What I have written will keep the spaces out of the generated html to produce:

```html
<th>genotype</th>
```

This might not make much of a difference in the current layout, but consider what happens when you have:

```html
<a href="blahblahblah">
  {{something that resolves to "genotype"}}
</a>
```

The resulting HTML will be:

```html
<a href="blahblahblah"> genotype </a>
```

and this produces the dreaded underline styling that leaks around the text with the anchor. Something like

`_genotype_`

Performing CSS styling tweaks will be *much* easier with clean HTML.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: See above. I agree this is a little funky but should make life much easier for CSS styling (and produces more correct HTML formatting).

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: This is exactly the same issue as above. In fact, I used a modified version of the same table markup here. As part of (near) future refactoring work this will become a `sampleTable` component in which I am planning to eliminate the need for this column of controls, so I'd like to defer tweaking, if I may, until we get to #285.

Although I expect this related refactoring to come soon it is not on the list for our October release. I agree it is a usability issue, so if you feel it should happen sooner, let's raise the priority of #285 or create a separate issue to revisit the CSS headache that currently haunts those controls.

mhuyck:
The `ML Data Source` cell is blank in that table because no raw data are even available for that sample (last I checked). This is the case for many samples in general within ArrayExpress, but very few samples we selected for ADAGE because we can only include samples in our models for which we can get data, of course. There are a small number of samples without data (as shown by this bug) that do appear in our collection of annotations.

I guess this is a general question: how should we display missing data here? Should the table cell remain blank or should we include "N/A" or should we use a symbol like "–"? I am open to suggestions.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: This is a good point. (Actually in `<th>...</th>` the leading and trailing spaces between the two tags will be ignored by default.) But I have to confess that it still looks weird to me.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
dongbohu: I think this is important to user experience. Can we boot its priority?

interface/src/app/analyze/analysis/analysisDetail.tpl.html
rzelayafavila: Ah gotcha, ok, that makes sense to me as well.

dongbohu:
@mhuyck: About the missing `ML Data Source`, I prefer using N/A because it seems more user-friendly than the other options.

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: This is true. The *default* styling for a `<th>` tag will not reveal the space between the tag and its contents. BUT: what if we wanted the column headers to appear as links because they should be clickable? This is not an uncommon practice... many tables will allow a user to sort by column by clicking the header of that column. This feature can be added to a plain table like this without rewriting the HTML through CSS and JavaScript, but it would only look right if the HTML were "clean."

interface/src/app/analyze/analysis/analysisDetail.tpl.html
mhuyck: Which approach do you want to take, @dongbohu? Work on a temporary solution via CSS or start working on #285 sooner? I am fine with going either way, but I'm not certain I could complete #285 before the end of this month without delaying another issue on my list.

dongbohu: What about putting #285 in next month release?

mhuyck: Totally happy with that way of doing it. I'll have to take a closer look toward the end of the month. I may want to break that issue into smaller pieces before tackling it, but right now my gut says that if it were top priority it would fit into the next release.

mhuyck:
The only thing I think is unresolved now is how to display missing data in the table cells (like the `ML Data Source` example @rzelayafavila found). Any thoughts?

dongbohu:
@mhuyck: I prefer using "N/A" to display the missing data in table cells.

rzelayafavila:
@mhuyck: I agree, I think "N/A" sounds like a good way to go.

mhuyck:
Okay, will do on the n/a change. Thank you both!

mhuyck:
This apparently simple addition of a filter for showing "N/A" flushed out a hidden bug that was a bit tricky to solve. I've nailed it now, though. Will update the test server as soon as I can. There is a fair chunk of new code to make this work: (1) the filter -- implemented as a reusable component, and (2) the bug fix, which involved adding a method to the `sampleData` objects returned by `adage.sample.service`. Please have a look at the changes and let me know if you find anything that needs clarification.

mhuyck:
Test server updated.

interface/src/app/formatMissing.filter.js
```diff
@@ -0,0 +1,14 @@
+angular.module('adage.formatMissing.filter', [])
+
+.filter('formatMissing', function() {
```
rzelayafavila: It's great that we now have a uniform way to handle this! :+1:

interface/src/app/formatMissing.filter.js
```diff
@@ -0,0 +1,14 @@
...
+  return function(inputStr, replacementStr) {
+    // default to 'N/A'
+    var replacement = (replacementStr === undefined ? 'N/A' : replacementStr);
+    if (inputStr.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '') === '') {
```
dongbohu: Can you add some comments to explain what this line is doing? (I guess it is replacing some unprintable UTF-16 characters?)

interface/src/app/formatMissing.filter.js
```diff
@@ -0,0 +1,14 @@
+angular.module('adage.formatMissing.filter', [])
+
+.filter('formatMissing', function() {
+  return function(inputStr, replacementStr) {
+    // default to 'N/A'
+    var replacement = (replacementStr === undefined ? 'N/A' : replacementStr);
```
dongbohu: I got a little confused about the two variables "replacement` and `replacementStr`. Can you rename one of them?

interface/src/app/sample.service.js
```diff
@@ -11,6 +11,35 @@ angular.module('adage.sample.service', [
...
+              }
+            };
+            if (status === 200) {
+              data = angular.fromJson(data);
+              if (data.hasOwnProperty('objects')) {
+                // data has a pagination wrapper, so must iterate over list
+                data.objects.forEach(function(obj) {
+                  obj.getAnnotation = getAnnotation;
+                });
+              } else {
```
dongbohu: In `GET` method, doesn't `data` always have `objects` property (even when there is only one record) unless `isArray` is set to `true`?
https://docs.angularjs.org/api/ngResource/service/$resource#$resource-returns

interface/src/app/formatMissing.filter.js
mhuyck: Good point. Will do that.

interface/src/app/formatMissing.filter.js
mhuyck: Heh. Sorry about that. It is catching not just plain ASCII whitespace but a couple of other cases. I'll see if I can find the code where I got this example and credit the source.

- [ ] add a comment for this gnarly regex

interface/src/app/sample.service.js
mhuyck: Great question. The answer turns out to be no, and I included this logic to handle two separate cases that were covered by our existing tests so they would keep working.

There are at least two ways to get a response without an `objects` property (like the one from setting `isArray: true` in the `$resource` definition). Using `isArray` is one, of course. The other way is to make a direct request for a REST "detail" endpoint rather than a "list" endpoint. In other words, we can GET a specific object rather than a list.

The difference is detailed well in these examples (see the sample responses) from the Tastypie documentation:

> **list endpoint**: http://django-tastypie.readthedocs.io/en/latest/interacting.html?highlight=detail#getting-a-collection-of-resources

> **detail endpoint**: http://django-tastypie.readthedocs.io/en/latest/interacting.html?highlight=detail#getting-a-detail-resource

interface/src/app/formatMissing.filter.js
mhuyck: This is actually from the Mozilla documentation for `String.prototype.trim`, found at:

> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim#Polyfill

The compatibility table at the bottom of that page indicates that it "should" be safe to just use `.trim()` in all modern browsers, but I'm being conservative since `.trim()` first appeared in the standard for ES 5.1 and `.replace()` is compatible back to the 3rd edition standard.

I've added a comment to explain the regex. If you think I'm being too conservative, let me know and I'll just use `.trim()`.

dongbohu: Thanks. I prefer to be conservative too.

interface/src/app/sample.service.js
dongbohu: Got it. Thanks.

Issue #256, opened 2017-09-15
New slider bar: Max Genes


dongbohu:
Based on the slack discussion with @cgreene:
```
i think we may want a "max genes" slider

default = 50 or something?

worth noting - new d3.network from aaron used in this (not that i think we need it, 
but it has some new controls):
http://hb.flatironinstitute.org/gene/3553
they do allow a limit on max # of genes
```
Aaron Wang's web UI: http://hb.flatironinstitute.org/gene/3553

In this demo:
http://165.123.67.202/#/gene_network/?genes=6552&mlmodel=5
when the edge cutoff is 0.2, the number of genes that are directly connected to gene `6552` is 511.

PR #293, opened 2017-10-23
Gene-Gene Network: new Slider that controls the max # of genes displayed

dongbohu:
This is a relatively large PR that addresses issue #256. 

A debug line is also removed from `signature.js`.

Instead of being loaded from github, `d3.network.js` is now added in the repo to avoid the conflicts with production server. I need to talk to Aaron Wang to figure out the best way to solve this problem. You don't have to review this file in the PR. I also changed codeclimate config so that this file won't be checked.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,6 +4,7 @@ engines:
     enabled: true
     exclude_paths:
     - "interface/Gruntfile.js"
+    - "interface/d3.network.js"
     config:
       languages:
         python:
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -18,6 +18,7 @@ engines:
     exclude_paths:
     - "interface/Gruntfile.js"
     - "interface/eslint.google.js"
+    - "interface/d3.network.js"
   fixme:
     enabled: false
   radon:
@@ -35,7 +36,6 @@ engines:
     - "interface/CHANGELOG.md"
     - "interface/CONTRIBUTING.md"
     - "interface/README.md"
-    - "interface/d3.network.js"
     - "interface/tools.md"
     - "interface/src/README.md"
     - "interface/src/app/README.md"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -35,6 +35,7 @@ engines:
     - "interface/CHANGELOG.md"
     - "interface/CONTRIBUTING.md"
     - "interface/README.md"
+    - "interface/d3.network.js"
     - "interface/tools.md"
     - "interface/src/README.md"
     - "interface/src/app/README.md"
--- a/interface/build.config.js
+++ b/interface/build.config.js
@@ -78,6 +78,7 @@ module.exports = {
       'vendor/ng-sortable/dist/ng-sortable.min.js',
       'vendor/angularjs-slider/dist/rzslider.min.js',
       'vendor/d3-tip/index.js',
+      'd3.network.js',
       'node_modules/hclusterjs/hcluster.js',
 
       // Note: greenelab.stats.ttest.js and greenelab.stats.multtest.js are
--- a/interface/d3.network.js
+++ b/interface/d3.network.js
@@ -1,0 +1,679 @@
+/* Reusable d3 force-directed layout for gene networks
+ * Follows the design pattern suggested by Mike Bostock:
+ * http://bost.ocks.org/mike/chart/
+ */
+
+(function() {
+
+d3.network = function() {
+
+    var options =
+        {start_color:"#006E2E",
+        mid_color:"#FFFF88",
+        end_color:"#FF0000"};
+
+    var force,
+        edges = [],
+        genes = [],
+        gene_map = {},
+        draw_genes = [],
+        draw_edges = [],
+        height,
+        width,
+        selection,
+        edgeLegendEnabled = false,
+        edge_events = {},
+        gene_events = {},
+        bind_networks = [],
+        edgeLegendText = 'Relationship confidence',
+        min_edge = 0,
+        max_edge = 1,
+        edgeLegendStart = min_edge,
+        edgeLegendEnd = max_edge,
+        // dhu: variables related to gene color legend bar
+        geneColorEnabled = false,
+        geneLegendText = 'Gene Expression Value',
+        min_gene_expr = -1,
+        max_gene_expr = 1,
+        geneLegendStart = min_gene_expr,
+        geneLegendEnd = max_gene_expr,
+        // dhu: flag that indicates whether the background correction
+        // should be applied when the genes are ranked. Default is true.
+        bgCorrection = true,
+
+        event = d3.dispatch("edgeadd", "edgeremove",
+                "geneadd", "generemove", "genechange");
+
+    // Functions for network attributes
+    var r = function(d) {return d.query ? 20 : Math.max(10,5+ d.query_degree ? d.query_degree*10 : 10);};
+    var w = function(d) {return Math.max(2, Math.abs(d.weight)*6);};
+    var edgeColor = d3.scale.linear().domain([options.start_edge,.15,1])
+        .range([options.start_color, options.mid_color, options.end_color]);
+
+    // dhu: gene circle's color
+    var geneColor = d3.scale.linear().domain([-1, .15, 1])
+        .range([options.start_color, options.mid_color, options.end_color]);
+
+    var geneText = function(d) { return d.standard_name; };
+
+    function my(_selection) {
+        selection = _selection;
+
+        if ( !width ) width = selection.attr("width");
+        if ( !height) height = selection.attr("height");
+
+        var i,
+            n = genes.length, m = edges.length,
+            gene, edge;
+
+        // Bind nodes to edges
+        for (i = 0; i < m; ++i) {
+          edge = edges[i];
+          if (typeof edge.source == "number") edge.source = genes[edge.source];
+          if (typeof edge.target == "number") edge.target = genes[edge.target];
+        }
+
+        // Calculate node degrees to query genes and all genes
+        for( i = 0; i < n; i++ ) {
+            gene = genes[i];
+            gene.query_degree = 0;
+            gene.query_degreen = 0;
+            gene.degree = 0;
+            gene.degreen = 0;
+        }
+        for( i = 0; i < m; i++ ) {
+            edge = edges[i];
+            if ( edge.source.query ) {
+                edge.target.query_degree += Math.abs(edge.weight);
+                edge.target.query_degreen ++;
+            }
+            if ( edge.target.query ) {
+                edge.source.query_degree += Math.abs(edge.weight);
+                edge.source.query_degreen ++;
+            }
+            edge.target.degree += Math.abs(edge.weight);
+            edge.source.degree += Math.abs(edge.weight);
+            edge.target.degreen ++;
+            edge.source.degreen ++;
+        }
+        for( i = 0; i < n; i++ ) {
+            genes[i].query_degree /= genes[i].query_degreen;
+            genes[i].degree /= genes[i].degreen;
+        }
+
+        // Sort nodes according to background corrected degree
+        // Store corresponding ranks
+        if (bgCorrection) {
+          genes.sort(function(a,b){return b.query_degree/b.degree
+                                   -a.query_degree/a.degree;});
+        } else { // option in greenelab adage-server
+          genes.sort(function(a,b){return b.query_degree-a.query_degree;});
+        }
+        for( i = 0; i < n; i++ ) { genes[i].rank = i; }
+    }
+
+    function draw(genes, edges) {
+        if ( typeof force != "undefined" ) force.stop();
+
+        force = d3.layout.force()
+            .charge(-800)
+            .distance(150)
+            .gravity(.15)
+            .nodes(genes)
+            .links(edges)
+            .size([width, height]).start();
+
+        var node_data = selection.selectAll("g.gene-group")
+                .data(genes, function(d) { return d.id; });
+        var link_data = selection.selectAll("line.edge")
+                .data(edges, function(d) { return d.id; });
+
+        var link_rm = link_data.exit().remove();
+        var link = link_data
+                .enter().insert("svg:line", "g")
+                .attr("class", "edge")
+                .on("mouseover", linkMouseover)
+                .on("mouseout", linkMouseout)
+                .attr("id", function(d) { return d.id; })
+                .style("stroke", function(d) {return edgeColor(d.weight);})
+                .style("stroke-width", w);
+        for( var type in edge_events ) link.on(type, edge_events[type]);
+        event.edgeadd(link.filter(function(d){return d;}));
+        event.edgeremove(link_rm.filter(function(d){return d;}));
+
+        var node_rm = node_data.exit().remove();
+        var node = node_data // dhu: customize node color here!
+                .enter().append("svg:g")
+                .attr("class", "gene-group")
+                .on("mouseover", nodeMouseover)
+                .on("mouseout", nodeMouseout)
+                .on("click", nodeMouseclick)
+                .on("mousedown", nodeMousedown)
+                .attr("id", function(d) { return d.id; });
+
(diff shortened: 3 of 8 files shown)


dongbohu:
Here is a demo page:
http://165.123.67.202/#/gene_network/?genes=6554&mlmodel=4

interface/src/app/gene/network/network.js
```diff
@@ -149,31 +151,18 @@ angular.module('adage.gene.network', [
...
-        function() {
-          if (self.isValidModel) {
-            self.slider.min = MlModelTracker.g2gEdgeCutoff;
-            self.slider.options.floor = MlModelTracker.g2gEdgeCutoff;
-            refreshSlider();
+      self.maxGeneNumSlider = {  // slider that controls max number of genes
+        value: 0,                // initial position of slider
+        options: {
+          floor: 1,              // minimum of the slider bar
+          ceil: 100,             // maximum of the slider bar
```
mhuyck: What are the units on these numbers? We are not limiting to 100 genes max, right? Is this a percentage?

interface/src/app/gene/network/network.tpl.html
```diff
@@ -14,25 +14,37 @@
...
-    <rzslider rz-slider-model="ctrl.slider.min"
-              rz-slider-options="ctrl.slider.options">
+    <rzslider id="edge-slider"
+              rz-slider-model="ctrl.minEdgeWeightSlider.value"
+              rz-slider-options="ctrl.minEdgeWeightSlider.options">
     </rzslider>
+    </div>
+
+    <div class="well well-sm">
+      <strong>Max # of genes:</strong>
```
mhuyck: I think it would be helpful to include some information for the user about *how* the genes are filtered. We are using some kind of rank, right? It is not clear from the UI how changing this slider alters which genes are shown and which are hidden.

interface/src/app/gene/network/network.less
```diff
@@ -54,3 +54,11 @@ text.legend-title {
...
   width: 600px;  // Default: 100%
 }
+
+#gene-slider .rz-pointer {
+  background-color: magenta;
+}
+
+#gene-slider .rz-selection {
+  background: magenta;
+}
```
mhuyck: I would not have chosen `magenta` here. Maybe we can have Deepa figure out a standardized color palate for us to work with?

interface/src/app/signature/signature.js
```diff
@@ -42,7 +42,6 @@ angular.module('adage.signature', [
         self.name = response.name;
         MlModelTracker.set(response.mlmodel);
         self.modelID = MlModelTracker.id;
-        console.log('self.modelID: ' + self.modelID);
```
mhuyck: Nice catch!

interface/src/index.html
```diff
@@ -26,9 +26,10 @@
     <!-- d3.network library, written by Aaron Wang (aaronkw)
       FIXME: Ideally d3.network.js should be installed in vendor/ subdirectory,
       but since it is under active development, it is requested online for now.
-      -->
+
     <script src="https://rawgit.com/dongbohu/d3-network/master/d3.network.js"
             charset="utf-8"></script>
+    -->
```
mhuyck: This specific `FIXME` is addressed by adding `d3.network.js` to our repository, so I'd recommend removing lines 26-32 entirely.

If you want to have the reminder to follow up on managing the workflow for incorporating our "fork" of Aaron's code as we develop, it should probably be tracked as an issue now. I have a few thoughts I could add as comments to the issue.

interface/d3.network.js
```diff
@@ -0,0 +1,679 @@
+/* Reusable d3 force-directed layout for gene networks
+ * Follows the design pattern suggested by Mike Bostock:
+ * http://bost.ocks.org/mike/chart/
+ */
```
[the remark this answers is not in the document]
dongbohu: Ideally the `d3.network` library should be modified in my fork and reviewed by Aaron before it is merged to his master repo. But when we started to add more and more customization to our gene-gene network, it doesn't seem to be practical to ask Aaron to review every change. So I decided to use my own fork in `index.html` a few months ago.

But the issue that Deb had yesterday made me realize that it was not a good long-term solution, because:
(1) Whatever change I made to `d3.network.js` should be reviewed by somebody else.
(2) It makes more sense to bundle `d3.network.js` into our release instead of loading it from github. That way any previous version of adage-server doesn't have to be dependent on the one on github.

That is why I am including `d3.network.js` in this PR. Eventually I would like to change its coding style to the one we use and file another PR in the future. Then our development of this library will be independent from Aaron's repo.

interface/d3.network.js
```diff
@@ -0,0 +1,679 @@
...
+            }
+
+            var cmpVal;
+            if (!weightSign) {
+                cmpVal = Math.abs(d.weight);
+            } else if (weightSign < 0) {
+                cmpVal = -d.weight;
+            } else if (weightSign > 0) {
+                cmpVal = d.weight;
+            } else {  // in case cmpVal is not even a valid number
```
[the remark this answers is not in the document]
dongbohu: Good catch! Thanks.

interface/d3.network.js
```diff
@@ -0,0 +1,679 @@
...
+
+    my.onEdge = function(type, listener) {
+        if ( typeof listener == "function" ) edge_events[type] = listener;
+        return my;
+    };
+
+    my.edgeColor = function(x) {
+        if (!arguments.length) return edgeColor;
+        if ( typeof x == "function" ) edgeColor = x;
+        else edgeColor = function(d) { return x; };
```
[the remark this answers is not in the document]
dongbohu: No, this is the trick that has been used very often in this library. It means if `edgeColor()` is invoked like this:
```network.edgeColor("red");```
then it resets the internal `edgeColor` variable into such a function:
```
edgeColor = function(d) {
    return "red";
}
```
Technically speaking, the parameter `d` is unnecessary, because it is not being used. But since it was written by Aaron, I never bother.

File: interface/d3.network.js [hunk 9]
```diff
@@ -0,0 +1,679 @@
...
+    my.maxEdge = function(x) {
+        if (!arguments.length) return max_edge;
+        max_edge = x;
+        return my;
+    };
+
+    my.edgeWidth = function(x) {
+        if (!arguments.length) return w;
+        if ( typeof x == "function" ) w = x;
+        else w = function(d) { return x; };
```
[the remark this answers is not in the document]
dongbohu: See my explanation above.

interface/d3.network.js
```diff
@@ -0,0 +1,679 @@
...
+    };
+    // dhu: new function to set gene circle's color
+    my.geneColor = function(x) {
+        if (!arguments.length) {
+            return geneColor;
+        }
+        if (typeof x == "function") {
+            geneColor = x;
+        } else {
+            geneColor = function(d) { return x; };
```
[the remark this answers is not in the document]
dongbohu: See my explanation above.

interface/d3.network.js
```diff
@@ -0,0 +1,679 @@
...
+    my.maxGeneExpr = function(x) {
+        if (!arguments.length) return max_gene_expr;
+        max_gene_expr = x;
+        return my;
+    };
+
+    my.geneRadius = function(x) {
+        if (!arguments.length) return r;
+        if ( typeof x == "function" ) r = x;
+        else r = function(d) { return x; };
```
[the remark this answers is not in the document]
dongbohu: See my explanation above.

interface/d3.network.js
```diff
@@ -0,0 +1,679 @@
...
+        return my;
+    };
+
+    my.geneText = function(x) {
+        if (!arguments.length) return geneText;
+        if ( typeof x == "function" ) geneText = x;
+        else geneText = function(d) { return x; };
```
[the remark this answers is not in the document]
dongbohu: See my explanation above.

interface/src/app/gene/network/network.js
```diff
@@ -113,6 +113,7 @@ angular.module('adage.gene.network', [
           .geneText(function(d) {
             return d.label;
           })
+          .bgCorrection(false)    // disable background correction
```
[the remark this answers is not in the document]
dongbohu: I forgot to mention that the `bgCorrection` variable was introduced into d3.network.js after I discussed the algorithm that Aaron used to rank the genes (to decide which genes should be displayed in the network with slider that controls the max # of genes). Here is how it was implemented in `d3.network.js`:

* On line 42 of `d3.network.js`, you can see that `bgCorrection` is set to `true` by default.;

* Lines 493-499 defines an API function that allows user to reset `bgCorrection` variable;

* Lines 106-111 are using two different algorithms to rank the genes based on whether `bgCorrection` is set or not.

That is why in adage-server, we are calling `bgCorrection(false)` to set this flag to `false` explicitly. Does this make sense?

interface/src/app/gene/network/network.js
```diff
@@ -130,16 +131,17 @@ angular.module('adage.gene.network', [
         } else {
           correlationSign = 1;
         }
-        network.filterEdgeWeight(self.slider.min, maxCorrelation,
-                                 correlationSign);
+        network.filterWithWeightSign(
+          self.minEdgeWeightSlider.value, maxCorrelation,
+          self.maxGeneNumSlider.value, correlationSign);
```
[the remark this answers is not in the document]
dongbohu: I used this name mainly because Aaron had one function just called `filter` in line 261 of `d3.network.js`. The only difference between `filter` and this one is that the former doesn't consider the sign, but the latter does.

interface/src/app/gene/network/network.js
dongbohu: "1" and "100" are just plain numbers without unit. `100` is kind of like a placeholder here. On line 254, it is reset to the actual number of genes in the network.

.codeclimate.yml
```diff
@@ -4,6 +4,7 @@ engines:
     enabled: true
     exclude_paths:
     - "interface/Gruntfile.js"
+    - "interface/d3.network.js"
```
rzelayafavila: :+1:

.codeclimate.yml
```diff
@@ -18,6 +19,7 @@ engines:
     exclude_paths:
     - "interface/Gruntfile.js"
     - "interface/eslint.google.js"
+    - "interface/d3.network.js"
```
rzelayafavila: :+1:

interface/src/app/gene/network/network.less
dongbohu: I played with a few colors and eventually selected "magenta" because I want to use a color that is:
(1) different from "blue" (which is used by the first slider);
(2) visually obvious;
(3) not included in the legend bar.

I will talk to Deepa.

interface/build.config.js
```diff
@@ -78,6 +78,7 @@ module.exports = {
       'vendor/ng-sortable/dist/ng-sortable.min.js',
       'vendor/angularjs-slider/dist/rzslider.min.js',
       'vendor/d3-tip/index.js',
+      'd3.network.js',
```
rzelayafavila: :+1: It's great that we are now including this in the code for adage-server. However, it does mean that we need to manually update the `interface/d3.network.js` file everytime we make a change to the code in the repository. Ideally this would be an npm installable package or something of the sort. We might need to make an issue for this - what do you think?

interface/src/app/gene/network/network.tpl.html
dongbohu: It is not easy to explain the exact algorithm that ranks the genes. I had to walk through the code in `d3.network.js` with Casey to fully understand it. I think this kind of information should be added in `README.md` or the tech help page.

interface/src/index.html
dongbohu: I didn't remove it because I am not sure whether it is a good idea to include ``d3.network.js` in our repo. I will remove it if both of you agree.

interface/src/app/gene/network/network.js
```diff
@@ -130,16 +131,17 @@ angular.module('adage.gene.network', [
...
         network.draw();
       };
 
-      self.slider = {  // range slider configuration
-        min: 0,                  // initial position of slider on the left
+      self.minEdgeWeightSlider = {  // slider that controls min edge weight
```

interface/src/app/gene/network/network.js
```diff
@@ -255,6 +244,22 @@ angular.module('adage.gene.network', [
         // any more.
       };
 
+      // Function that renders the sliders:
```
rzelayafavila: This is a very minor comment, but I think this comment line is actually not necessary since the `renderSliders` function name is already descriptive enough! I think this comment just repeats that.

interface/src/app/gene/network/network.tpl.html
rzelayafavila: I was wondering about this as well. I assumed the same thing, but it might be good to clarify to the users.

interface/src/index.html
rzelayafavila: I agree, I made a similar comment about this.

interface/build.config.js
dongbohu: I think we can create a `d3.network` repo in GreeneLab, and I will fork it for code review, and use npm to install the one in GreeneLab. If this is okay with you guys, I will create an issue.

interface/src/app/gene/network/network.js
dongbohu: Sure. I will do that.

dongbohu:
I revised the files based on your feedback. Please let me know what you think. Thanks.

interface/build.config.js
rzelayafavila: :+1:

interface/src/app/gene/network/network.tpl.html
rzelayafavila: I agree, if it's not that trivial then clarifying in the README might make more sense

interface/d3.network.js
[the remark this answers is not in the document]
mhuyck: Okay, gotcha. This looked funky at first, but when I started seeing it lots of places I started to realize it must have been intentional.

interface/src/app/gene/network/network.js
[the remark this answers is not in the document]
mhuyck: Well, I followed the code, but I don't understand why we have to alter the ranking algorithm. Could you include a comment giving the rough idea behind each algorithm (enough to understand why we might use one vs. the other) at lines 106-111? I think the comment at 104-105 needs to be updated and expanded for each implementation because it doesn't apply to both cases.

interface/src/app/gene/network/network.less
mhuyck: This could be an issue for follow-up. Others may disagree with me about the use of magenta.

interface/src/app/gene/network/network.tpl.html
mhuyck: This ties in with the ranking code where I'm asking for more detail. I think we need a clear picture of how this works. If we cannot explain it, we do not understand it ourselves and I think that is a problem.

interface/src/app/gene/network/network.js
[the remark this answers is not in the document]
dongbohu: The ranking algorithms are lines 66-112. I don't think I can explain better than that. It basically considers both the edge weight and number of edges when calculating the rank. The key difference between the two algorithms is on lines 107 and 110. 

The comments on lines 104-105 were written by Aaron. I prefer modifying them later, not in this PR.

I have created an issue to keep track it. Need to figure out the best way to explain it to a regular user.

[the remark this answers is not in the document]
mhuyck: Tracking as an issue works for me. I think it will be confusing, and right now I could not explain why it does what it does what it does if a user asked.

interface/src/app/gene/network/network.js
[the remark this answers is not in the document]
mhuyck: If the naming (and parameter order) makes sense in context then it's fine with me. As you said, we aren't really reviewing the `d3.network.js` file here and I think we have a good plan for getting it reviewed moving forward.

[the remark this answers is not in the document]
mhuyck: Tagging issue #295 here to link it to this set of comments.

PR #296, opened 2017-10-25
Fix snake case in search service and related refactoring

### LINKED ISSUE #272: convert snake_case identifiers to camelCase
Codeclimate noted several places in `experimentDetail.view.js` (during review of https://github.com/greenelab/adage-server/pull/269) where we use non-standard identifiers (`item_type`, `related_items` and `related_items`). These should be updated throughout the interface code to follow camelCase convention.

This needs to be handled as a separate issue because these identifiers (and others) are in use in many places, including in the back-end at `api.py`. The JSON response to a `search` produces a list of objects with the attributes above, which is why this Python identifier standard has "leaked" into the interface.

This should be addressed by ensuring that our JSON identifiers adhere to the camelCase standard wherever possible.

mhuyck:
Fixes #272


interface/src/app/search.service.js
```diff
@@ -0,0 +1,42 @@
+angular.module('adage.search.service', [
+  'ngResource'
+])
+
+.factory('Search', ['$resource', function($resource) {
+  return $resource(
+    '/api/v0/search/',
```
rzelayafavila: Question - should this use the 'ApiBasePath' constant in app/utils.js for standardization?

interface/src/app/search.service.js
```diff
@@ -0,0 +1,42 @@
...
+    // then, can change "limit" below to something sensible
+    {q: '', limit: 0},
+    // Angular expects a query service to give only a list but our Tastypie
+    // interface wraps the response list with pagination so we tell Angular to
+    // expect an object instead via isArray: false
+    {
+      'query': {
+        method: 'GET',
+        isArray: false,
+        transformResponse: function(data, headersGetter, status) {
```
rzelayafavila: Aha, I see that in this `transformResponse` function is where the magic happens!

interface/src/app/search.service.js
```diff
@@ -0,0 +1,42 @@
...
+              delete obj[oldProp];
+            }
+          };
+          if (status === 200) {
+            data = angular.fromJson(data);
+            if (data.hasOwnProperty('objects')) {
+              data.objects.forEach(function(obj) {
+                renameProperty(obj, 'item_type', 'itemType');
+                renameProperty(obj, 'related_items', 'relatedItems');
+                renameProperty(obj, 'resource_uri', 'resourceUri');
```
rzelayafavila: :+1:

interface/src/app/search.service.js
mhuyck: Ooh, nice catch. I'll fix that.

- [ ] use `ApiBasePath` to define REST endpoint

interface/src/app/search.service.js
mhuyck: I thought about making this *more* magical, by writing a function that automatically converts from, e.g. `an_identifier` to `anIdentifier`, because it wouldn't be hard to write. In the end I decided against that because I think it would be harder to *read* code like that.

PR #297, opened 2017-10-30
[TWEAK] bump version number for October 2017 release

mhuyck:


interface/bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "adage",
-  "version": "0.9.1b",
+  "version": "0.10.0b",
```
rzelayafavila: Any reason for going with `0.10.0b` instead of `0.9.2b`? I'm just curious. It is nice that it coincides with the month number of the release!

mhuyck: I thought about a bug fix increment (to `0.9.2b`) but I felt that @dongbohu's changes to the Gene Network were more significant than that since the node filter is actually a new feature. It is still a minor update, though. Following semantic versioning practice, that meant bumping the "minor revision" field.

mhuyck: I just looked at the semantic versioning spec here:

> http://semver.org/spec/v2.0.0.html

According to this, it appears to be preferable to use a format like this to denote our "beta" status:

`0.10.0-beta`

Should we go with that to comply with the standard? It's obviously a very small change from what we have now.

dongbohu: @mhuyck: Thank you for the research. I think having `beta` explicitly in the version number is probably better to the users too. Since it is a minor change, let's do it (unless @rzelayafavila has a strong preference against it).

rzelayafavila: Ah good call. Ok, sounds good!

Issue #298, opened 2017-11-01
Gene-Gene Network: Change options to radio buttons


dongbohu:
This is a feedback from Deepa. It makes more sense to use three radio button instead of the two `Include negative weight` and `Include positive weight` option boxes.

dongbohu:
Fixed in PR #315.

Issue #305, opened 2017-11-17
Gene-Gene Network Page: retain the network state when "back" button is clicked


dongbohu:
**Deepa feedback Priority 3**
Once the network is generated, the Signature details open in the tab. When the users hit the back button, the network is generated again and the position of the nodes is back to default positions. This can be a pain-point, especially if there are a lot of nodes and the user has invested time to move them around to positions which make sense for them. Also, if they are interested in comparing or viewing multiple signatures, they will have to keep moving back and forth between the two pages, which is just a lot of work and waiting. 

Recommendation: Open the signature details page in a new tab.

Dr. Deb Hogan also mentioned that she would like the network to be retained when clicking "back" button.

mhuyck:
Consider adding parameters to capture the state of the "min edge weight" radio buttons and "max # of genes" slider so that link sharing preserves more of the state.

I guess these both will be needed if we are to try and preserve the positions of the nodes, but I thought I would explicitly call attention to that since it came up while discussing #315.

mhuyck:
Another thought, while I am here: capturing the positions of all of the nodes could turn into a lot of data very fast. It is not clear to me that this could be done *solely* via parameters in the URL.

One option is to try and get clever with encoding so that node positions take up as few bits as possible and then convert that binary data into some efficient URL-encoded parameter (base64?) but I would guess that even that could rapidly hit [practical limits on URL length](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers).

Another approach is to use a link-shortening technique (think of the ubiquitous https://bitly.com), the essence of which is to store a mapping between the short URL's ID and the full URL in a database. In a similar fashion, we could save the graph's state as a record in a table designed to support this feature via a bitly-style ID. Some care would be needed to avoid letting that table grow out of hand while indefinitely preserving graphs that people want to share.

There might be other ways to go about this, but I thought I'd note these here since they came to mind.

cgreene:
Let's make the link open in a new tab for 1.0.

dongbohu:
Fixed in PR #342.

PR #315, opened 2017-11-27
Network Web UI: Change checkboxes into radio buttons

dongbohu:
This is a fix of issue #298 (thanks to the feedback from Deepa).

### DIFF
--- a/interface/src/app/gene/network/network.tpl.html
+++ b/interface/src/app/gene/network/network.tpl.html
@@ -17,19 +17,17 @@
     <div class="well well-sm">
     <form>
       <strong>Min edge weight:</strong>
-      <label>
-        <input type="checkbox" ng-model="ctrl.includeNegative"
-               ng-disabled="!ctrl.includePositive"
-               ng-change="ctrl.renderNetwork()">
-        Include negative weight
-      </label>
+      <input type="radio" name="edgeSign" ng-model="ctrl.edgeSign"
+             ng-change="ctrl.renderNetwork()"
+             value="negOnly">Negative only
       &nbsp;
-      <label>
-        <input type="checkbox" ng-model="ctrl.includePositive"
-               ng-disabled="!ctrl.includeNegative"
-               ng-change="ctrl.renderNetwork()">
-        Include positive weight
-      </label>
+      <input type="radio" name="edgeSign" ng-model="ctrl.edgeSign"
+             ng-change="ctrl.renderNetwork()"
+             value="posOnly">Positive only
+      &nbsp;
+      <input type="radio" name="edgeSign" ng-model="ctrl.edgeSign"
+             ng-change="ctrl.renderNetwork()"
+             value="both">Both
     </form>
     <rzslider id="edge-slider"
               rz-slider-model="ctrl.minEdgeWeightSlider.value"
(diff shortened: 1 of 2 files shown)


mhuyck:
This looks like a nice, straightforward change from a code perspective. Do you have a demo site available so I can see what it looks like?

dongbohu:
@mhuyck: good question. I forgot to give you guys the demo's URL:

mhuyck:
Nice. Thank you!

interface/src/app/gene/network/network.tpl.html
```diff
@@ -17,19 +17,17 @@ <h3 class="text-warning" ng-bind="ctrl.statusMessage"></h3>
...
       <strong>Min edge weight:</strong>
-      <label>
-        <input type="checkbox" ng-model="ctrl.includeNegative"
-               ng-disabled="!ctrl.includePositive"
-               ng-change="ctrl.renderNetwork()">
-        Include negative weight
-      </label>
+      <input type="radio" name="edgeSign" ng-model="ctrl.edgeSign"
+             ng-change="ctrl.renderNetwork()"
+             value="negOnly">Negative only
```
mhuyck: I noticed on the demo page that there is no space between the radio button and its label and they appear to "collide" with each other on the page. The code here is a little different from what is shown in the AngularJS documentation:

> https://code.angularjs.org/1.5.11/docs/api/ng/input/input[radio]

In the Angular sample code there is a newline between the `<input>` tag and the label. There is also a `<label>` tag wrapping the `<input>`. Probably a space or newline would be sufficient here but if that doesn't help, try using the `<label>` wrapper as well (as you had before with the checkboxes).

mhuyck:
It's tricky to find edges with negative weight, but I found this:

> http://165.123.67.202/#/gene_network/?genes=6549&mlmodel=4

It looks as if you'll need to select the "Negative only" radio button and drag the "Max #" slider all the way to the right. These controls are not parameterized, it appears. Should they be?

dongbohu:
@mhuyck: Thank you for your comments. Right now the states of slider bars and radio buttons are not parametric. #305 is related to your comments.

I will add `<label>` and see how it looks like tomorrow morning.

dongbohu:
@mhuyck: I tried the `<label>` tag and found that it only makes the text of radio buttons into bold font, but it doesn't really add any extra space. I added one more space manually in the last commit.

dongbohu:
Please let me know whether this is good enough.

mhuyck:
LGTM! Labels have just the right amount of space now.

dongbohu:
@rzelayafavila: This is a good question. I never thought of this scenario when adding the `Max # of genes` slider. I will take a look at how Aaron Wang's web UI handles this situation. What you said does make a lot of sense: when `Negative only` is selected, `Max # of genes` slider probably should consider only negative edges, so when those 13 negative edges should show up when the `Max # of genes slider is set to 13 or larger.

If we agree this is the right direction to go, I will create an issue and have it implemented for next month release.

dongbohu:
Just realized that Aaron Wang's gene-gene network doesn't have negative edge weight at all, so this issue doesn't exist in his web UI. 

After putting more thoughts on whether `Max # of genes` should change when a user changes the radio button. I feel that the current implementation makes some sense too. Suppose initially `Max # of genes` is set to 50, and the radio button is `Both`, and some nodes are shown. When the user clicks `Negative Only` button, he may just want to filter out all positive edges from the current network. It could be a surprise if he sees some new negative edges (and new genes) show up in the network.
 
I will create an issue to keep track of Rene's question.

mhuyck:
Great! I am on the bus back from Boston so will queue up a deployment to run tonight.

Issue #294, opened 2017-10-24
Bundle d3.network.js into release


dongbohu:
The best way to include `d3.network.js` seems to create a repo that includes our own version of `d3.network.js`, then developers in Greene Lab will fork it for code reviews. Then include the greenelab version in `interface/package.json`:
https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

mhuyck:
I like this plan, and it goes a step beyond what I was thinking by figuring out how to incorporate from that repo via our existing `npm` build step... The part I had thought about (which is maybe what @dongbohu means) is:

1. create a fork of `d3.network.js` from [Aaron Wong's repository](https://github.com/aaronkw/d3-network) *using the GreeneLab GitHub account*.
2. GreeneLab developers contributing to this repo can fork from the GreeneLab fork and we can perform our code reviews and PRs as usual
3. any of the GreeneLab changes that might be useful can be offered back to the "upstream" repository via normal pull requests as well (but this should follow whatever procedure Aaron wants to use)

I guess this is pretty much the obvious approach to take.

We should also discuss (as a group ourselves *and* with @aaronkw) which GreeneLab coding standards ought to be applied as part of our code reviews on this file. My initial thought is that we should follow the parts of our protocol aimed at ensuring extra eyes see the code before it's committed to the GreeneLab repo, but we should probably preserve the coding style of the original project. But that should obviously be a discussion.

rzelayafavila:
@dongbohu and @mhuyck: that makes sense - that is what I was thinking as well.

aaronkw:
BTW, I’m not going to maintain or add features to the project since it’s a bit outdated, and we'd like to migrate to d3 4.0 soon. Feel free to maintain it if you'd like!

dongbohu:
Thanks @aaronkw for the update!

PR #336, opened 2017-12-15
Fix of issue #294

dongbohu:
Now that `d3-network` library has been hosted in our own repo (See https://github.com/greenelab/d3-network/pull/1), I am using this PR to fix issue #294. It includes:

(1) Removing the local `d3.network.js`;
(2) Changing the building process;
(3) `network.js` is also modified because of the code change in `d3.network.js`.


interface/package.json
```diff
@@ -17,7 +17,8 @@
   "dependencies": {
     "hclusterjs": "^1.3.1",
     "multtest": "^0.1.7",
-    "ttest": "^1.1.0"
+    "ttest": "^1.1.0",
+    "d3-network": "git+https://github.com/greenelab/d3-network.git"
```
rzelayafavila: :+1:

interface/build.config.js
```diff
@@ -78,7 +78,7 @@ module.exports = {
       'vendor/ng-sortable/dist/ng-sortable.min.js',
       'vendor/angularjs-slider/dist/rzslider.min.js',
       'vendor/d3-tip/index.js',
-      'd3.network.js',
+      'node_modules/d3-network/d3.network.js',
```
rzelayafavila: :+1:

interface/d3.network.js
```diff
@@ -1,678 +0,0 @@
...
-            .attr("text-anchor", "middle")
-            .attr("transform", "translate(" + dx + ", " + dy + ") rotate(-90)")
-            .text(geneLegendText);
-
-    }
-
-    return d3.rebind(my, event, "on");
-} // network
-
-})();
```
rzelayafavila: :+1:

interface/package.json
rzelayafavila: Actually, I have one more question here: Shouldn't this point to the URL for Aaron's repo since it is the one that is in the package.json file of the d3-network package we are now using (and I remember in the previous PR you were saying that this was so because it was the master of the greenelab one).

dongbohu: Good question. I did it on purpose in case Aaron changed his mind. Then we will still use our own fork.

rzelayafavila: Ok, good idea :+1:

Issue #275, opened 2017-10-02
Description of model, when selected


cgreene:
When the user chooses a model on the homepage, can we provide a description of it?

> On the home page or About page, what do you think about adding text like: 
> Referencing ADAGE: 
> The eADAGE model is described in XXXX. Support for the development of this tool came from the Gordon and Betty Moore Foundation (PI, Casey Greene) and the Cystic Fibrosis Foundation (Greene and Hogan, PIs). 
> For an r package that enables the analysis of unpublished datasets not yet incorporated in the model, please refer to XXXX.

The text for eADAGE would be:

>The eADAGE model is described in [Tan et al., 2017. _Cell Systems_](www.cell.com/cell-systems/abstract/S2405-4712(17)30231-4). The development of this model was supported by GBMF 4552 from the Gordon and Betty Moore Foundation (PI, Casey Greene) and the Cystic Fibrosis Foundation STANTO15R0 (Greene and Hogan, PIs). 
> 
> An R package that can be used to analyze datasets that have not yet been shared is available and has been described in [Tan et al., 2017. _bioRxiv_](biorxiv.org/content/early/2017/06/27/156620).

cgreene:
This could be a small text blurb below the model selection field instead of a tooltip if it's easier.

dongbohu:
Fixed in PR #340.

PR #340, opened 2017-12-20
Home: Add model description text on web UI

dongbohu:
This PR is to fix issue #275. I also fixed a couple of minor bugs that I found when working on this one.

On the backend:
(1) Add a new `desc_html` field in `MlModel`;
(2) A new migration file was generated (no need to review); 
(2) Management command and deployment is also changed.

On the front end:
(1) The corresponding `Home` js and html template files are revised;
(2) The `MlModelTracker` service is modified because of the new field in the model;
(3) A minor issue in `ml-model-selector` component is corrected.


dongbohu:
Codeclimate reported a few issues. I keep them on purpose because it is probably better to have the whole URL on one line.

mhuyck:
When did Codeclimate start complaining about things like this?

> Function add_ml_model has 5 arguments (exceeds 4 allowed). Consider refactoring.

That seems new to me, and I don't think we have anything like this included in our coding style guidelines so maybe we should see why it's suddenly popping up. Please correct me if I am wrong.

adage/analyze/models.py
```diff
@@ -149,6 +149,7 @@ class MLModel(models.Model):
     # False (not directed).
     directed_g2g_edge = models.BooleanField(default=False)
     g2g_edge_cutoff = models.FloatField(default=0.0)
+    desc_html = models.CharField(max_length=2048, blank=True)
```
mhuyck: I was a little surprised to find that Django doesn't have any sort of `HTMLField` but I guess a `CharField` can hold HTML just fine. Did you consider using a `TextField` here instead? I don't think I have a strong opinion one way or another. The main difference is that `TextField` has no pre-set limit on length.

There is great potential for abuse of unfiltered user-provided HTML on web sites, but in this case we only allow an administrator to supply a `desc_html` (via the management command) so as long as that remains true we should be safe on that front.

fabfile/adage_server.py
```diff
@@ -197,6 +197,17 @@ def import_data_and_index():
...
+    '<a href="http://www.cell.com/cell-systems/abstract/S2405-4712(17)30231-4">'
+    'Tan et al., 2017. <em>Cell Systems.</em></a> The development of this '
+    'model was supported by GBMF 4552 from the Gordon and Betty Moore '
+    'Foundation (PI, Casey Greene) and the Cystic Fibrosis Foundation '
+    'STANTO15R0 (Greene and Hogan, PIs).'
+    '<p>An R package that can be used to analyze datasets that have not yet '
+    'been shared is available and has been described in '
+    '<a href="https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-017-1905-4">'
+    'Tan et al., 2017. <em>BMC Bioinformatics.</em></a>'
+
```
mhuyck: Indentation seems a little off here so I'm wondering why Codeclimate didn't pick that up. Does this pass muster with a Python linter?

interface/src/app/home/home.tpl.html
```diff
@@ -20,32 +20,36 @@ <h3>ADAGE: Home</h3>
...
-        <p><a ui-sref="analyze({mlmodel: modelInfo.id})">
-            Explore assays and experiments
-        </a></p>
-      </li>
+  <ul>
+    <li>
+      <a ui-sref="analyze({mlmodel: modelInfo.id})">
+        Explore assays and experiments
+      </a>
+    </li>
```
mhuyck: Nice! Thank you for cleaning up the HTML here.

interface/src/app/mlmodel/mlModelComponents.js
```diff
@@ -6,7 +6,7 @@ angular.module('adage.mlmodel.components', [
 .component('mlModelSelector', {
   templateUrl: 'mlmodel/selector.tpl.html',
   bindings: {
-    selectedMlModel: '=',
+    selectedMlModel: '<',
```
mhuyck: 👍

interface/src/app/mlmodel/mlModelComponents.js
```diff
@@ -65,7 +65,7 @@ angular.module('adage.mlmodel.components', [
           self.isValidModel = true;
         },
         function error(err) {
-          MlModelTracker.reset();
+          MlModelTracker.init();
```
mhuyck: I was wondering if it was safe to call `.init()` in an error handler, so I looked up `MLModelTracker` (should really be defined in `src/app/mlmodel`, by the way, because I had a hard time finding it in `utils.js`) and found that it is indeed safe because all it does is clear its properties. I think the name `reset` or `clear` would be more accurate, but I understand why you're using `init()`. Perhaps a simple way around this would be to actually define a `reset()` and call that from `init()`.

Anyway, it's not a huge concern since `init()` currently doesn't currently do anything unsafe, but it would become a problem down the road if we added some code there that wouldn't be good to call during an error handler. I'd recommend at minimum a comment in `MLModelTracker.init()` to guard against that.

interface/src/app/home/home.js
```diff
@@ -2,6 +2,7 @@
  * adage.home module.
  */
 angular.module('adage.home', [
+  'ngSanitize',
```
mhuyck: It looks as if `ngSanitize` is not used anywhere. Am I missing something?

adage/analyze/models.py
dongbohu: I thought about it when adding the new field. But I saw some online article claiming that `CharField` is slightly faster than a `TextField` (I forgot from where though). So I eventually used the former.

fabfile/adage_server.py
dongbohu: Yes, pep8 is happy with it.

interface/src/app/mlmodel/mlModelComponents.js
dongbohu: Good point. Part of the reason why `MlModelTracker` is located in `utils.js` is because this service is needed by almost every front-end module, and what it does is keep track of the current mlmodel, and it is not directly related to any front-end display.
 
The distinction of `init()` and `reset()` makes a lot of sense. Let me add a `reset` method which will call `init()` for now.

interface/src/app/home/home.js
dongbohu: According to:
https://docs.angularjs.org/api/ngSanitize/service/$sanitize
`ng-bind-html` seems to automatically call `$sanitize`, which requires `ngSanitize`.

dongbohu: I tested the example at the end of this page:
https://docs.angularjs.org/api/ng/directive/ngBindHtml
If `ngSanitize` module is removed from dependency, the page won't be rendered correctly. Here is the error:
```
Error: [$sce:unsafe] http://errors.angularjs.org/1.6.9-build.5530+sha.07d84dd/$sce/unsafe
    at https://code.angularjs.org/snapshot/angular.min.js:7:93
    at f (https://code.angularjs.org/snapshot/angular.min.js:157:225)
    at getTrusted (https://code.angularjs.org/snapshot/angular.min.js:158:462)
    at Object.c.(anonymous function) [as getTrustedHtml] (https://code.angularjs.org/snapshot/angular.min.js:160:175)
    at https://code.angularjs.org/snapshot/angular.min.js:286:459
    at m.$digest (https://code.angularjs.org/snapshot/angular.min.js:148:416)
    at m.$apply (https://code.angularjs.org/snapshot/angular.min.js:151:505)
    at https://code.angularjs.org/snapshot/angular.min.js:22:117
    at Object.invoke (https://code.angularjs.org/snapshot/angular.min.js:44:442)
    at c (https://code.angularjs.org/snapshot/angular.min.js:22:38)
```

interface/src/app/mlmodel/mlModelComponents.js
mhuyck: Why don't we *define* the `MLModelTracker` service in `/src/app/mlmodel`, so it can be with all other `MLModel` code, and (since, as you say, it's needed everywhere) *include* it in `utils.js`?

mhuyck: To be more precise: when I say "include", I mean use Angular's dependency mechanism to ensure it is available.

interface/src/app/home/home.js
mhuyck: 👍 

Since that connection is not obvious, can we put a short comment at line 5 saying something like:

```javascript
  'ngSanitize',  // used by ng-bind-html
```

adage/analyze/models.py
mhuyck: I'd guess the trade-off is that `CharField` is less efficient in storage or memory space, but the defined `max_length` makes it faster to execute.

interface/src/app/mlmodel/mlModelComponents.js
dongbohu: Right now `mlmodel` directory has two js files: `mlModelComponents.js` and `resource.js`. Neither of them seems to be a good place for `MLModelTracker` service. `MlModelTracker` doesn't rely on any external modules, I would argue that it is more like a utility. (But I agree that its name may be misleading.) I also don't want to make this PR too complex.

dongbohu:
@mhuyck: I believe I have addressed all your concerns except the relocation of `MlModelTracker` service. Please let me know if you have any other questions. Thanks.

interface/src/app/mlmodel/mlModelComponents.js
mhuyck: I agree that this is not an essential part of this PR... it just came up because I had trouble tracking down `MLModelTracker` during the review since it wasn't where I expected to find it. Since we have more details to discuss let's file an issue so it doesn't hold up the main point of this PR.

dongbohu: Done!

interface/src/app/utils.js
```diff
@@ -34,9 +35,12 @@ angular.module('adage.utils', [])
       this.title = inputModel.title || null;
       this.organism = inputModel.organism || null;
       this.g2gEdgeCutoff = inputModel.g2g_edge_cutoff;
+      this.desc = inputModel.desc_html;
     }
   };
 
+  this.reset = this.init;
+
```
mhuyck: Hah! Clever.

mhuyck:
Ok, this works for me. Thank you, @dongbohu!

Issue #295, opened 2017-10-25
Gene-Gene Network: Explain the algorithms being used to rank the genes 


dongbohu:
With the new slider bar that controls the max number of genes displayed in the network, a natural question from users will be: how are the genes in the network ranked? 

So far `d3.network.js` has two algorithms to rank the genes. But it is not trivial to explain them in a few sentences. The best place to describe the ranking algorithms is probably the `help` page, and the gene-gene network web UI should have a link that points to this description.

cgreene:
@dongbohu : can you point me towards the point in the source code where these are defined. I can try to write some text.

dongbohu:
@cgreene: See lines 60-92 of this file:
https://github.com/aaronkw/d3-network/blob/master/d3.network.js

cgreene:
Ok - here's the summary:

```
Some genes are connected to many genes. Other genes are connected to few genes. Without background correction, the network view prioritizes the genes that are most connected to the query genes. With the optional background correction step, the network view prioritizes genes that are more connected to the query genes than to other genes.
```

My impression is that this selection isn't going to be user facing for us right now though - is that right? I am under the impression that the ADAGE server's default behavior is non-background corrected (in contrast with GIANT, where the default is background corrected).

dongbohu:
@cgreene: Yes, right now the ADAGE server's default behavior is non-background corrected. Do you think version 1.0 should allow users to enable/disable this option?

cgreene:
Hmm - may be confusing. The background correction was added to GIANT based on user feedback, if I recall correctly. I think we debated making it optional and decided not to because it was thought that it'd be too confusing. The same thing is probably true here. If we get users complaining that "the same gene shows up for every query" then we may want to consider enabling it. But these networks may be different enough from GIANT that this is less of a problem, so my inclination is to go with off unless we get some requests.

We should probably put this text somewhere as a comment in the code though. This way if we ever do enable it, we have it close at hand :)

dongbohu:
Let me add the summary above in `d3.network.js` above and link this issue's URL as well.

dongbohu:
@mhuyck and @rzelayafavila: I addressed most of your comments. @mhuyck: for your comment on lines 216-223, I proposed one resolution. If you agree, I will implement it. Thanks.

dongbohu:
Fixed in PR #342.

PR #342, opened 2018-01-10
Network Page: Fixing issue #305 and #295

dongbohu:
This simple PR is for issue #305 and #295.

### DIFF
--- a/interface/src/app/help/help.tpl.html
+++ b/interface/src/app/help/help.tpl.html
@@ -52,7 +52,14 @@
     Information about the PAO1 genes used to build this server was obtained
     from the following NCBI website:
     <a href="ftp://ftp.ncbi.nih.gov/gene/DATA/GENE_INFO/Archaea_Bacteria/Pseudomonas_aeruginosa_PAO1.gene_info.gz"
-    target="_blank">NCBI</a>.
+       target="_blank">NCBI</a>.
+    <br>
+    Some genes are connected to many genes while others are connected to fewer
+    genes. Right now the genes in the network view are ranked without
+    background correction so that the view prioritizes the genes that are most
+    connected to the query genes. (If the correction were applied, the view
+    would prioritize genes that are more connected to the query genes than to
+    other genes.)
   </div>
   <br>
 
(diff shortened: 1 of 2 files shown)


dongbohu:
Here is a demo page:
Click one of the edges whose popup window has a clickable signature name, and a new tab should show up, but the original tab is still there.

dongbohu:
@mhuyck: I added another commit to address issue #295 in this PR too.

interface/src/app/help/help.tpl.html
```diff
@@ -52,7 +52,14 @@ <h3><u>Genes</u></h3>
...
     <a href="ftp://ftp.ncbi.nih.gov/gene/DATA/GENE_INFO/Archaea_Bacteria/Pseudomonas_aeruginosa_PAO1.gene_info.gz"
-    target="_blank">NCBI</a>.
+       target="_blank">NCBI</a>.
+    <br>
+    Some genes are connected to many genes while others are connected to fewer
+    genes. Right now the genes in the network view are ranked without
+    background correction so that the view prioritizes the genes that are most
+    connected to the query genes. (If the correction were applied, the view
+    would prioritize genes that are more connected to the query genes than to
+    other genes.)
```
mhuyck: This is a great explanation. Thank you for nailing it down!

A couple of things will help make this section more readable:

1. This **Genes** section, unlike all others on this page, has three distinct paragraphs rather than just one. They should each be formatted with `<p>...</p>` instead of having `<br>` to allow the browser to show that separation more clearly. In fact, the rest of the page sections should be corrected to use `<p>` tags rather than just bare text after the headers.
2. The explanation in this last part is now clear to me, but the difference between corrected _vs._ uncorrected ranking could be made still more clear with a little emphasis added, I think:

> Some genes are connected to many genes while others are connected to fewer
   genes. Right now, the genes in the network view are ranked <em>without</em>
   background correction so that the view prioritizes the genes that are most
   connected to the query genes. (If the correction were applied, the view
   would prioritize genes that are <em>more connected</em> to the query genes 
   than they are to other genes.)

dongbohu:
@mhuyck I like your comments about adding `<p>` tags in `Genes` section. I have added them in this section, but I didn't add it in the other sections, because so far they only have one paragraph anyway. 

I also followed your comment on making some keywords italic, and removed the extra `<br>` at the end of each section to make the page more compact. Please let me know what you think.

mhuyck:
LGTM. Thanks, @dongbohu!

Issue #153, opened 2017-05-17
Given a set of genes, it would be nice to identify enriched signatures


cgreene:
Deb's lab has people who would like to input a set of genes (maybe as a Tribe geneset?) to identify nodes with significant enrichment. This would be the flip side of functionality in the node table that shows pathways enriched for a node.

dongbohu:
If they just need a list of nodes (with the link to the corresponding node page), we can probably add that functionality into the current gene search web UI:
http://adage.greenelab.com/#/gene_search
All we need to do is add a button like "Show enriched nodes (or signatures)", which when clicked will show the enriched nodes (aka. signatures).

I will talk to Rene when he is back.

cgreene:
@dongbohu : I think this would meet their goals. Thanks!

rzelayafavila:
Changing "nodes" to "signatures" to reflect our current nomenclature.

rzelayafavila:
This is actually much more involved than originally expected, and currently involves many different calls to the backend. Will probably need to create a new api endpoint to retrieve many signatures at once and all of their associated genes (for a given ParticipationType). Kicking this back into next release.

dongbohu:
Rene's WIP PR #343 is supposed to address this issue.

dongbohu:
Implemented in #345.

PR #343, opened 2018-01-12
WIP: Finding enriched signatures for a given set of genes

rzelayafavila:
Changes for https://github.com/greenelab/adage-server/issues/153

### DIFF
--- a/interface/src/app/signature/enriched_signatures.js
+++ b/interface/src/app/signature/enriched_signatures.js
@@ -1,0 +1,222 @@
+/**
+ * "adage.enriched_signatures" module.
+ */
+
+angular.module('adage.enrichedSignatures', [
+  'ui.router',
+  'adage.tribe_client.resource',
+  'adage.gene.resource'
+])
+
+.config(['$stateProvider', function($stateProvider) {
+  $stateProvider.state('enriched_signatures', {
+    url: '/signature/enriched_signatures?mlmodel&genes',
+    views: {
+      main: {
+        templateUrl: 'signature/enriched_signatures.tpl.html',
+        controller: 'EnrichedSignaturesCtrl as ctrl'
+      }
+    },
+    data: {pageTitle: 'Enriched Signatures'}
+  });
+}])
+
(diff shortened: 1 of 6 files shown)


interface/src/app/signature/enriched_signatures.js
```diff
@@ -0,0 +1,195 @@
...
+      var m = userSelectedGenes.length;
+
+      var pValueArray = [];
+
+      var signatureIds = Object.keys(signatures);
+
+      for (var i = 0; i < signatureIds.length; i++) {
+        var signatureId = signatureIds[i];
+        var genes = signatures[signatureId];
+        var n = genes.length;
```
dongbohu: @rzelayafavila: Is `n` the total number of heavy genes associated with `signatureId`?

interface/src/app/signature/enriched_signatures.js
```diff
@@ -0,0 +1,195 @@
...
+      significantSignatureArray.sort(function(a, b) {
+        return a.pValue - b.pValue;
+      });
+
+      return significantSignatureArray;
+    };
+
+
+    Participation.get(
+      {'mlmodel': self.modelInUrl, 'gene__in': $stateParams.genes, 'limit': 0},
```
dongbohu: @rzelayafavila: If `n` on line 104 (see my comment above) is the total number of heavy genes in a certain signature (not limited to the genes in the URL), why we need `gene__in` parameter here? I feel that this parameter should be removed.

interface/src/app/signature/enriched_signatures.js
rzelayafavila: Hi @dongbohu - Yes, it is meant to be the total number of genes that have participation in a signature (for a specific type of participation, can be high weight genes or another type of participation).

interface/src/app/signature/enriched_signatures.js
rzelayafavila: Ah I see - I think that is true if you want to get ALL participations back from the database and then figure out which participations (and therefore signatures) are relevant to this analysis via JavaScript here.

Issue #347, opened 2018-02-09
Allow network export to tab delimited and/or csv


cgreene:
Deb Hogan would like to be able to get the genes and weights out of a network view as a tab delimited or CSV file.

In this case, it would get all of the edges in this network: http://adage.greenelab.com/#/gene_network/?genes=865,937,5403&mlmodel=1

In the format:
`Gene1,Gene2,Weight`

dongbohu:
Considering that some gene names may not be unique, maybe we should also include gene's entrez ID.

cgreene:
Ahh - what about this format to make sure that they get the unique ID (entrez) with the convenient name? `Gene1` here indicates the displayed name in the network:
`Gene1|Entrez,Gene2|Entrez,Weight`

This way, if someone processes it as CSV they still get as expected, but they can also separate on | if desired to get the synonyms split.

mhuyck:
Note: this issue is a variant of #273 and #261. Seems like it would be very useful to have it implemented.

cgreene:
Ahh - yea - this has been requested enough times that I think we could consider this a "very important feature."

mhuyck:
This is currently implemented as described above: with pipes (`|`) separating gene labels from Entrez IDs. It looks funky to me in practice... I propose that we simply  make the Entrez IDs into separate columns, since it's a `.csv` file and needs to be parsed into columns anyway. Will check with the Hogan lab, too, to get their thoughts in case I'm missing something.

Demo site: http://greenelab.fgtech.com/#/gene_network/?genes=5403&mlmodel=1

mhuyck:
Fixed in #350.

PR #350, opened 2018-03-15
Export gene network

### LINKED ISSUE #273: Export gene list from Gene Network view
It would be nice to have a way to export a list of the genes currently showing in the Gene Network view.

(Request from Hogan lab via @cgreene)

mhuyck:
Fixes #273 and #347.

This implements the requests by generating files formatted for download as the network is loaded. Currently offering CSV files only. I've also tightened up the layout of the slider controls a bit to make room for the "Download" menu.


cgreene:
Can you also send to Deb & co via slack for user perspectives?

mhuyck:
@cgreene: 👍  will ping the Hogan lab on Slack after I demo with the tech team.

mhuyck:
Ok... this is ready for review now. The test server at http://greenelab.fgtech.com/#/gene_search?mlmodel=1 has the latest code.

interface/src/app/gene/network/network.js
```diff
@@ -221,6 +221,22 @@ angular.module('adage.gene.network', [
             g.query = true;
           }
         });
+
+        // format a list of genes for download
+        if (self.geneDownload !== null && self.geneDownload !== undefined) {
```
ramenhog: can this just be `if (!!self.geneDownload)`? same goes for `self.edgeDownload`

interface/src/app/app.js
```diff
@@ -33,6 +33,11 @@ angular.module('adage', [
   $resourceProvider.defaults.stripTrailingSlashes = false;
 }])
 
+// To download `blob:` objects, we need that scheme in our whitelist
+.config(['$compileProvider', function($compileProvider) {
+  $compileProvider.aHrefSanitizationWhitelist(/^\s*(https?|ftp|mailto|blob):/);
```
ramenhog: 👍

interface/src/app/gene/network/network.js
```diff
@@ -221,6 +221,22 @@ angular.module('adage.gene.network', [
...
+          // release a previously-created Blob
+          URL.revokeObjectURL(self.geneDownload);
+        }
+        self.geneDownload = URL.createObjectURL(new Blob(
+          genes.map(function(g, index) {
+            var geneStr = g.label + ',' + g.entrezid + '\n';
```
ramenhog: it looks like we're using Babel? If so, we can probably just use a template literal for this.

interface/src/app/gene/network/network.less
```diff
@@ -62,3 +96,8 @@ text.legend-title {
 #gene-slider .rz-selection {
   background: magenta;
 }
+
+#download-menu {
```
ramenhog: It's usually not ideal to use id's for CSS styling because it can lead to specificity wars:
https://stuffandnonsense.co.uk/archives/css_specificity_wars.html (an old article but still holds true...and you can't beat star wars metaphors 😉)

ramenhog:
Just a few small comments here and there, but overall looks good!

interface/src/app/gene/network/network.js
mhuyck: That is a nice thought. We have incorporated Babel for a couple specific libraries that were written in ES6, but everything else in the codebase is ES5 and the build process does not currently support piping through Babel in a general way. (As an aside: I'd love to be using a more current build methodology NOW, but it will be a substantial project to retool. We have several steps planned that will take us in the right direction, but it is not the top priority.)

interface/src/app/gene/network/network.js
mhuyck: Yes, that is *way* more succinct. I had the `!== null` first and it turned out I had meant `!== undefined`, so that's how it ended up this way.

interface/src/app/gene/network/network.less
mhuyck: I'm aware of that, but didn't have a better idea here. Suggestions?

We're already using IDs for the `#gene-slider` and `#edge-slider`, so we can't fully eliminate the technique unless we solve it for those cases also. You probably know this, but keep in mind that the `.less` we write gets concatenated into one large `.css` file during our build, so we need enough specificity that we don't inadvertently target unrelated components and views.

We definitely need better structure in our `.less` organization with this project, so I very much welcome your advice.

ramenhog: My favorite approach for reducing specificity but still making sure styles are targeting the correct components is to use [BEM](http://getbem.com/) 😍 . Its a very straightforward naming convention for modular CSS.

But obviously, that's a larger fix, so this isn't a blocker for approving this PR. Just something to think about :)

mhuyck: Excellent. I have seen BEM before, but not sure how I came across it. Maybe Deepa? It's good to have a way forward for better CSS structure.

Anyway, in this context, I think the part of that methodology that applies is the concept of the "block". You are right that we can't fix the whole codebase now, but I'd like to not actually make it worse. I'm going to switch to using `.class` here to take a step in the right direction.

We're really in this mess because nobody sticks to writing semantic HTML (bootstrap is a major offender, IMO) so we have to use CSS in ways it wasn't meant for to make up for that... but that's an even *bigger* problem than the CSS in this codebase.

mhuyck:
Pushed fixes that should address everything. Thank you for the feedback @ramenhog!

PR #351, opened 2018-04-12
Separate heatmap into its own view

### LINKED ISSUE #62: Heatmap page needs to be parametrized
Noted by @dongbohu in PR #59:

> (1) Since the heatmap is a separate page now (instead of modal), does it make sense to include the sample IDs in the URL instead of using a generic #/analysis-detail? That way if a user wants to show the heatmap to another user, he/she can just send the URL. (I understand that if there are too many samples involved in the heatmap, the URL may reach its maximum length, so we may need a better solution if this functionality is really needed.)

I fully agree that the heatmap state should be parametrized to make URL sharing work properly.

Steps to complete this work:
- [x] refactor the `analysis-detail` page to put the heatmap into a distinct component (https://github.com/greenelab/adage-server/issues/278)
- [x] update the heatmap's vega spec to show sample labels rather than IDs (https://github.com/greenelab/adage-server/issues/279)
- [x] separate the heatmap into its own "view" page, which can be parameterized to accept an MLModel and a list of sample IDs via the URL (https://github.com/greenelab/adage-server/issues/280)

### LINKED ISSUE #279: update the heatmap's vega spec to show sample labels rather than IDs
This is a sub-issue (part 2) of #62.

### LINKED ISSUE #280: separate the heatmap into its own "view" page
This is a sub-issue (part 3) of #62.

This view page can now be parameterized to accept an MLModel and a list of sample IDs via the URL, thus closing #62.

mhuyck:
This set of changes:

- fixes #62,
- fixes #279, and
- fixes #280.

Test server up and running at http://greenelab.fgtech.com/


interface/src/app/sample.service.js
```diff
@@ -74,8 +76,43 @@ angular.module('adage.sample.service', [
           $log.error($scope.analysis.queryStatus);
         }
       ).$promise;
+      Sample.promises[pk] = pSample;
       return pSample;
     };
+    Sample.getSampleListPromise = function(pkList) {
```
dongbohu: Can you add a blank line between lines 81 and 82? I think it will make the code more readable.

interface/src/app/heatmap/heatmap-vgspec.js
```diff
@@ -80,8 +98,8 @@ angular.module('adage.heatmap-vgspec', [])
       'name': 'samples',
       'type': 'ordinal',
       'domain': {
-        'data': 'activity_normalized',
-        'field': 'sample_order._id',
+        'data': 'samples_labeled',
```
dongbohu: I noticed that you changed "sample_objects" to "sampleObjects" earlier. If you are trying to convert the variables to camelCase, you may want to convert this one as well.

ramenhog:
It looks good to me too 👍 !

mhuyck:
Awesome. Thanks to you both! I will incorporate @dongbohu’s suggestions and bump the version up to `0.11.0-beta` before deploying.

interface/src/app/heatmap/heatmap-vgspec.js
mhuyck: Aha... this is inside the Vega spec, which is actually JSON, not JavaScript. I had switched the `sampleObjects` label since that dataset originated from the JavaScript that invokes the heatmap. I took a quick pass through the examples on the Vega web site and they are all over the map with camelCase, snake_case, and sometimes even both in the same spec.

It looks as if there isn't really a more general standard for JSON either. StackOverflow has this good discussion, which pretty much covers all of the angles:

> https://stackoverflow.com/questions/5543490/json-naming-convention

I think we should lean toward camelCase, since we're using AngularJS (JavaScript, obviously) as a supporting framework. I will change just this label now to minimize last minute code changes and do the rest as part of our migration from Vega 2 to Vega 3.

PR #352, opened 2018-05-09
Filter download files to match Gene Network view

mhuyck:
The recently-added feature to download genes and edges from the Gene Network view did not take account of the filter settings. This ability was requested by a user, so it is now the default that downloaded files will match what is shown on the page at the time.

This approach does not lose any features: adjusting the sliders to perform no filtering will result in download files that match the original implementation.


mhuyck:
There is a test server running here: http://greenelab.fgtech.com/#/gene_network/?genes=844&mlmodel=1

mhuyck:
Huh. That is an odd message from CircleCI, and it's not totally clear what's causing that build to fail (even after checking the logs). Codeship build was fine.

mhuyck:
Thanks, @dongbohu! I had originally considered making a setting to do what you're talking about. I realized after implementing this that users *still* have that option by setting the sliders back to show everything. We could add, say, a checkbox to alter the download but I think that would clutter the UI and make it more confusing than necessary.

I think I'd like to try releasing this feature as-is and see if any real-world use drives a request for such a setting. Does that sound like a good plan to you?

dongbohu:
@mhuyck: I see your point. I am a little concerned that when a user clicks one of these two buttons, he/she may expect to get ALL genes or edges instead of only the ones displayed with current slider setup. Should we make the texts of the download two buttons more obvious (such as "visible genes/edges", or something like that)?

mhuyck:
🤔 I share your concern. I also agree that it is a little concern, so I don't want it to hold things up since we have a solution that our user is happy with.

I think both of us have intuition that there's something missing usability-wise on this page. After giving it some thought, it seems to me that the root problem is that it's not totally clear when the filters are doing any actual filtering and when you're seeing all of the available data. I think I can make a simple change that will clarify that. Taking a stab at that idea now.

PR #353, opened 2018-05-10
[FIX] stop add button from leaving search results page

### LINKED ISSUE #284: "+ Add" button should not leave the search results page
Reported by @jaclyn-taroni in #142:

> if I add all samples from an experiment with the +Add button, I am still taken to the experimental detail page.

### LINKED ISSUE #319: Analyze Page: The user is navigated away from the search results when they add a sample set to analyze
**Priority 3: Major usability problem: important to fix, so should be given high priority**
This may be a little disorienting if they wanted to add more samples.

mhuyck:
This update fixes #284 by capturing the `$event` from the `+ Add` button press and telling the browser to stop propagation back to the `<a>` tag wrapping the rest of the search result. It turned out that similar problems existed in the event handling for the neighboring `Analyze` button, so that is now fixed as well.

Credit to this very well-written Stack Overflow post for the Angular-specific fix:

> https://stackoverflow.com/questions/20300866/angularjs-ng-click-stoppropagation

Also fixes #319 (which is a dup).


mhuyck:
Rats. CircleCI's complaint on the last PR was apparently not a fluke. Nothing changed with our build code, so there must be a change on the CircleCI side that breaks our build there. I will check into that.

mhuyck:
Looks like I nailed the CircleCI fix. @dongbohu: have a look and let me know if you think this is a good approach. Line 11 looks like a duplicate, but I left it alone because the comment there reminded me we added that to solve a separate installation dependency issue.

Credit to this SO issue for the resolution I used:

> https://stackoverflow.com/questions/49932759/pip-10-and-apt-how-to-avoid-cannot-uninstall-x-errors-for-distutils-packages

PR #354, opened 2018-06-05
[FIX] remove vestiges of analysis.detail view

### LINKED ISSUE #170: Review app/analyze component to fix detail.showing root cause
In PR #161, a bug arose that turned out to be caused by the fact that `interface/src/app/analyze/analyze.tpl.html` had initialized the variable `$scope.detail.showing` as a side-effect. This "implied" `detail` object should really be explicitly initialized in the controller and the fact that it's not indicates a flaw in the design logic for this module. This code should be reviewed to ensure it is robust.

mhuyck:
Fixes #170.

This eliminates the last of the dead code remaining from the funky split screen analysis view. I had some good ideas in there, but they got in the way of a properly parameterized web view.


mhuyck:
Aha, good question, @dongbohu. Since this PR removes dead code there isn't any change to see... but it's a good idea to verify that all is still working. I did that on my dev instance (of course), but not the demo server. Just now updated the site at http://greenelab.fgtech.com so you can review as well.

PR #369, opened 2019-01-22
Remove Codeship config and update Django packages

dongbohu:
This PR removes `Codeship` from Django settings, and updates a few Django packages.

These changes have been validated on my desktop and confirmed by CircleCI tests.

### DIFF
--- a/adage/adage/settings.py
+++ b/adage/adage/settings.py
@@ -19,11 +19,11 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # come from config.py, which is never checked into source control.
 # If we're running under CircleCI, we build a CONFIG on the fly using
 # `config.py.template`.
-if os.environ.get('CIRCLECI') == 'true':
+if os.environ.get('CIRCLECI', 'false'):
     with open(os.path.join(BASE_DIR, 'adage', 'config.py.template')) as f:
         exec f
     CONFIG = CIRCLECI_CONFIG
-elif os.environ.get('DOCKER_DEV') == 'true':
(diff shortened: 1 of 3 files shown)


adage/requirements.txt
```diff
@@ -1,9 +1,9 @@
 argparse==1.2.1
```
dhimmel: From https://pypi.org/project/argparse/

> As of Python >= 2.7 and >= 3.2, the argparse module is maintained within the Python standard library. For users who still need to support Python < 2.7 or < 3.2, it is also provided as a separate package, which tries to stay compatible with the module in the standard library, but also supports older Python versions.

Should argparse be removed from environment?

adage/adage/settings.py
```diff
@@ -16,15 +16,10 @@
...
-    with open(os.path.join(BASE_DIR, 'adage', 'config.py.template')) as f:
-        exec f
-    CONFIG = CODESHIP_CONFIG
-    CONFIG['databases']['default']['USER'] = os.environ.get('PG_USER')
-    CONFIG['databases']['default']['PASSWORD'] = os.environ.get('PG_PASSWORD')
-elif os.environ.get('CIRCLECI') == 'true':
+# come from config.py, which is never checked into source control.
+# If we're running under CircleCI, we build a CONFIG on the fly using
+# `config.py.template`.
+if os.environ.get('CIRCLECI') == 'true':
```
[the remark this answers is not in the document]
dongbohu: Good point, I will do that.

adage/requirements.txt
dongbohu: Yet another good point.

